PR #4, opened 2023-06-21
Implement celestia types

oblique:


types/src/blob.rs
```diff
@@ -1,60 +1,44 @@
-use celestia_proto::tendermint::types as tmtypes;
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Serialize};
+use tendermint_proto::v0_34::types::Blob as RawBlob;
+use tendermint_proto::Protobuf;
 
+use crate::nmt::Namespace;
+use crate::{Bytes, Error};
+
+// NOTE: We do not the `serce(try_from)` pattern for this type
```
zvolin, line 8:
```suggestion
// NOTE: We don't do the `serde(try_from)` pattern for this type
```

types/src/blob.rs
```diff
@@ -1,60 +1,44 @@
...
+// becase JSON representation needs to have `commitment` field but
+// Protobuf definition doesn't.
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Blob {
-    blob: tmtypes::Blob,
-    commitment: Vec<u8>,
+    pub namespace: Namespace,
+    pub data: Bytes,
```
zvolin, line 14: can't we annotate those like with other protobuf types? this would allow us to remove the Bytes module
```
#[serde(with = "tendermint_proto::serializers::bytes::vec_base64string")]
```

types/src/blob.rs
```diff
@@ -1,60 +1,44 @@
...
+    pub share_version: u8,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub commitment: Option<Bytes>,
```
zvolin, line 17: is an `Option` there kind of a todo? I'm not sure `celestia-node` allows for an empty commitment

zvolin (approved):
Looks good. left some questions

types/src/blob.rs
oblique: yes, i can do that

types/src/blob.rs
oblique: will check

types/src/blob.rs
```diff
@@ -1,60 +1,44 @@
...
+use serde::{Deserialize, Serialize};
+use tendermint_proto::v0_34::types::Blob as RawBlob;
+use tendermint_proto::Protobuf;
 
+use crate::nmt::Namespace;
+use crate::{Bytes, Error};
+
+// NOTE: We do not the `serce(try_from)` pattern for this type
+// becase JSON representation needs to have `commitment` field but
+// Protobuf definition doesn't.
```
zvolin, line 10: I'm not sure why is that preventing the `try_from` and `into`, the Go impl seems to be creating the commitment on blob creation, so we could have that in `try_from` and be consistent with this pattern. You can leave a todo for creating a valid commitment and just put an empty vec / zeroes there

zvolin: ahh true

PR #7, opened 2023-06-23
feat: add NamespacedShares type

oblique:


types/src/nmt.rs
```diff
@@ -4,38 +4,76 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
 
-        let mut bytes = [0u8; NS_ID_SIZE];
-        bytes[NS_ID_SIZE - id.len()..].copy_from_slice(id);
+        let mut bytes = [0u8; NS_SIZE];
+        bytes[NS_SIZE - id.len()..].copy_from_slice(id);
 
         Ok(Namespace(nmt_rs::NamespaceId(bytes)))
     }
 
+    fn new_v255(id: &[u8]) -> Result<Self> {
```
oblique, line 61: I left this intentionally non-public because v255 is an invalid version

types/src/nmt.rs
```diff
@@ -4,38 +4,76 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
         }
     }
 
     pub fn new_v0(id: &[u8]) -> Result<Self> {
-        if id.len() > NS_ID_V0_MAX_SIZE {
-            return Err(Error::InvalidNamespaceIdSize(id.len()));
+        let id_pos = match id.len() {
+            // Allow 28 bytes len
+            NS_ID_SIZE => NS_ID_SIZE - NS_ID_V0_SIZE,
+            // Allow 18 bytes len or less
```
zvolin, line 41:
```suggestion
            // Allow 10 bytes len or less
```
this or using actual name of constant maybe?

types/src/nmt.rs
zvolin, line 61: I'm not sure if I'm for `new_v255` name. I'd name it `try_max` because it just does validation and returns the `max()` namespace. The name `new_v255` may indicate there is some actual 255 version, where it's just a special constant like `-1` in C, where all bytes have to be 255, not just version.

types/src/nmt.rs
```diff
@@ -4,38 +4,76 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+        if id.iter().all(|&x| x == 0xff) {
+            Ok(Namespace::max())
+        } else {
+            Err(Error::UnsupportedNamespaceVersion(255))
+        }
+    }
+
+    pub const fn max() -> Self {
+        Namespace(nmt_rs::NamespaceId::max_id())
+    }
```
zvolin, line 71: We could make it an associated const instead, as const functions do not mean they'll be evaluated at compilation time, it just means they can be. So calling this in a code will actually construct a new namespace on a stack where using an associated const wouldn't

types/src/nmt.rs
zvolin: not sure if that matters as one can just use public from raw

types/src/share.rs
```diff
@@ -0,0 +1,172 @@
...
+        let shares = value
+            .shares
+            .into_iter()
+            .map(Share::new)
+            .collect::<Result<Vec<_>>>()?;
+
+        let proof = value
+            .proof
+            .map(proof_from_proto)
+            .transpose()?
```
zvolin, line 64: oh it got stabilized :pray:

types/src/share.rs
```diff
@@ -0,0 +1,172 @@
...
+        return Err(Error::InvalidNamespacedHash);
+    }
+
+    let min_ns = Namespace::from_raw(&node[..NS_SIZE])?;
+    let max_ns = Namespace::from_raw(&node[NS_SIZE..NS_SIZE * 2])?;
+
+    let mut hash = [0u8; nmt_rs::HASH_LEN];
+    hash.copy_from_slice(&node[NS_SIZE * 2..]);
+
+    Ok(NamespacedHash::new(min_ns.into(), max_ns.into(), hash))
```
zvolin, line 110:
```suggestion
    node.try_into().map_err(|_| Error::InvalidNamespacedHash)
```

types/src/share.rs
```diff
@@ -0,0 +1,172 @@
...
+            namespace: Namespace::from_raw(ns)?,
+            data: data.to_vec(),
+        })
+    }
+
+    fn _to_vec(&self) -> Vec<u8> {
+        let mut bytes = self.namespace.as_bytes().to_vec();
+        bytes.extend_from_slice(&self.data);
+        bytes
+    }
```
zvolin, line 46: shouldn't this (and `new`) be public instead?

zvolin: *instead of `_` prefix

zvolin (approved):
lgtm, left some ideas

types/src/share.rs
oblique: I didn't saw that, thanks

types/src/nmt.rs
oblique: `try_max` sounds good

types/src/share.rs
oblique: The main reason that `new` is not public is because the user will never construct it manually from raw bytes. Construction will be done only from `RawRow`. I added `_` prefix on `to_vec` intetionally since it was dead code and it is something that we will use in `impl From<NamespacedRow> for RawRow`. Another option was to add `#[allow(dead_code)]`. I am ok with both of them.

PR #9, opened 2023-06-28
feat(types): implement `ExtendedHeader::verify`

oblique:


types/src/extended_header.rs
```diff
@@ -198,4 +272,45 @@ mod tests {
...
+        let eh_block_27 = sample_eh_block_27();
+        eh_block_27.verify(&eh_block_27).unwrap_err();
+    }
+
+    #[test]
+    fn verify_invalid_chain_id() {
+        let eh_block_1 = sample_eh_block_1();
+        let mut eh_block_27 = sample_eh_block_27();
+
+        eh_block_27.header.chain_id = "1112222".parse::<Id>().unwrap();
```
zvolin, line 295:
```suggestion
        eh_block_27.header.chain_id = "1112222".parse().unwrap();
```
isn't type redundant?

types/src/trust_level.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    pub fn numerator(&self) -> u64 {
+        self.numerator
+    }
+
+    pub fn denominator(&self) -> u64 {
+        self.denominator
+    }
+
+    pub fn total_power_needed(
```
zvolin, line 26:
```suggestion
    pub fn voting_power_needed(
```
I'd name it like so, as there are `total_voting_power` and `voting_power_needed`

types/src/validator_set.rs
```diff
@@ -78,13 +92,71 @@ impl ValidatorSetExt for Set {
...
+            if tallied_voting_power > voting_power_needed {
+                return Ok(());
+            }
+        }
+
         Err(Error::NotEnoughVotingPower(
             tallied_voting_power,
             voting_power_needed,
-        ))
+        ))?
```
zvolin, line 149: this `?` changes the type of this `Err` to `Err::<Result<()>, Error>`. I've been wrapping my head how does it work for a good while. If you don't mind, I'd prefer this (and one from previous `verify_commit...`) without `?` so that this will be of type `Err::<(), Error>`

zvolin (commented):
lgtm, only left some nits

types/src/validator_set.rs
oblique: Sorry it was a leftover when I was experimenting with errors. I will remove it.

zvolin (approved):
Looks good

PR #24, opened 2023-07-04
feat(rpc): Add all calls for Blob, Share, and Header

### LINKED ISSUE #25: Fix inconsistencies between nmt-rs and go-nmt proofs

oblique:
closes #25


types/src/share.rs
```diff
@@ -57,6 +58,32 @@ impl Share {
     }
 }
 
+impl Serialize for Share {
```
zvolin, line 61: could we follow the serde `from` and `to` convention also there?
```rs
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(try_from = "RawShare", into = "RawShare")]
pub struct Share {
    pub namespace: Namespace,
    pub data: Vec<u8>,
}

#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(transparent)]
struct RawShare {
    #[serde(with = "tendermint_proto::serializers::bytes::base64string")]
    data: Vec<u8>,
}
```

types/src/rsmt2d.rs
zvolin, line 1: while not `eds.rs` or something like that? is that a celestia convention?

rpc/tests/share.rs
```diff
@@ -3,20 +3,22 @@ use celestia_types::consts::appconsts::{
...
 
 pub mod utils;
 
-use utils::{random_bytes, random_ns, test_client, AuthLevel};
+use crate::utils::client::blob_submit;
+use crate::utils::{random_bytes, random_ns, random_ns_range, test_client, AuthLevel};
 
-async fn test_get_shares_by_namespace(client: &HttpClient) {
+#[tokio::test]
+async fn test_get_shares_by_namespace() {
```
zvolin, line 15: could you remove all the `test_` prefixes? I added them because it was just a regular function before

types/src/data_availability_header.rs
```diff
@@ -15,11 +16,17 @@ use crate::{bail_validation, Error, Result, ValidateBasic, ValidationError};
...
 pub struct DataAvailabilityHeader {
-    pub row_roots: Vec<Vec<u8>>,
-    pub column_roots: Vec<Vec<u8>>,
+    pub row_roots: Vec<NamespacedHash>,
+    pub column_roots: Vec<NamespacedHash>,
     pub hash: Hash,
 }
 
+impl DataAvailabilityHeader {
+    pub fn root_hash(&self) -> Option<&NamespacedHash> {
```
zvolin, line 25: each row in row_roots are root hashes, in tests we use the 0 index because as only we submit blobs to the blocks, and we can only submit one per block so it always ends up in first row.
We could instead have an api like this
```rs
pub fn row_root(row: usize) -> Option<&NamespacedHash>
pub fn column_root(row: usize) -> Option<&NamespacedHash>
```

rpc/tests/header.rs
```diff
@@ -36,8 +34,83 @@ async fn test_get_by_height() {
...
+
+    let genesis_header = client.header_get_by_height(1).await.unwrap();
+    let second_header = client.header_get_by_height(2).await.unwrap();
+
+    let headers = client
+        .header_get_verified_range_by_height(&genesis_header, 3)
+        .await
+        .unwrap();
+
+    assert_eq!(headers.len(), 1);
```
zvolin, line 68: as per docs, this should return 2 headers, no?
https://node-rpc-docs.celestia.org/#header

rpc/tests/share.rs
```diff
@@ -72,45 +80,62 @@ async fn test_get_shares_by_namespace_wrong_ns(client: &HttpClient) {
...
+    assert_eq!(ns_shares.rows.len(), 1);
+    assert!(ns_shares.rows[0].shares.is_empty());
+
+    let proof = &ns_shares.rows[0].proof;
+    assert!(proof.is_of_absence());
+
+    let no_leafs: &[&[u8]] = &[];
+
+    proof
+        .verify_complete_namespace(&root_hash, no_leafs, random_ns.into())
```
zvolin, line 97:
```suggestion
    let no_leaves: &[&[u8]] = &[];

    proof
        .verify_complete_namespace(&root_hash, no_leaves, random_ns.into())
```

rpc/tests/utils/client.rs
```diff
@@ -0,0 +1,72 @@
...
+use celestia_types::Blob;
+use jsonrpsee::core::client::ClientT;
+use jsonrpsee::core::Error;
+use jsonrpsee::http_client::HttpClient;
+use jsonrpsee::ws_client::WsClient;
+use once_cell::sync::Lazy;
+use std::env;
+use tokio::sync::Mutex;
+
+static LOCK: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
```
zvolin, line 13: can we have this static inside the blob_submit function? that'd make it impossible to be locked anywhere else

zvolin (approved):
lgtm

rpc/tests/utils/client.rs
zvolin: Also I know `Lazy` is not stable yet, but we could remove the `once_cell` dependency and do it through stabilized `once_lock`
```rs
{
    static LOCK: OnceLock<Mutex<()>> = OnceLock::new();
    let _guard = LOCK.get_or_init(|| Mutex::new(())).lock().await;
    client.blob_submit(blobs).await
}
```

rpc/tests/header.rs
oblique: this is what i understood too, but it actually returns `(from, to)`. i think this is because you already have the `from` header.

zvolin: sure, that's what I thought. For me it's still the bug tho that's worth reporting

rpc/tests/utils/client.rs
```diff
@@ -6,15 +9,16 @@ use jsonrpsee::core::client::ClientT;
...
-use std::env;
-use tokio::sync::Mutex;
-
-static LOCK: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
+use tokio::sync::{Mutex, MutexGuard};
 
 const WS_URL: &str = "ws://localhost:26658";
 const HTTP_URL: &str = "http://localhost:26658";
 
+async fn lock() -> MutexGuard<'static, ()> {
```
zvolin, line 17: can we name it `write_lock`? I'd prefer it inlined in `submit_blob` but I guess you keep it separated in case in future we add another write api that needs that lock.

rpc/tests/header.rs
oblique: it is an issue in the documentation: https://github.com/celestiaorg/go-header/issues/72

zvolin (approved):
lgtm

PR #26, opened 2023-07-05
feat: Add proves verification in integration tests

### LINKED ISSUE #25: Fix inconsistencies between nmt-rs and go-nmt proofs

zvolin:
closes #25

### DIFF
--- a/rpc/tests/blob.rs
+++ b/rpc/tests/blob.rs
@@ -1,6 +1,7 @@
 use std::time::Duration;
 
 use celestia_rpc::prelude::*;
+use celestia_types::nmt::NamespacedHash;
 use celestia_types::{Blob, Commitment};
 use jsonrpsee::http_client::HttpClient;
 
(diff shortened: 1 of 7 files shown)


rpc/tests/blob.rs
```diff
@@ -1,7 +1,7 @@
 use std::time::Duration;
 
 use celestia_rpc::prelude::*;
-use celestia_types::{Blob, Commitment};
+use celestia_types::{nmt::NamespacedHash, Blob, Commitment};
```
oblique, line 4:
```suggestion
use celestia_types::nmt::NamespacedHash;
use celestia_types::{Blob, Commitment};
```

types/src/blob.rs
```diff
@@ -34,15 +36,18 @@ impl Blob {
     }
 
     pub fn validate(&self) -> Result<()> {
-        let computed_commitment =
-            Commitment::generate(self.namespace, self.share_version, &self.data)?;
+        let computed_commitment = Commitment::generate(self.namespace, self.to_shares()?)?;
```
oblique, line 39: I think I prefer the following API: 

* Rename old `Commitment::generate` to `Commitment::with_blob_data` (or `Commitment::from_blob_data`?)
* Add `Commitment::with_shares` (or `Commitment::from_shares`?)

zvolin: I can just leave the `Commitment::generate` if you prefer, we don't need the version from shares

oblique: Yes, I prefer the older API

zvolin: reverted

oblique:
Closing because we incorporated the changes in #24

PR #31, opened 2023-07-17
feat: Add State RPC and types

oblique:


types/src/state/balance.rs
```diff
@@ -0,0 +1,167 @@
...
+    let mut chars = denom.chars();
+
+    // First character must be a letter
+    if !matches!(chars.next(), Some('a'..='z' | 'A'..='Z')) {
+        return Err(Error::InvalidBalanceDenomination(denom.to_owned()));
+    }
+
+    // The rest can be a letter, a number, or a symbol from '/', ':', '.', '_', '-'
+    if chars.all(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '/' | ':' | '.' | '_' | '-')) {
+        Ok(())
```
zvolin, line 79: wondering about using `char::is_ascii_alphabetic` and `char::is_ascii_alphanumeric` where possible, but it looks clean that way anyway so probably not worth it

types/src/state/address.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    fn try_from(value: Raw) -> Result<Self, Self::Error> {
+        value.addr.parse()
+    }
+}
+
+impl From<Address> for Raw {
+    fn from(value: Address) -> Self {
+        let addr = value.to_string();
+        Raw { addr }
```
zvolin, line 138: this could be possibly implemented for `T: AddressTrait`

zvolin (approved):
LGTM

PR #40, opened 2023-08-29
chore: Update protobuf definitions

oblique:


zvolin (commented):
this will probably need an update as we've got new protos' source

oblique:
Yeah. I will do a final update before I mark it as ready.

proto/build.rs
```diff
@@ -98,7 +90,7 @@ fn main() -> Result<()> {
                 "vendor/cosmos/tx/v1beta1/tx.proto",
                 "vendor/go-header/p2p/pb/header_request.proto",
             ],
-            &["vendor"],
+            &["vendor", "vendor/nmt"],
```
zvolin, line 93: is this a bug (or inconsistency) in their change?

oblique: Inconsistency from our point of view. Their code is correct. I had two options:

1. `proto/vendor/pb/proof.proto`
2. `proto/vendor/nmt/pb/proof.proto`

I choose 2 to make it consistent to our directory structure.

tools/update-proto-vendor.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+cd -- "$(dirname -- "${BASH_SOURCE[0]}")" || exit 1
+cd ../proto || exit 1
+set -x -e
```
zvolin, line 5: if you start with setting `-e` then `exit 1` becomes redundant. I also like adding `-u` and `-o pipefail` to detect some cases more easily

```suggestion
set -xeuo pipefail

cd -- "$(dirname -- "${BASH_SOURCE[0]}")"
cd ../proto
```

zvolin (approved):
looks good :+1:

proto/build.rs
zvolin: yeah, I think they just haven't noticed, but we don't have to bother them

proto/README.md
```diff
@@ -1,7 +1,18 @@
+## Update
```
oblique, line 1: @zvolin  Which path do you prefer `proto/vendor/README.md` or `proto/README.md`?

The problem with `proto/README.md` is that it would not be nice when we publish the crate.
The problem with `proto/vendor/README.md` is that is not noticable.

PR #45, opened 2023-09-05
feat(p2p): add NetworkInfo command

zvolin:


node/src/p2p.rs
```diff
@@ -110,33 +125,55 @@ impl P2p {
...
                 },
-                // TODO: receive command from `P2p` and handle it
+                Some(cmd) = command_stream.next() => {
+                    match cmd {
+                        P2pCmd::NetworkInfo { respond_to } => {
+                            if respond_to.send(self.swarm.network_info()).is_err() {
+                                warn!("Couldn't send response for the network info cmd");
+                            }
+                        }
+                    }
```
oblique, line 175: This is going to get very big at some point, I suggest this:

```
Some(cmd) = command_stream.next() => self.on_command(cmd).await
```

and then something like this:

```
async fn on_command(&mut self, cmd: P2pCmd) -> Result<()> {
    match cmd {
        P2pCmd::NetworkInfo { respond_to } => {
            // We don't handle the error because we don't care if handle got dropped
            let _ = respond_to.send(self.swarm.network_info());
        }
    }
}
```

If handling of `NetworkInfo` was bigger than 3 line I would also define `fn on_network_info_cmd()`.

node/src/p2p.rs
```diff
@@ -110,33 +125,55 @@ impl P2p {
...
     pub async fn next_event(&self) -> Option<P2pEvent> {
         todo!();
     }
+
+    pub async fn network_info(&self) -> Result<NetworkInfo> {
+        let (tx, rx) = oneshot::channel();
+        self.cmd_tx
+            .send_async(P2pCmd::NetworkInfo { respond_to: tx })
+            .await
+            .map_err(|_| P2pError::WorkerDied)?;
```
oblique, line 152: Wrap these in `async fn send_cmd(cmd: P2pCmd) -> Result<()>`

fl0rek (approved):
:guitar:

PR #46, opened 2023-09-05
fix: make celestia-rpc to compile in wasm32 target

oblique:


zvolin (commented):
maybe we could include the wasm build on CI?

oblique:
Added a `build-wasm` job on CI, however I don't build tests. We need to implement more stuff for that.

celestia/Cargo.toml
```diff
@@ -4,11 +4,15 @@ version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
 
+#[lib]
+#crate-type = ["cdylib"]
+
```
fl0rek, line 9: Are we planning on changing that for example for wasm? Or is it just leftover from editing?

oblique: fixed

fl0rek (approved):
:guitar:

zvolin (approved):

PR #52, opened 2023-09-06
feat: add RPC calls for p2p module and tests for them

fl0rek:


rpc/Cargo.toml
```diff
@@ -19,3 +22,18 @@ anyhow = "1.0.71"
 dotenvy = "0.15.7"
 rand = "0.8.5"
 tokio = { version = "1.32.0", features = ["rt", "macros"] }
+env_logger = "*"
+log = "*"
+futures = "0.3.28"
+libp2p = { version = "*", features = [
```
oblique, line 28: define the version of the crates

rpc/Cargo.toml
```diff
@@ -19,3 +22,18 @@ anyhow = "1.0.71"
...
+  "ed25519",
+  "gossipsub",
+  "identify",
+  "macros",
+  "noise",
+  "ping",
+  "request-response",
+  "tcp",
+  "tokio",
+  "yamux",
```
oblique: remove the unneeded features

rpc/src/p2p.rs
```diff
@@ -1,17 +1,207 @@
...
+    pub system: Stat,
+    #[serde(rename = "Transient")]
+    pub transient: Stat,
+    #[serde(rename = "Services")]
+    pub services: HashMap<String, Stat>,
+    #[serde(rename = "Protocols")]
+    pub protocols: HashMap<String, Stat>,
+    #[serde(rename = "Peers")]
+    pub peers: HashMap<String, Stat>,
+}
```
oblique, line 44: use [`rename_all`](https://serde.rs/container-attrs.html#rename_all) for the field names

rpc/src/p2p.rs
```diff
@@ -1,17 +1,207 @@
...
+
+    #[serde(rename = "TotalOut")]
+    pub total_out: f32,
+
+    #[serde(rename = "RateIn")]
+    pub rate_in: f32,
+
+    #[serde(rename = "RateOut")]
+    pub rate_out: f32,
+}
```
oblique, line 101: use `rename_all`

rpc/src/p2p.rs
```diff
@@ -1,17 +1,207 @@
...
+            1 => Ok(Connectedness::Connected),
+            2 => Ok(Connectedness::CanConnect),
+            3 => Ok(Connectedness::CannotConnect),
+            v => Err(D::Error::unknown_variant(
+                &v.to_string(),
+                &["0", "1", "2", "3"],
+            )),
+        }
+    }
+}
```
oblique, line 128: I think this can be derived. check other places too

rpc/src/p2p.rs
```diff
@@ -1,17 +1,207 @@
...
-    pub id: String,
-    // TODO: multiaddr
+    pub id: PeerId,
     #[serde(rename = "Addrs")]
-    pub addrs: Vec<String>,
+    pub addrs: Vec<Multiaddr>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Stat {
```
oblique, line 17: type definitions should be in the `type` crate

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,296 @@
...
+
+#[tokio::test]
+async fn info_test() {
+    let client = new_test_client(AuthLevel::Admin).await.unwrap();
+    client.p2p_info().await.expect("Failed to get node info");
+}
+
+#[tokio::test]
+async fn add_remove_peer_test() {
+    //! add and then remove a peer, testing outputs from `p2p.Peers` and `p2p.Connectedness`
```
oblique, line 22: use `//`, the `//!` is a doc-comment that describes the whole file, and they should be only on the beginning of the file. you have it in other places too

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,296 @@
...
+        .p2p_connect(&addr_info)
+        .await
+        .expect("request to connect to second node failed");
+    rpc_call_delay().await;
+
+    assert!(
+        client
+            .p2p_connectedness(&addr_info.id)
+            .await
+            .expect("failed to check connection to peer after connect request")
```
oblique, line 259: save the result in a variable and then use `assert_eq` instead. check other places too

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,296 @@
...
+        .await
+        .expect("failed to spin up second node");
+    log::debug!("{addr_info:?}");
+    let client = new_test_client(AuthLevel::Admin).await.unwrap();
+
+    let initial_peers = client
+        .p2p_peers()
+        .await
+        .expect("failed to get initial peer list");
+    assert!(initial_peers.contains(&addr_info.id) == false);
```
oblique, line 32: use `assert_eq`. check other asserts too

types/src/block.rs
```diff
@@ -270,7 +270,8 @@ mod tests {
             validator_address,
             timestamp,
             ..
-        } = commit.signatures[0].clone() else {
+        } = commit.signatures[0].clone()
+        else {
```
fl0rek, line 274: changes in `types` are outside of scope of the PR, but CI fmt was unhappy about it.

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,292 @@
...
+        client
+            .p2p_is_protected(&addr_info.id, PROTECT_TAG)
+            .await
+            .expect("failed to check protect status after unprotect reqest"),
+        false
+    );
+
+    task.abort();
+}
+#[tokio::test]
```
oblique, line 136: add empty line above

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,292 @@
...
+    let client = new_test_client(AuthLevel::Admin).await.unwrap();
+    client
+        .p2p_bandwidth_stats()
+        .await
+        .expect("failed to get bandwidth stats");
+}
+
+#[tokio::test]
+async fn bandwidth_for_peer_test() {
+    use libp2p::{identity, PeerId};
```
oblique, line 198: move this at the beginning of the file

rpc/Cargo.toml
fl0rek: removed unnecessary dependencies and set version for the remaining `log`

rpc/tests/utils/tiny_node.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        Self {
+            identify,
+            keep_alive: keep_alive::Behaviour,
+        }
+    }
+}
+
+pub async fn start_tiny_node() -> anyhow::Result<(p2p::AddrInfo, JoinHandle<()>)> {
+    async fn get_addresses<T: NetworkBehaviour>(swarm: &mut libp2p::Swarm<T>) -> Vec<Multiaddr> {
```
oblique, line 43: do not define a function within a function

oblique (changes requested):

types/src/block.rs
zvolin: yeah, the rustfmt got updated with let else formatting at 1.72

rpc/src/p2p.rs
fl0rek: done, note that these struct have been moved to `types`.

rpc/tests/p2p.rs
fl0rek: done

rpc/tests/utils/tiny_node.rs
fl0rek: I wanted to keep them together since this function is mostly for code separation, but I don't have strong feelings about moving it. 
Is it particular objection to using nested function here or in general?

rpc/src/p2p.rs
fl0rek: done, note that these struct have been moved to [`types`](https://github.com/eigerco/celestia-node-rs/pull/52/files#diff-ed39cfb7fdc8b39ea00751ee4b418a96975ad12b235bc308b96be3246a2c8686).

rpc/src/p2p.rs
fl0rek: moved all the new types to [`types`](https://github.com/eigerco/celestia-node-rs/pull/52/files#diff-ed39cfb7fdc8b39ea00751ee4b418a96975ad12b235bc308b96be3246a2c8686).

rpc/tests/p2p.rs
fl0rek: switched `assert!` to `assert_eq!` in tests

rpc/tests/p2p.rs
fl0rek: changed

rpc/src/p2p.rs
fl0rek: I added manual implementation because default `serde` one assumes that for enums, json representation is a string/map. Thankfully active community member already created a [`crate`](https://github.com/dtolnay/serde-repr) just for this situation. Hopefully there are no sneaky binary blobs inside :)

Added automatic derivation using [`serde_repr`](https://github.com/dtolnay/serde-repr) crate.

rpc/tests/utils/tiny_node.rs
oblique: In general. It makes reading the code harder and the function look larger than they are.

rpc/tests/p2p.rs
fl0rek: I wanted to do a doc-comment that refers to the outer block (i.e. the test function), but I think I agree it makes sense to switch them to regular comments.

rpc/src/p2p.rs
zvolin: hmm, it's arguable if `celestia-types` is a place for the p2p related celestia types. I was previously more thinking about it as of kind of 'celestia-native' types, like eg. EH. I'm not sure I like that the celestia-types will depend also on the `libp2p`. If those types are needed only for the rpc I like them there, or in future closer to celestia-node if it's gonna mimic the json-rpc server

rpc/Cargo.toml
```diff
@@ -19,3 +22,18 @@ anyhow = "1.0.71"
 dotenvy = "0.15.7"
 rand = "0.8.5"
 tokio = { version = "1.32.0", features = ["rt", "macros"] }
+env_logger = "*"
+log = "*"
```
zvolin: :star:

rpc/Cargo.toml
```diff
@@ -9,6 +9,9 @@ thiserror = "1.0.40"
 jsonrpsee = { version = "0.20", features = ["client-core", "macros"] }
 celestia-types = { workspace = true }
 serde = { version = "1.0.188", features = ["derive"] }
+libp2p = "0.52.3"
```
zvolin: what'd you say to having libp2p and p2p module behind a feature flag? It's probably not so common case, especially guarded with ADMIN permission level, so I think it can be a default but still a feature

rpc/Cargo.toml
```diff
@@ -9,6 +9,9 @@ thiserror = "1.0.40"
...
+libp2p-identity = "0.2.3"
```
zvolin: I'd enable it as a libp2p feature instead to make sure the versions are intact

rpc/Cargo.toml
```diff
@@ -9,6 +9,9 @@ thiserror = "1.0.40"
...
+bs58 = "0.5.0"
```
zvolin: they really use it? :open_mouth:

rpc/tests/utils/tiny_node.rs
zvolin: > Is it particular objection to using nested function here or in general?

yeah, it's rather rare to do so

zvolin:
sry, haven't refreshed for a while, please ignore previous comments as they seem already adressed

rpc/Cargo.toml
fl0rek: yes, looks like I didn't need to import `-identity` directly, but could use it from `libp2p`

rpc/Cargo.toml
fl0rek: yeah :D https://docs.libp2p.io/concepts/fundamentals/peers/#how-are-peer-ids-represented-as-strings

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,273 @@
...
+    let (addr_info, task) = utils::tiny_node::start_tiny_node()
+        .await
+        .expect("failed to spin up second node");
+    let client = new_test_client(AuthLevel::Admin).await.unwrap();
+
+    let initial_peers = client
+        .p2p_peers()
+        .await
+        .expect("failed to get initial peer list");
+    assert_eq!(initial_peers.contains(&addr_info.id), false);
```
zvolin, line 33: for the boolean asserts you should stick with `assert!(...)` and `assert!(!...)`. It's a bug that clippy passed on CI, please fix it with `cargo clippy --all --all-targets` and fix it's suggestions

fl0rek: I had a sneaking suspicion that CI was going too smooth, thanks for noting the clippy bug.
Fixed `bool_assert_comparison` and other clippy errors.

zvolin: I meant to fix the CI too :sweat_smile:

fl0rek: right, didn't realise that would also land in this PR :facepalm: 
added `--all-targets` flag to clippy CI config

zvolin: > would also land in this PR

I personally don't mind it as it's small fix and not in code directly, but if you prefer I'd happily accept another PR or ff merge after some later rebase

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,273 @@
...
+        .expect("Failed to close peer");
+    rpc_call_delay().await;
+
+    let final_peers = client
+        .p2p_peers()
+        .await
+        .expect("failed to get peer list after close peer request");
+    assert!(!final_peers.contains(&addr_info.id));
+
+    task.abort();
```
zvolin, line 71: cleanup aborts inside `tokio::test` macro are redundant as this is handled by the macro, please apply to other too

nevertheless, those are particularly a wrong pattern, you should create some new struct with `Drop` for such cleanup as `drop` fires on panics, where this does not. E.g. if someone would create a code that fails some of the asserts, he won't have the process cleaned up

zvolin: > cleanup aborts inside tokio::test macro are redundant as this is handled by the macro

not sure that's perfectly true tho

zvolin: yeah, so that wasn't true. So non-awaited tasks that keep running after main thread finishes are just killed

zvolin: this seems to be also an issue for our current node architecture, will try to fix that with the traits PR

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,273 @@
...
+    let is_protected_another_tag = client
+        .p2p_is_protected(&addr_info.id, ANOTHER_PROTECT_TAG)
+        .await
+        .expect("failed to check protect status for another tag after protect request");
+    assert!(!is_protected_another_tag);
+
+    client
+        .p2p_unprotect(&addr_info.id, PROTECT_TAG)
+        .await
+        .expect("unprotect request failed");
```
zvolin, line 120: it looks like this also should be followed by the `rpc_call_delay` as it's state changing, even tho it works there for some reason

rpc/src/p2p.rs
```diff
@@ -1,17 +1,61 @@
...
+    async fn p2p_bandwidth_stats(&self) -> Result<BandwidthStats, Error>;
+
+    #[method(name = "p2p.BlockPeer")]
+    async fn p2p_block_peer(&self, peer_id: &PeerId);
+
+    #[method(name = "p2p.ClosePeer")]
+    async fn p2p_close_peer(&self, peer_id: &PeerId);
+
+    #[method(name = "p2p.Connect")]
+    async fn p2p_connect(&self, address: &AddrInfo);
```
zvolin, line 24: Please add docs to those with a warning that errors are not reported and why, with the link to the relevant issue

rpc/src/p2p.rs
```diff
@@ -1,17 +1,61 @@
...
+    async fn p2p_protect(&self, peer_id: &PeerId, tag: &str);
+
+    #[method(name = "p2p.PubSubPeers")]
+    async fn p2p_pub_sub_peers(&self, topic: &str) -> Result<Option<Vec<PeerId>>, Error>;
+
+    #[method(name = "p2p.ResourceState")]
+    async fn p2p_resource_state(&self) -> Result<ResourceManagerStats, Error>;
+
+    #[method(name = "p2p.UnblockPeer")]
+    async fn p2p_unblock_peer(&self, peer_id: &PeerId);
```
zvolin, line 57: also there

rpc/src/p2p.rs
```diff
@@ -1,17 +1,61 @@
...
+
+    #[method(name = "p2p.PeerInfo")]
+    async fn p2p_peer_info(&self, peer_id: &PeerId) -> Result<AddrInfo, Error>;
+
+    #[method(name = "p2p.Peers")]
+    async fn p2p_peers(&self) -> Result<Vec<PeerId>, Error>;
+
+    #[method(name = "p2p.Protect")]
+    async fn p2p_protect(&self, peer_id: &PeerId, tag: &str);
+
```
zvolin, line 49: and there

rpc/src/p2p.rs
zvolin: actually could you try an `Option<Error>` there for fun? :smiling_imp:

rpc/tests/utils/tiny_node.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    swarm.listen_on("/ip4/0.0.0.0/tcp/0".parse()?)?;
+
+    let addrs = wait_for_addresses(&mut swarm).await;
+
+    let task = tokio::task::spawn(async move {
+        loop {
+            swarm.select_next_some().await;
+        }
+    });
```
zvolin, line 88: so we could delete the `wait_for_addresses` function and do sth like this, as we only care about docker's subnet (bridge cannot reach outside easily anyway)
```rust
    // this somewhere at the top
    const DOCKER_SUBNET_RANGE: &str = "172.17.0.";

    let (addr_tx, mut addr_rx) = tokio::sync::mpsc::channel(1);

    tokio::task::spawn(async move {
        let addr_tx = &addr_tx;
        loop {
            if let Some(SwarmEvent::NewListenAddr { address, .. }) = swarm.next().await {
                println!("{address:?}");
                if address.to_string().contains(DOCKER_SUBNET_RANGE) {
                    addr_tx.send(address).await.unwrap();
                }
            }
        }
    });

    let addr = p2p::AddrInfo {
        id: p2p::PeerId(local_peer_id),
        addrs: vec![addr_rx.recv().await.context("Waiting for listener")?],
    };
```

rpc/tests/utils/tiny_node.rs
```diff
@@ -0,0 +1,97 @@
...
+            () = &mut timeout => {
+                break;
+            }
+        }
+    }
+
+    addresses
+}
+
+pub async fn start_tiny_node() -> anyhow::Result<(p2p::AddrInfo, JoinHandle<()>)> {
```
zvolin, line 63: as in other comment, if this just spawns a task, we don't need a handle for it, it will get killed at the end of the test

rpc/tests/utils/tiny_node.rs
zvolin: we could extract it to the eparated fn to get things smaller

types/src/p2p.rs
```diff
@@ -0,0 +1,104 @@
...
+    type Err = PeerIdError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let bytes = bs58::decode(s)
+            .into_vec()
+            .map_err(|_| PeerIdError::Base58Error)?;
+
+        Ok(Self(libp2p::PeerId::from_bytes(&bytes)?))
+    }
+}
```
zvolin, line 78:
```suggestion
impl FromStr for PeerId {
    type Err = <libp2p::PeerId as FromStr>::Err;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        s.parse().map(Self)
    }
}
```

this allows for removing both `PeerIdError` and `bs58` dependency

types/src/p2p.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl Serialize for PeerId {
+    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: ser::Serializer,
+    {
+        s.serialize_str(&self.0.to_base58())
+    }
+}
```
zvolin, line 66: probably not the prettiest, but avoids boilerplate:
```
#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
#[serde(transparent)]
pub struct PeerId(#[serde(with = "tendermint_proto::serializers::from_str")] pub libp2p::PeerId);
```

rpc/tests/utils/tiny_node.rs
```diff
@@ -0,0 +1,97 @@
...
+};
+
+// how long to wait during startup for node to start listening on interfaces, before we return a
+// list of addresses
+const NODE_ADDRESS_ACQUIRE_DELAY_TIME: Duration = Duration::from_millis(100);
+
+/// Our network behaviour.
+#[derive(NetworkBehaviour)]
+struct Behaviour {
+    identify: identify::Behaviour,
```
zvolin, line 26: I think this is unneeded

rpc/src/p2p.rs
fl0rek: Created https://github.com/eigerco/celestia-node-rs/issues/53 for tracking and added a note to each affected method.

As for the option - I've tried that, but `jsonprsee` wants either `Result` or `RpcResult` there:
```
error: The return type must be Result or RpcResult
  --> rpc/src/p2p.rs:19:57
   |
19 |     async fn p2p_block_peer(&self, peer_id: &PeerId) -> Option<()>;

```

types/Cargo.toml
```diff
@@ -4,17 +4,24 @@ version = "0.1.0"
...
 celestia-proto = { workspace = true }
+cid = { version = "0.10.1", default-features = false, features = ["std"] }
 const_format = "0.2.31"
 enum_dispatch = "0.3.12"
-cid = { version = "0.10.1", default-features = false, features = ["std"] }
+libp2p = { version = "0.52.3", optional = true }
 nmt-rs = { workspace = true }
 ruint = { version = "1.8.0", features = ["serde"] }
 serde = { version = "1.0.164", features = ["derive"] }
+serde_repr = "0.1"
```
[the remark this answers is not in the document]
fl0rek: After cleanup only `libp2p` and `serde_repr` are left, marked them both as optional and behind the feature.

rpc/tests/utils/tiny_node.rs
fl0rek: yes, removed

types/src/p2p.rs
fl0rek: replaced this and `Deserialize`/`Serialize`  with serde/tendermint_proto::serializers::from_str suggestion.

types/src/p2p.rs
fl0rek: definitely much simpler, changed

rpc/src/p2p.rs
fl0rek: added note

rpc/src/p2p.rs
fl0rek: added note

rpc/tests/utils/tiny_node.rs
fl0rek: As discussed on stand, we probably cannot assume any specific subnet for docker, since that's configurable (and dynamically allocated, on my machine it's `172.23.0.*`).

I've rolled `wait_for_addresses` back into the `start_tiny_node` and merged the event loops.

types/Cargo.toml
```diff
@@ -4,6 +4,10 @@ version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
 
+[features]
+default = ["libp2p"]
+libp2p = ["dep:libp2p", "dep:serde_repr"]
```
oblique, line 9: move it at the end of file

types/src/p2p.rs
```diff
@@ -0,0 +1,63 @@
...
+use serde_repr::Deserialize_repr;
+use std::collections::HashMap;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AddrInfo {
+    #[serde(rename = "ID")]
+    pub id: PeerId,
+    #[serde(rename = "Addrs")]
+    pub addrs: Vec<Multiaddr>,
+}
```
oblique, line 12: Use `#[serde(rename_all = "PascalCase")]` and have a rename only for `ID`

types/src/p2p.rs
```diff
@@ -0,0 +1,63 @@
...
+    CannotConnect = 3,
+}
+
+#[derive(Debug, PartialEq, Eq, Deserialize_repr)]
+#[repr(u8)]
+pub enum Reachability {
+    Unknown = 0,
+    Public = 1,
+    Private = 2,
+}
```
oblique, line 63: Add serialization too, even if it's not needed for now

rpc/tests/p2p.rs
fl0rek: Connect/disconnect always required delay, so maybe this is just quick enough to usually work. Added delay just in case.

oblique (changes requested):

types/Cargo.toml
```diff
@@ -12,9 +16,11 @@ celestia-proto = { workspace = true }
 const_format = "0.2.31"
 enum_dispatch = "0.3.12"
 cid = { version = "0.10.1", default-features = false, features = ["std"] }
+libp2p = { version = "0.52.3", optional = true }
```
oblique, line 19: use `multiaddr` and `libp2p-identity` crates instead, but still keep them under `libp2p` feature.

rpc/tests/p2p.rs
oblique: in that case i would prefer the `///` doc-comment above the function

rpc/src/p2p.rs
```diff
@@ -1,17 +1,66 @@
...
+    #[method(name = "p2p.Peers")]
+    async fn p2p_peers(&self) -> Result<Vec<PeerId>, Error>;
+
+    /// This method does not report errors due to a workaround to a go-jsonrpc bug, see https://github.com/eigerco/celestia-node-rs/issues/53
+    #[method(name = "p2p.Protect")]
+    async fn p2p_protect(&self, peer_id: &PeerId, tag: &str);
+
+    #[method(name = "p2p.PubSubPeers")]
+    async fn p2p_pub_sub_peers(&self, topic: &str) -> Result<Option<Vec<PeerId>>, Error>;
```
oblique, line 55: In other places that a list was returned you had `Vec<PeerId>` instead of `Option<Vec<PeerId>>`. Remove `Option` to keep it consistent.

rpc/Cargo.toml
```diff
@@ -4,11 +4,16 @@ version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
 
+[features]
+default = ["libp2p"]
+libp2p = ["dep:libp2p", "celestia-types/libp2p"]
```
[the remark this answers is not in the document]
fl0rek: sure, moved

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,273 @@
...
+use libp2p::{identity, PeerId};
+use tokio::time::{sleep, Duration};
+
+pub mod utils;
+
+async fn rpc_call_delay() {
+    // delay for RPC calls like connect/close to let node finish the operation before we query it
+    // again. Below 150 ms I start getting intermittent failures.
+    sleep(Duration::from_millis(150)).await;
+}
```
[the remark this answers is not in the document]
fl0rek: moved

types/Cargo.toml
fl0rek: moved

types/src/p2p.rs
fl0rek: changed

types/src/p2p.rs
fl0rek: added derive for `Serialize_repr`

types/Cargo.toml
fl0rek: replaced `libp2p` with `mutliaddr` and `libp2p-identity` in appropriate version :+1:

oblique (changes requested):

rpc/src/p2p.rs
oblique: i think keeping them in `celestia-types` and feature guarded with `libp2p` is a good solution for now

rpc/src/p2p.rs
fl0rek: This might be another artifact of `go-jsonrpc`, but for topics without subs (or nonsense topics we send in tests) we get `null` in response rather than empty list. This causes deserialisation to fail, if it's not wrapped in `Option`. 
For other calls (peer list, block list, etc.) we get empty list as expected.

rpc/Cargo.toml
```diff
@@ -9,6 +9,7 @@ thiserror = "1.0.40"
 jsonrpsee = { version = "0.20", features = ["client-core", "macros"] }
 celestia-types = { workspace = true }
 serde = { version = "1.0.188", features = ["derive"] }
+libp2p = { version = "0.52.3", optional = true }
```
oblique, line 12: I think this must be moved to `dev-dependencies`

rpc/src/p2p.rs
```diff
@@ -1,17 +1,67 @@
...
+    async fn p2p_peer_info(&self, peer_id: &PeerId) -> Result<AddrInfo, Error>;
+
+    #[method(name = "p2p.Peers")]
+    async fn p2p_peers(&self) -> Result<Vec<PeerId>, Error>;
+
+    /// This method does not report errors due to a workaround to a go-jsonrpc bug, see https://github.com/eigerco/celestia-node-rs/issues/53
+    #[method(name = "p2p.Protect")]
+    async fn p2p_protect(&self, peer_id: &PeerId, tag: &str);
+
+    /// We might get null in response here, so Option is needed
```
oblique, line 54: Make this and other nodes a normal comment. This doesn't need to be in the documentation

rpc/src/p2p.rs
```diff
@@ -1,17 +1,67 @@
...
-    pub id: String,
-    // TODO: multiaddr
-    #[serde(rename = "Addrs")]
-    pub addrs: Vec<String>,
-}
 
 #[rpc(client)]
 pub trait P2P {
+    #[method(name = "p2p.BandwidthForPeer")]
+    async fn p2p_bandwidth_for_peer(&self, peer_id: &PeerId) -> Result<BandwidthStats, Error>;
```
oblique, line 9: If you notice in other modules (for example in blob) we have as doc-comments the [official description](https://node-rpc-docs.celestia.org/#p2p) of each function. Just copy-paste it, no need to adjust anything.

fl0rek: Copied official documentation into doc-comments for appropriate methods.

rpc/src/p2p.rs
fl0rek: moved this and workaround notes to normal code comments.

rpc/Cargo.toml
fl0rek: Moved :+1: 
Note that we cannot have optional dev-dependencies, so it'll always be pulled for tests (which I guess is a good thing, unless we want to feature gate tests too?)

rpc/Cargo.toml
```diff
@@ -17,5 +17,12 @@ jsonrpsee = { version = "0.20", features = ["http-client", "ws-client"] }
...
 dotenvy = "0.15.7"
+futures = "0.3.28"
+libp2p = "0.52.3"
+log = "0.4"
 rand = "0.8.5"
 tokio = { version = "1.32.0", features = ["rt", "macros"] }
+
+[features]
+default = ["libp2p"]
+libp2p = ["celestia-types/libp2p"]
```
oblique, line 28: name it just `p2p`

rpc/tests/p2p.rs
fl0rek: https://docs.rs/tokio/latest/tokio/attr.test.html#current-thread-runtime each test is run in it's own runtime and I see that tasks spawned by test stop when test exists, so removing manual abort.

rpc/Cargo.toml
fl0rek: renamed

oblique (approved):
lgtm

rpc/tests/p2p.rs
```diff
@@ -0,0 +1,267 @@
+use crate::utils::client::{new_test_client, AuthLevel};
```
zvolin, line 1:
```suggestion
#![cfg(feature = "p2p")]
use crate::utils::client::{new_test_client, AuthLevel};
```
to make sure this runs ok:
```
cd rpc && cargo test --no-default-features
```

PR #54, opened 2023-09-07
feat(node): hide all services behind traits

zvolin:


node/src/p2p.rs
```diff
@@ -105,65 +146,34 @@ impl P2p {
...
-        self.send_cmd(P2pCmd::NetworkInfo { respond_to: tx })
-            .await?;
-        rx.await
-            .map_err(|_| P2pError::CmdCancelled("NetworkInfo".to_owned()))
-    }
-}
-
-impl Worker {
     async fn run(&mut self) {
         let mut command_stream = self.cmd_rx.clone().into_stream().fuse();
```
oblique, line 178: I didn't notice this in the previous PR. I think I prefer `self.cmd_rx.recv_async()`

node/src/p2p.rs
```diff
@@ -66,37 +44,100 @@ pub enum P2pError {
...
+
+#[async_trait]
+pub trait P2pService: Service<Args = P2pArgs, Command = P2pCmd, Error = P2pError> {
+    async fn network_info(&self) -> Result<NetworkInfo> {
+        let (tx, rx) = oneshot::channel();
+
+        self.send_command(P2pCmd::NetworkInfo { respond_to: tx })
+            .await?;
+        rx.await
+            .map_err(|_| P2pError::CmdCancelled("NetworkInfo".to_owned()))
```
oblique, line 104: Shouldn't this be `WorkerDied`? I don't see how a command can be cancelled

node/src/p2p.rs
```diff
@@ -66,37 +44,100 @@ pub enum P2pError {
...
+    }
+
+    // async fn subscribe_for_header_sub(&self) -> impl Stream<Item = ExtendedHeader>;
```
oblique, line 107: remove this

node/src/p2p.rs
```diff
@@ -66,37 +44,100 @@ pub enum P2pError {
...
+    pub transport: Boxed<(PeerId, StreamMuxerBox)>,
+    pub network_id: String,
+    pub local_keypair: Keypair,
+    pub bootstrap_peers: Vec<Multiaddr>,
+    pub listen_on: Vec<Multiaddr>,
 }
 
 #[derive(Debug)]
-enum P2pCmd {
+pub enum P2pCmd {
```
oblique, line 61: I think this shouldn't be public because we will implement wrappers for the commands

node/src/syncer.rs
```diff
@@ -1,16 +1,60 @@
...
+    Service<Args = SyncerArgs<P2pSrv>, Command = SyncerCmd, Error = SyncerError>
+{
+}
+
+#[async_trait]
+impl<P2pSrv: P2pService> SyncerService<P2pSrv> for Syncer<P2pSrv> {}
+
+impl<P2pSrv: P2pService> Syncer<P2pSrv> {
+    pub fn new(store: Arc<RwLock<Store>>, p2p: Arc<P2pSrv>) -> Self {
+        Syncer { store, p2p }
```
oblique, line 58: This is not needed because `start` is the constructor

node/src/syncer.rs
```diff
@@ -1,16 +1,60 @@
...
-impl Syncer {
-    pub fn new(store: Arc<RwLock<Store>>) -> Self {
-        Syncer { store }
+pub struct SyncerArgs<P2pSrv: P2pService> {
+    pub p2p: Arc<P2pSrv>,
+    pub store: Arc<RwLock<Store>>,
+}
+
+#[derive(Debug)]
+pub enum SyncerCmd {}
```
oblique, line 26: Same as `P2pCmd`, this should be private

node/src/p2p.rs
zvolin: I couldn't find any info whether `recv_async` is cancel safe tho

node/src/p2p.rs
zvolin: making it private would mean making everything from `Service` trait private too and preventing any external implementations

node/src/p2p.rs
zvolin: removed cmdcancelled entirely

node/src/p2p.rs
oblique: That is true, please use `#[doc(hide)]` for now. If we have an issue in the future we may need to wrap it in a struct and have the field as private.

node/src/p2p.rs
oblique: It is cancel safe but not documented, we can keep the fused stream for now.

node/src/p2p.rs
zvolin: hidden

node/src/syncer.rs
zvolin: hidden

oblique (approved):

fl0rek (approved):
:guitar:

PR #63, opened 2023-09-08
feat(node): Implement exchange client

### LINKED ISSUE #56: Develop Exchange Client

oblique:
Closes #56

### DIFF
--- a/node/src/exchange_client.rs
+++ b/node/src/exchange_client.rs
@@ -4,13 +4,18 @@ use celestia_proto::p2p::pb::{HeaderResponse, StatusCode};
 use celestia_types::ExtendedHeader;
 use libp2p::request_response::{OutboundFailure, RequestId};
 use tendermint_proto::Protobuf;
-use tracing::trace;
+use tracing::{instrument, trace};
 
-use crate::exchange::ExchangeError;
-use crate::utils::{OneshotResultSender, OneshotSenderExt};
+use crate::p2p::P2pError;
+use crate::utils::{OneshotResultSender, OneshotResultSenderExt};
 
 pub(crate) struct ExchangeClientHandler {
-    reqs: HashMap<RequestId, OneshotResultSender<ExtendedHeader, ExchangeError>>,
+    reqs: HashMap<RequestId, ReqInfo>,
+}
+
+struct ReqInfo {
+    amount: u64,
+    respond_to: OneshotResultSender<Vec<ExtendedHeader>, P2pError>,
 }
 
 impl ExchangeClientHandler {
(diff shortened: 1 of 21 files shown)


node/src/p2p.rs
```diff
@@ -189,66 +211,91 @@ impl Worker {
...
+                        .map_err(|_| P2pError::UnsupportedHeaderResponse),
+                };
+
+                trace!("Response: {res:?}");
+
+                let respond_to = self.exchange_reqs.remove(&request_id).unwrap();
+                respond_to.maybe_send(res);
+            }
+
+            exchange::Event::InboundFailure { request_id, .. }
```
zvolin, line 284: iirc InboundFailure is for when we fail to receive request from someone, in case of us making requests only outboundfailure should be considered

node/src/p2p.rs
```diff
@@ -39,6 +43,15 @@ pub enum P2pError {
...
     WorkerDied,
+
+    #[error("Not connected to any peers")]
+    NoPeers,
+
+    #[error("Header not found")]
+    HeaderNotFound,
+
+    #[error("Unsupported header response")]
+    UnsupportedHeaderResponse,
```
zvolin, line 54: maybe more sth like `ExchangeHeaderInvalid`? it combines the exchange protocol and `Invalid` status code

node/src/p2p.rs
```diff
@@ -39,6 +43,15 @@ pub enum P2pError {
 
     #[error("Worker died")]
     WorkerDied,
+
+    #[error("Not connected to any peers")]
+    NoPeers,
+
+    #[error("Header not found")]
+    HeaderNotFound,
```
zvolin, line 51: wdyt `ExchangeHeaderNotFound`?

node/src/p2p.rs
```diff
@@ -39,6 +43,15 @@ pub enum P2pError {
 
     #[error("Worker died")]
     WorkerDied,
+
+    #[error("Not connected to any peers")]
+    NoPeers,
```
zvolin, line 48: wdyt `ExchangeNoPeers`?

node/src/p2p.rs
oblique: sure

node/src/p2p.rs
oblique: sure

node/src/exchange_client.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        respond_to.maybe_send_ok(headers);
+    }
+
+    #[instrument(level = "trace", skip(self))]
+    pub(crate) fn on_failure(&mut self, request_id: RequestId, error: OutboundFailure) {
+        trace!("Outbound failure");
+
+        if let Some(req_info) = self.reqs.remove(&request_id) {
+            // TODO: should we actually report a connection error?
```
zvolin, line 93: yes

node/src/p2p.rs
```diff
@@ -93,12 +128,106 @@ impl Service for P2p {
...
 
         self.send_command(P2pCmd::NetworkInfo { respond_to: tx })
             .await?;
-        rx.await.map_err(|_| P2pError::WorkerDied)
+
+        Ok(rx.await?)
+    }
+
+    async fn exchange_header_request(&self, request: HeaderRequest) -> Result<Vec<ExtendedHeader>> {
+        if request.amount == 0 {
```
zvolin, line 150: doesn't it indicate some strange behaviour? I mean Err there?

node/src/p2p.rs
```diff
@@ -93,12 +128,106 @@ impl Service for P2p {
...
+            data: Some(header_request::Data::Origin(height)),
+            amount: 1,
+        })
+        .await?
+        .into_iter()
+        .next()
+        .ok_or(P2pError::ExchangeHeaderNotFound)
+    }
+
+    async fn get_verified_header_range(
```
zvolin, line 206:
```suggestion
    async fn get_verified_headers_range(
```

node/src/p2p.rs
```diff
@@ -190,65 +330,64 @@ impl Worker {
...
+                BehaviourEvent::HeaderEx(_) | BehaviourEvent::KeepAlive(_) => {}
             },
+            SwarmEvent::ConnectionEstablished { peer_id, .. } => {
+                self.peer_tracker.add(peer_id);
+
+                // This vector is not going to be used later on so deallocate
+                // its internal memory by consuming it with mem::take.
+                for tx in mem::take(&mut self.wait_connected_tx) {
+                    tx.maybe_send(());
+                }
```
zvolin, line 348: I'd not use `VecDeque` and `mem::take` for `wait_connected_tx`. You could use the `Vec`, replace `push_back` with `push` and here use the `.drain(..)`

oblique: I forgot to change it to `Vec`. However even in case of `Vec`, `mem::take` is needed because `.drain(..)` does not shrink the capacity. This is an one-off queue.

zvolin (approved):
left few nits but this looks really good

node/src/p2p.rs
zvolin: if it is a one-off queue, then keep this `Vec` as an option. otherwise it's not, if we disconnect from all the peers (network error or whatever) then it'll again behave like so

node/src/p2p.rs
oblique: Like `InvalidAmount`?

node/tests/node.rs
```diff
@@ -54,9 +52,60 @@ async fn connects_to_the_go_bridge_node() {
...
+    let genesis = node.p2p().get_header_by_height(1).await.unwrap();
+    let head = node.p2p().get_head_header().await.unwrap();
+
+    genesis.verify(&head).unwrap();
+}
+
+#[tokio::test]
+async fn invalid_height() {
+    let node = new_connected_node().await;
+    node.p2p().get_header_by_height(0).await.unwrap_err();
```
zvolin, line 110: I'm not sure if that behaviour is valid (by valid I mean consistent with celestia). I think
```
celestia rpc header GetByHeight 0
```
would mimic `Head` request there

oblique: Ok I will check, but I don't like mimicking the `Head`.

node/src/p2p.rs
zvolin: or just InvalidRequest?

zvolin: I just think if someone is making such request, should we indicate that or rather allow them?

oblique: Celestia implementation does not return anything if we allow them and we get `OutboundFailure::Timeout` later on. So I prefer to either return empty Vec or error. Both are valid for me.

node/tests/node.rs
zvolin: okay, it's likely it won't mimic head, but I guess it'll just respond with the node's own local head

node/src/exchange_client.rs
oblique: In case of `OutboundFailure::Timeout` should I return `Err(ExchangeHeaderNotFound)` or `Err(ExchangeHeaderOutboundFailure(error))`?

fl0rek (approved):

zvolin (approved):

PR #64, opened 2023-09-11
feat(exchange): Add pre-allocating the buffer for reading in HeaderCodec

### LINKED ISSUE #43: utilize `prost::decode_length_delimiter` and `prost::length_delimiter_len` to pre-alloc when decoding messages in Exchange
We could try to utilize `prost::decode_length_delimiter` and `prost::length_delimiter_len` to just pre-alloc the needed vector that we'll read to.
Commenting just to open the issue

_Originally posted by @zvolin in https://github.com/eigerco/celestia-node-rs/pull/42#discussion_r1315777857_

fl0rek:
closes #43 

Read up to first 10 bytes (max length of the length delimiter varint) in a loop until prost::decode_length_delimiter is able to figure out message length. Then, when reading, preallocate the buffer and read the exact number of bytes needed.


oblique:
`read_response` should implement the same mechanism. I suggest implementing a generic `read_raw_message` that returns the raw bytes and then use it in both.

node/src/exchange.rs
```diff
@@ -41,11 +44,49 @@ impl Codec for HeaderCodec {
...
+            };
+
+            if let Ok(len) = prost::decode_length_delimiter(&buf[..read]) {
+                break len;
+            }
+        };
+
+        // const value conversion, safe for 32bit and larger usize
+        if len
+            > REQUEST_SIZE_MAXIMUM
```
zvolin, line 69: wouldn't it be better to just define those constants as usize?

fl0rek: We're the only users of this const, so actually yes. Changed :+1:

zvolin (approved):
looks good.

Recently we've been discussing the [`proptest`](https://crates.io/crates/proptest) as guys in sovereign seems using this alot and this looks like a perfect use for it. We could add fuzz tests for the codec but I leave it up to you guys, @fl0rek @oblique

node/src/exchange.rs
```diff
@@ -99,17 +96,15 @@ impl Codec for HeaderCodec {
     where
         T: AsyncRead + Unpin + Send,
     {
-        HeaderCodec::read_raw_messages(io, REQUEST_SIZE_MAXIMUM)
+        let mut buf = Vec::with_capacity(512);
+        buf.resize(PROTOBUF_MAX_LENGTH_DELIMITER_LEN, 0);
```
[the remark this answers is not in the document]
oblique: hmm actually there is an edge case when using clear here, so don't do it. lets discuss it tomorrow.

node/src/exchange.rs
```diff
@@ -31,21 +34,78 @@ pub fn new_behaviour(network: &str) -> Behaviour {
...
+                            ReadHeaderError::StreamClosed,
+                        ));
+                    }
+                }
+                n => read += n,
+            };
+
+            if let Ok(len) = prost::decode_length_delimiter(&buf[..read]) {
+                break len;
+            }
```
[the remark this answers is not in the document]
fl0rek: Right, this was incorrectly mapped to `StreamClosed` since we were trying to `io.read` to zero length buffer.
Fixed and added `ReadHeaderError::VarintOverflow`.

node/src/exchange.rs
```diff
@@ -31,21 +34,99 @@ pub fn new_behaviour(network: &str) -> Behaviour {
...
+
+            if read >= PROTOBUF_MAX_LENGTH_DELIMITER_LEN {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    ReadHeaderError::VarintOverflow,
+                ));
+            }
+
+            buf.resize(PROTOBUF_MAX_LENGTH_DELIMITER_LEN, 0);
```
oblique, line 60: this should be done only if `buf.len()` is less than `PROTOBUF_MAX_LENGTH_DELIMITER_LEN`, otherwise it will discard data.

```suggestion
            if buf.len() < PROTOBUF_MAX_LENGTH_DELIMITER_LEN {
                buf.resize(PROTOBUF_MAX_LENGTH_DELIMITER_LEN, 0);
            }
```

node/src/exchange.rs
```diff
@@ -31,21 +34,99 @@ pub fn new_behaviour(network: &str) -> Behaviour {
...
-
-        Ok(HeaderRequest::decode_length_delimited(&vec[..])?)
+        let mut buf = Vec::with_capacity(512);
+        HeaderCodec::read_raw_message(io, &mut buf, REQUEST_SIZE_MAXIMUM)
+            .await
+            .and_then(|maybe_msg| {
+                maybe_msg.ok_or_else(|| {
+                    io::Error::new(io::ErrorKind::UnexpectedEof, ReadHeaderError::StreamClosed)
+                })
+            })
```
oblique, line 127:
```suggestion

        HeaderCodec::read_raw_message(io, &mut buf, REQUEST_SIZE_MAXIMUM)
            .await?
            .ok_or_else(|| {
                io::Error::new(io::ErrorKind::UnexpectedEof, ReadHeaderError::StreamClosed)
            })
```

node/src/exchange.rs
```diff
@@ -83,15 +171,356 @@ impl Codec for HeaderCodec {
...
     ) -> io::Result<()>
     where
         T: AsyncWrite + Unpin + Send,
     {
-        let data = resp.encode_length_delimited_to_vec();
+        for resp in resps {
+            let data = resp.encode_length_delimited_to_vec();
 
-        io.write_all(data.as_ref()).await?;
+            io.write_all(data.as_ref()).await?;
```
oblique, line 180:
```suggestion
            io.write_all(&data).await?;
```

node/src/exchange.rs
oblique: this can be removed now because of the buf resizing before the loop

fl0rek: For previous implementation I believe it held that we had at most `PROTOBUF_MAX_LENGTH_DELIMITER_LEN` bytes cached, since after varint read (to 10byte buffer), we either had complete message and passed the remainder to next iteration; or we didn't and we read exact number of bytes needed, giving us empty buffer on next invocation.
With resizing to 512 before the loop whole thing changes though, so removing.

node/src/exchange.rs
```diff
@@ -31,21 +34,101 @@ pub fn new_behaviour(network: &str) -> Behaviour {
 #[derive(Clone, Copy, Debug, Default)]
 pub struct HeaderCodec;
 
+impl HeaderCodec {
+    async fn read_raw_message<T, R>(
```
oblique, line 38: single we return a decoded message, the `raw` is a misleading. I suggest this:

```rust
fn read_raw_message<R, T>(reader: &mut R, buf: &mut Vec<u8>, max_len: usize) -> io::Result<Option<T>>
```

node/src/exchange.rs
```diff
@@ -83,15 +173,356 @@ impl Codec for HeaderCodec {
...
+
+#[derive(Debug, thiserror::Error, PartialEq)]
+pub enum ReadHeaderError {
+    #[error("stream closed while trying to get header length")]
+    StreamClosed,
+    #[error("varint overflow")]
+    VarintOverflow,
+    #[error("request too large: {0}")]
+    ResponseTooLarge(usize),
+}
```
oblique, line 199: Move this after `pub struct HeaderCodec;`

node/src/exchange.rs
```diff
@@ -83,15 +173,356 @@ impl Codec for HeaderCodec {
...
+            .get_ref()
+            .unwrap()
+            .downcast_ref::<ReadHeaderError>()
+            .unwrap();
+        assert_eq!(inner_err, &ReadHeaderError::VarintOverflow);
+    }
+
+    #[tokio::test]
+    async fn test_decode_header_double_response_data() {
+        let mut header_response_buffer = bytes::BytesMut::with_capacity(512);
```
oblique, line 357: replace all `bytes::BytesMut` with `BytesMut`

node/src/exchange.rs
```diff
@@ -83,15 +173,356 @@ impl Codec for HeaderCodec {
...
+        ) -> Poll<Result<usize, Error>> {
+            let mut small_buf = [0u8; CHUNK_SIZE];
+            match Pin::new(&mut self.inner).poll_read(cx, &mut small_buf) {
+                Poll::Ready(Ok(read)) => {
+                    buf[..read].copy_from_slice(&small_buf[..read]);
+                    Poll::Ready(Ok(read))
+                }
+                ret @ Poll::Ready(Err(_)) => ret,
+                Poll::Pending => Poll::Pending,
+            }
```
oblique, line 525: Intermediate buffer is not needed. You can do this instead:

```suggestion
            let len = buf.len().min(CHUNK_SIZE);
            Pin::new(&mut self.inner).poll_read(cx, &mut buf[..len])
```

node/src/p2p.rs
```diff
@@ -232,13 +232,15 @@ impl Worker {
...
-                debug!("Header: {header:?}");
+                for r in response {
+                    println!(
+                        "Response for request: {request_id}, from peer: {peer}, status: {:?}",
+                        r.status_code()
+                    );
+                    let header = ExtendedHeader::decode(&r.body[..]).unwrap();
+                    // TODO: Forward response back with one shot channel
+                    println!("Header: {header:?}");
+                }
```
oblique, line 243: is this a leftover?

fl0rek: yes, fixed `println`->`debug`

node/src/exchange.rs
fl0rek: nice :rocket: 

changed

node/src/exchange.rs
fl0rek: replaced with `BytesMut`

node/src/exchange.rs
```diff
@@ -83,15 +173,356 @@ impl Codec for HeaderCodec {
...
+    #[tokio::test]
+    async fn test_decode_multiple_small_header_response() {
+        const MSG_COUNT: usize = 10;
+        let header_response = HeaderResponse {
+            body: vec![1, 2, 3],
+            status_code: 1,
+        };
+
+        let encoded_header_response = header_response.encode_length_delimited_to_vec();
+        //let mut reader = Cursor::new(encoded_header_response);
```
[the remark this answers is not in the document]
fl0rek: removed

node/src/exchange.rs
fl0rek: moved :+1:

node/src/exchange.rs
```diff
@@ -31,21 +34,112 @@ pub fn new_behaviour(network: &str) -> Behaviour {
...
+
+impl HeaderCodec {
+    async fn read_message<T, R>(
+        io: &mut T,
+        buf: &mut Vec<u8>,
+        max_len: usize,
+    ) -> io::Result<Option<R>>
+    where
+        T: AsyncRead + Unpin + Send,
+        R: Message + Default,
```
oblique, line 55: It makes more sense to have `R` for the `io` and `T` on the return

node/src/exchange.rs
```diff
@@ -31,21 +34,101 @@ pub fn new_behaviour(network: &str) -> Behaviour {
...
     type Request = HeaderRequest;
-    type Response = HeaderResponse;
+    type Response = Vec<HeaderResponse>;
 
     async fn read_request<T>(&mut self, _: &Self::Protocol, io: &mut T) -> io::Result<Self::Request>
     where
         T: AsyncRead + Unpin + Send,
     {
-        let mut vec = Vec::new();
+        let mut buf = Vec::with_capacity(512);
```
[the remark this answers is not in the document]
fl0rek: right, removed

node/src/exchange.rs
```diff
@@ -31,21 +34,101 @@ pub fn new_behaviour(network: &str) -> Behaviour {
...
+                }
+                n => read += n,
+            };
+        };
+
+        // truncate buffer to the data that was actually read
+        buf.truncate(read);
+
+        let length_delimiter_len = length_delimiter_len(data_len);
+        let single_message_length = length_delimiter_len + data_len;
```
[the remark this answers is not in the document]
fl0rek: renamed

node/Cargo.toml
```diff
@@ -10,6 +10,7 @@ celestia-types = { workspace = true }
 tendermint-proto = { workspace = true }
 
 async-trait = "0.1.73"
+bytes = "1.4.0"
```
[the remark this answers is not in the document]
fl0rek: moved :+1:

node/src/exchange.rs
fl0rek: renamed the fn and generic types to match

oblique (approved):
lgtm

PR #65, opened 2023-09-13
chore: align to celestia node v0.11-rc

zvolin:


ci/Dockerfile.bridge
```diff
@@ -8,8 +8,9 @@ ENV CELESTIA_HOME=/root
 RUN apk update && apk add --no-cache bash jq
 
 # Copy in the binary
-COPY --from=ghcr.io/celestiaorg/celestia-node:4b96022 /bin/celestia /bin/celestia
-COPY --from=ghcr.io/celestiaorg/celestia-node:4b96022 /bin/cel-key /bin/cel-key
+# v0.11.0-rc13
+COPY --from=ghcr.io/celestiaorg/celestia-node:8ca777e /bin/celestia /bin/celestia
+COPY --from=ghcr.io/celestiaorg/celestia-node:8ca777e /bin/cel-key /bin/cel-key
```
oblique, line 13: we can actually use the tag

```suggestion
COPY --from=ghcr.io/celestiaorg/celestia-node:v0.11.0-rc13 /bin/celestia /bin/celestia
COPY --from=ghcr.io/celestiaorg/celestia-node:v0.11.0-rc13 /bin/cel-key /bin/cel-key
```

ci/Dockerfile.validator
```diff
@@ -8,7 +8,8 @@ ENV CELESTIA_HOME=/root
 RUN apk update && apk add --no-cache bash jq
 
 # Copy in the binary
-COPY --from=ghcr.io/celestiaorg/celestia-app:02f37a3 /bin/celestia-appd /bin/celestia-appd
+# v1.0.0-rc14
+COPY --from=ghcr.io/celestiaorg/celestia-app:cdc6bff /bin/celestia-appd /bin/celestia-appd
```
oblique, line 12:
```suggestion
COPY --from=ghcr.io/celestiaorg/celestia-app:v1.0.0-rc14 /bin/celestia-appd /bin/celestia-appd
```

types/src/blob.rs
```diff
@@ -9,6 +9,15 @@ use crate::consts::appconsts;
 use crate::nmt::Namespace;
 use crate::{bail_validation, Error, Result};
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct SubmitOptions {
+    // TODO: giving negative fees always results in insufficient fee err, but celestia
+    //       have it like so, should we do a u64 there?
```
fl0rek, line 16: I see `int64` there, is celestia able to correctly reject `i64::MAX..u64::MAX` values? If so I don't think there's harm to it

ci/Dockerfile.bridge
zvolin: I don't think they tag images but will check

zvolin: indeed, thanks

types/src/blob.rs
oblique: Actually in other places we used `Uint` for fees and it was serialized as numeric strings. What celestia does here? I prefer `u64` if we can not use `Uint` and have a `serialize_with` that checks if the number is bigger than `i64`. Or maybe open a bug to them.

oblique: actually having `u64` at let the server return an error is more acceptable for me than `i64`.

zvolin: unfortunately requests with value in this range just fails, I'll stick to the exact types celestia is using and delete this todo

fl0rek (approved):
:guitar:

oblique (approved):

PR #67, opened 2023-09-14
feat(node/exchange): Request HEAD from multiple peers and choose the best result

oblique:


node/src/exchange/client.rs
```diff
@@ -58,13 +76,79 @@ impl ExchangeClientHandler {
...
+
+            // Otherwise return the header with the maxium height
+            respond_to.maybe_send_ok(vec![resps[0].to_owned()]);
+        });
     }
 
     #[instrument(level = "trace", skip(self, responses), fields(responses.len = responses.len()))]
     pub(super) fn on_response_received(
         &mut self,
+        peer: PeerId,
```
zvolin, line 151: this peer_id is tricky as it is only used by the `instrument` macro. wdyt about moving it to the end. prefixing with an underscore and renaming back to `peer` in instrument macro?

zvolin: applies to all places

node/src/peer_tracker.rs
```diff
@@ -1,47 +1,52 @@
...
+    }
+
+    pub fn best_n_peers(&self, limit: usize) -> Option<Vec<PeerId>> {
+        // TODO: Implement peer score and return the best N peers.
+        let peers = self
+            .peers
+            .iter()
+            .take(limit)
+            .map(|v| v.key().to_owned())
+            .collect::<Vec<_>>();
```
zvolin, line 43:
```suggestion
            // collect instead of returning an iter to not block the dashmap
            .collect::<Vec<_>>();
```

node/src/exchange/client.rs
```diff
@@ -58,13 +76,79 @@ impl ExchangeClientHandler {
 
         let req_id = req_resp.send_request(&peer, request);
         self.reqs.insert(req_id, ReqInfo { amount, respond_to });
+    }
 
-        trace!("Request initiated");
+    fn send_head_request(
```
zvolin, line 81: is there a way to test that?

oblique: not, currently. we need to implement peer finding and connect to a network with multiple peers.

oblique: actually i thought a way, i will implement it tomorrow

node/src/exchange/client.rs
oblique: For now it is used only instrument. But the reason I added it was for `PeerTracker` scoring later on.

node/src/exchange/client.rs
oblique: tests are added

node/src/peer_tracker.rs
```diff
@@ -1,47 +1,64 @@
...
     }
 
     pub fn best_peer(&self) -> Option<PeerId> {
         // TODO: Implement peer score and return the best.
-        let inner = self.inner.read().unwrap();
-        inner.peers.iter().next().copied()
+        self.peers.iter().next().map(|v| v.key().to_owned())
+    }
+
+    pub fn best_n_peers(&self, limit: usize) -> Option<Vec<PeerId>> {
```
fl0rek, line 48: Having `Option<Vec>`  only to never return empty vec feels a bit weird to me, empty is empty :P 
Not feeling particularly strongly about it though

node/src/exchange/client.rs
```diff
@@ -58,14 +97,80 @@ impl ExchangeClientHandler {
...
+                respond_to: tx,
+            };
+
+            self.reqs.insert(req_id, req_info);
+            rxs.push(rx);
+        }
+
+        // Choose the best HEAD.
+        //
+        // Algorithm: https://github.com/celestiaorg/go-header/blob/e50090545cc7e049d2f965d2b5c773eaa4a2c0b2/p2p/exchange.go#L357-L381
```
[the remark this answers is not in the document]
oblique: `request_response::Behaviour` has [default timeout of 10 seconds](https://github.com/libp2p/rust-libp2p/blob/2c3d6ad6da8663f0bc5751138aa163988fadd8ab/protocols/request-response/src/lib.rs#L290-L297), so we are ok. We many need to check how many seconds Celestia has defined

node/src/peer_tracker.rs
oblique: Well, it works nicely with let-else:

```rust

        let Some(peers) = self.peer_tracker.best_n_peers(MAX_PEERS) else {
            respond_to.maybe_send_err(P2pError::NoPeers);
            return;
        };
```

node/src/exchange/client.rs
```diff
@@ -80,17 +185,12 @@ impl ExchangeClientHandler {
             return;
         }
 
+        // TODO: should we make sure tahat range or the hash is matched to the requested ones?
```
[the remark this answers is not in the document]
oblique: added validation anyway

node/src/peer_tracker.rs
fl0rek: I was thinking about vec destructuring, but indeed there doesn't seem to be equivalentely concise syntax to `let .. else`.
`PeerTracker` seems to be part of public API? My worry is that having `Option<Vec>` suggests there's a difference between no-result and empty-result. We can weight it against convenience and I think I can live with both :P

fl0rek (approved):
nice work! 🎸

node/src/peer_tracker.rs
oblique: I don't have strong opinion tbh. I can change it to just Vec.

zvolin: me neither, I like the concise usage (which is also a part of public api, so if we want users to use it like that then maybe add some docstring with this exact example and we good?

node/src/exchange/client.rs
zvolin: yeah, I just started scanning the code to look why is it needed and it's not really obvious that instrument uses that, so wanted to indicate that fact before we implement the actual usage, but it's a nitpick anyway so up to you

node/src/exchange/client.rs
```diff
@@ -34,7 +61,7 @@ impl ExchangeClientHandler {
     #[instrument(level = "trace", skip(self, req_resp, respond_to))]
     pub(super) fn on_send_request(
         &mut self,
-        req_resp: &mut ReqRespBehaviour,
+        req_resp: &mut R,
```
zvolin, line 64: please change all args `req_resp` to just `req`

zvolin: I don't like the name clash with request tho :thinking:

zvolin: maybe to `sender` then?

node/src/peer_tracker.rs
oblique: I prefer a simple Vec instead. I will change it later

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
 
-            match res {
-                Ok(header) => {
-                    trace!("Header: {header:?}");
-                    headers.push(header);
+        match (&request.data, request.amount) {
+            (_, 0) => return Err(ExchangeError::InvalidResponse),
+
+            // Allow HEAD requests to have any height in their response
+            (Some(Data::Origin(0)), _) => {}
```
zvolin, line 228:
```suggestion
            // Allow HEAD requests to have any height in their response
            (Some(Data::Origin(0)), 1) => {}
```
Shouldn't we match on `1` there too? I know it's validated on sending side but just for case

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
+                    if header.height().value() != height {
+                        return Err(ExchangeError::InvalidResponse);
+                    }
                 }
-                Err(e) => {
-                    respond_to.maybe_send_err(e);
-                    return;
+            }
+
+            // Check if header has the reuqested hash
```
zvolin, line 239:
```suggestion
            // Check if header has the requested hash
```

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
+    use celestia_types::{DataAvailabilityHeader, Hash, ValidatorSet};
+    use std::collections::VecDeque;
+    use std::sync::atomic::{AtomicU64, Ordering};
+    use tendermint::block::header::Version;
+    use tendermint::block::{Commit, Header, Id};
+    use tendermint::hash::AppHash;
+    use tendermint::Time;
+
+    #[tokio::test]
+    async fn request_heigh() {
```
zvolin, line 280:
```suggestion
    async fn request_height() {
```

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
+    impl MockReqId {
+        fn new() -> MockReqId {
+            static NEXT: AtomicU64 = AtomicU64::new(0);
+            let id = NEXT.fetch_add(1, Ordering::Relaxed);
+            MockReqId(id)
+        }
+    }
+
+    struct MockReq {
+        reqs: VecDeque<MockReqInfo>,
```
zvolin, line 693: I'm curious what's the point of `VecDeque`? is drain here more efficient?

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
+            },
+            tx,
+        );
+        assert!(matches!(
+            rx.await,
+            Ok(Err(P2pError::Exchange(ExchangeError::InvalidRequest)))
+        ));
+    }
+
+    /// Expects the highest heigh that was reported by at least 2 peers
```
zvolin, line 576:
```suggestion
    /// Expects the highest height that was reported by at least 2 peers
```

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
+
+        handler.on_send_request(&mut mock_req, HeaderRequest::with_origin(0, 1), tx);
+
+        let expected = gen_header_response(5);
+        let expected_header = expected.to_extended_header().unwrap();
+
+        mock_req.send_n_responses(&mut handler, 1, vec![gen_header_response(3)]);
+        mock_req.send_n_responses(&mut handler, 2, vec![gen_header_response(4)]);
+        mock_req.send_n_responses(&mut handler, 1, vec![gen_header_response(5)]);
+        mock_req.send_n_responses(&mut handler, 2, vec![expected]);
```
zvolin, line 593: not sure if that matters, but we actually send height 5 three times here

node/src/exchange/client.rs
zvolin: I'm again having feelings with this `Request` trait name, it's confusing that it's used just next to `Request` which is just a data structure. If you don't mind I find `SendReq` or `ReqSender` much less confusing

zvolin: @fl0rek wdyt?

zvolin: maybe just `Sender`?

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
+
+        mock_req.send_n_responses(&mut handler, 1, vec![expected]);
+
+        let result = rx.await.unwrap().unwrap();
+        assert_eq!(result.len(), 1);
+        assert_eq!(result[0], expected_header);
+    }
+
+    #[tokio::test]
+    async fn request_range() {
```
zvolin, line 323: It'd be nice to also have a test that range requests sort the returned headers by height, or is it implementation detail?

node/src/exchange/client.rs
oblique: changed to `Sender`

node/src/exchange/client.rs
oblique: fixed

node/src/exchange/client.rs
oblique: Yes. `Vec::drain` moves the remaining items at the beginning of the allocated memory.

zvolin: woow, that's a game changer, thanks!

node/src/exchange/client.rs
oblique: It was on purpose. The first one has different hash than the `expected`. Algorithm should choose the `expected`.

node/src/exchange/client.rs
oblique: The current Celestia implementation returns the results sorted. But I decided to not trust it. I can add a test if you want.

node/src/exchange/client.rs
```diff
@@ -56,63 +96,697 @@ impl ExchangeClientHandler {
...
+        }
+
+        // Choose the best HEAD.
+        //
+        // Algorithm: https://github.com/celestiaorg/go-header/blob/e50090545cc7e049d2f965d2b5c773eaa4a2c0b2/p2p/exchange.go#L357-L381
+        spawn(async move {
+            let mut resps: Vec<_> = join_all(rxs)
+                .await
+                .into_iter()
+                .filter_map(|v| v.ok()?.ok()?.into_iter().next())
```
zvolin, line 146: this get's me thinking, shouldn't we there punish peers responding with multiple requests rather than ignoring it? I think in p2p the more defensive code the better. if you agree please also add a test

node/src/exchange/client.rs
zvolin: yeah it's probably because `Store` expects it to be sorted. I think the better place for that check would be in the `Store` itself. It can mean double sorting, but it is fine if we stick to the default `pdqsort` as it's O(n) for ascending / descending sequences

node/src/exchange/client.rs
zvolin: :100: 
Can we have a comment so I won't miss that again?

```suggestion
        // this header also has height = 5 but has different hash
        mock_req.send_n_responses(&mut handler, 1, vec![gen_header_response(5)]);
        mock_req.send_n_responses(&mut handler, 2, vec![expected]);
```

node/src/exchange/client.rs
oblique: At this point everything will be of length 1 because it is validated in `decode_and_verify_responses` method

node/src/exchange/client.rs
```diff
@@ -153,14 +153,14 @@ where
...
-            // Sort by height in descending order
-            resps.sort_by_key(|resp| Reverse(resp.height()));
-
             // Count peers per response
             for resp in &resps {
                 *counter.entry(resp.hash()).or_default() += 1;
             }
 
+            // Sort by height and then peers in descending order
+            resps.sort_by_key(|resp| Reverse((resp.height(), counter[&resp.hash()])));
```
zvolin, line 162: cool, never seen

node/src/exchange/client.rs
oblique: added a test just to ensure

node/src/exchange/client.rs
oblique: added a test for this

node/src/exchange/client.rs
```diff
@@ -56,63 +96,821 @@ impl ExchangeClientHandler {
...
+            // In that case propagate the decoded error.
+            if responses.len() == 1 {
+                responses[0].to_extended_header()?;
+            }
+
+            // Otherwise report it as InvalidResponse
+            //
+            // TODO: Should we relax this? Server can return a sub-range.
+            return Err(ExchangeError::InvalidResponse);
         }
```
oblique, line 228: This is wrong and should be relaxed. This is what Celestia is doing:

1. Send request for the range (e.g: start = 10, amount = 10)
2. Receives a sub-range response (e.g. [10, 14])
3. Accepts the partial response and submits a new one for the remaining range (e.g. start = 15, amount = 5)

zvolin: isn't it what we should do in https://github.com/eigerco/celestia-node-rs/issues/62?

oblique: Yes. I relaxed the check for now and a partial range will be returned. If you prefer the more strict version let me know.

oblique: Actually I changed my mind. I reverted the relaxed version because it changes the behavior and may affect other logic later on. We should guarantee that response of `get_range_*` should be the whole range or an error.

zvolin (approved):

fl0rek (approved):
🎸

PR #68, opened 2023-09-15
Fix docker for apple silicon macs

fl0rek:
Explicitly set platform in docker compose, otherwise docker tries to start it as linux/arm64 and musl is unhappy. https://stackoverflow.com/questions/71040681/qemu-x86-64-could-not-open-lib64-ld-linux-x86-64-so-2-no-such-file-or-direc

`upper=${lower^^}` syntax exists from bash 4.0, macos has 3.2. Replace it with `tr`. https://unix.stackexchange.com/questions/51983/how-to-uppercase-the-command-line-argument

Mac's sed is more BSD-like and ~~`-i` suffix cannot be ommited when empty.~~ there's no command that works for both GNU and Mac's BSD-like *and* doesn't create a backup file. 
  https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

This all should be portable enough to work on linux and recent macs .

### DIFF
--- a/tools/gen_auth_tokens.sh
+++ b/tools/gen_auth_tokens.sh
@@ -27,12 +27,16 @@ generate_token() {
 }
 
 write_token() {
-  local auth_level=$(echo "$1" | tr '[:lower:]' '[:upper:]')
+  local auth_level="$1" $(echo "$1" | tr '[:lower:]' '[:upper:]')
   local token="$2"
+
+  local auth_level=$(echo "$auth_level" | tr '[:lower:]' '[:upper:]')
 
   local var_name="CELESTIA_NODE_AUTH_TOKEN_${auth_level}"
 
-  sed -i'' -e "s/.*$var_name.*/$var_name=$token/" "$DOTENV"
+  sed -i '.bak' -e "s/.*$var_name.*/$var_name=$token/" "$DOTENV"
+  rm "$DOTENV.bak" # there's no compatible way to tell sed not to do a backup file
+                   # accept it and remove the file afterwards
 }
 
 main() {
--- a/tools/gen_auth_tokens.sh
+++ b/tools/gen_auth_tokens.sh
@@ -27,16 +27,14 @@ generate_token() {
 }
 
 write_token() {
-  local auth_level="$1" $(echo "$1" | tr '[:lower:]' '[:upper:]')
+  local auth_level="$1"
   local token="$2"
 
   local auth_level=$(echo "$auth_level" | tr '[:lower:]' '[:upper:]')
 
   local var_name="CELESTIA_NODE_AUTH_TOKEN_${auth_level}"
 
-  sed -i '.bak' -e "s/.*$var_name.*/$var_name=$token/" "$DOTENV"
-  rm "$DOTENV.bak" # there's no compatible way to tell sed not to do a backup file
-                   # accept it and remove the file afterwards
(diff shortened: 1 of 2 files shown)


tools/gen_auth_tokens.sh
```diff
@@ -27,12 +27,12 @@ generate_token() {
 }
 
 write_token() {
-  local auth_level="$1"
+  local auth_level=$(echo "$1" | tr '[:lower:]' '[:upper:]')
```
zvolin, line 30: please apply this conversion separately from `local` variable declaration

tools/gen_auth_tokens.sh
```diff
@@ -27,12 +27,16 @@ generate_token() {
 }
 
 write_token() {
-  local auth_level="$1"
+  local auth_level="$1" $(echo "$1" | tr '[:lower:]' '[:upper:]')
```
oblique, line 30: this looks like a leftover

tools/gen_auth_tokens.sh
```diff
@@ -27,12 +27,16 @@ generate_token() {
...
   local token="$2"
 
-  local var_name="CELESTIA_NODE_AUTH_TOKEN_${auth_level^^}"
+  local auth_level=$(echo "$auth_level" | tr '[:lower:]' '[:upper:]')
```
oblique, line 33:
```suggestion
  local auth_level="$(echo "$auth_level" | tr '[:lower:]' '[:upper:]')"
```

tools/gen_auth_tokens.sh
```diff
@@ -27,12 +27,16 @@ generate_token() {
...
+
+  local var_name="CELESTIA_NODE_AUTH_TOKEN_${auth_level}"
 
-  sed -i "s/.*$var_name.*/$var_name=$token/" "$DOTENV"
+  sed -i '.bak' -e "s/.*$var_name.*/$var_name=$token/" "$DOTENV"
+  rm "$DOTENV.bak" # there's no compatible way to tell sed not to do a backup file
+                   # accept it and remove the file afterwards
```
oblique, line 39: I prefer the comments above the line

```suggestion
  # there's no compatible way to tell sed not to do a backup file
  # accept it and remove the file afterwards
  rm "$DOTENV.bak" 
```

tools/gen_auth_tokens.sh
zvolin: also please remove second local

tools/gen_auth_tokens.sh
```diff
@@ -30,9 +30,11 @@ write_token() {
   local auth_level="$1"
   local token="$2"
 
-  local var_name="CELESTIA_NODE_AUTH_TOKEN_${auth_level^^}"
+  auth_level="$(echo "$auth_level" | tr '[:lower:]' '[:upper:]')"
 
-  sed -i "s/.*$var_name.*/$var_name=$token/" "$DOTENV"
+  local var_name="CELESTIA_NODE_AUTH_TOKEN_${auth_level}"
+
+  ex "+%s/.*$var_name.*/$var_name=$token/" -scwq "$DOTENV"
```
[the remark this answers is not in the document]
zvolin: I mean, who uses that, sed is already scary for most xd

oblique (approved):
lgtm

zvolin (approved):

PR #72, opened 2023-09-18
Implement exchange server and header storage

fl0rek:


node/src/exchange/server.rs
```diff
@@ -1,52 +1,396 @@
...
-use tracing::instrument;
+use tendermint::hash::Algorithm;
+use tracing::{instrument, trace};
 
+use crate::exchange::utils::{ExtendedHeaderExt, HeaderRequestExt, HeaderResponseExt};
+use crate::exchange::{ReqRespBehaviour, ResponseType};
 use crate::store::Store;
 
-pub(super) struct ExchangeServerHandler<S>
+type StoreJobType<C> = dyn Future<Output = (C, ResponseType)> + Send;
```
zvolin, line 22: could `C` be `Channel`? I like single name generics only when I know what they mean :D

node/src/exchange/server.rs
```diff
@@ -1,52 +1,396 @@
...
+        ExchangeServerHandler {
+            store,
+            store_jobs: FuturesUnordered::new(),
+        }
     }
 
-    #[instrument(level = "trace", skip(self, _respond_to))]
-    pub(super) fn on_request_received(
+    #[instrument(level = "trace", skip(self, response_channel))]
+    pub(super) fn on_request_received<ID>(
```
zvolin, line 62: I'd use `Id` for non-const generics

node/src/exchange/server.rs
oblique: Tbh we dont need this, we can just use [`futures::future::BoxFuture`](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html) and it will be still readable.

Btw `C` was fine by me. :D

node/src/exchange/server.rs
```diff
@@ -1,52 +1,396 @@
...
+
+pub(super) struct ExchangeServerHandler<S, R = ReqRespBehaviour>
 where
-    S: Store + 'static,
+    S: Store,
+    R: ResponseSender,
 {
-    _store: Arc<S>,
+    store: Arc<S>,
+    store_jobs: FuturesUnordered<Pin<Box<StoreJobType<R::Channel>>>>,
```
oblique, line 30: As I said in another comment we can use [`BoxFuture`](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html):

```suggestion
    store_jobs: FuturesUnordered<BoxFuture<(R::Channel, ResponseType)>>,
```

node/src/exchange/server.rs
```diff
@@ -1,52 +1,396 @@
...
+        let Some((amount, data)) = self.parse_request(request) else {
+            self.store_jobs
+                .push(handle_invalid_request(response_channel).boxed());
+            return;
+        };
+
+        let store_job = match data {
+            header_request::Data::Origin(0) => {
+                handle_request_current_head(self.store.clone(), response_channel).boxed()
+            }
```
oblique, line 80: I prefer standalone functions only when they are an independent utility or when there is partial borrowing issue. 

You could define the handle method like this:

```rust
    fn handle_request_current_head(&mut self, channel: R::Channel) {
        let store = self.store.clone();

        let fut = async move {
            let response = store
                .get_head()
                .await
                .map(|head| head.to_header_response())
                .unwrap_or_else(|_| HeaderResponse::not_found());

            (channel, vec![response])
        };

        self.store_jobs.push(fut.boxed());
    }
```

and here use it like this:

```rust
            header_request::Data::Origin(0) => {
                self.handle_request_current_head(response_channel);
            }
```

which IMO will result to cleaner code

node/src/exchange/server.rs
```diff
@@ -1,52 +1,396 @@
...
+}
+
+async fn handle_request_current_head<S, C>(store: Arc<S>, channel: C) -> (C, ResponseType)
+where
+    S: Store,
+{
+    let response = match store.get_head().await {
+        Ok(head) => head.to_header_response(),
+        Err(_) => HeaderResponse::not_found(),
+    };
```
oblique, line 141: This is a good candidate for `unwrap_or_else`:

```rust
            let response = store
                .get_head()
                .await
                .map(|head| head.to_header_response())
                .unwrap_or_else(|_| HeaderResponse::not_found());
```

node/src/exchange/server.rs
```diff
@@ -1,52 +1,396 @@
...
+        let HeaderRequest {
+            amount,
+            data: Some(data),
+        } = request
+        else {
+            return None;
+        };
+
+        Some((amount, data))
+    }
```
oblique, line 121: This is a utility, so have it as standalone function

node/src/exchange/utils.rs
```diff
@@ -56,6 +67,20 @@ impl HeaderResponseExt for HeaderResponse {
             }
         }
     }
+
+    // TODO: how forthcoming should we be with errors and description?
+    fn not_found() -> HeaderResponse {
+        HeaderResponse {
+            body: NOT_FOUND_MSG.as_bytes().to_vec(),
```
oblique, line 74: Is Celestia encoding an error message in body?

oblique: [They do not](https://github.com/celestiaorg/go-header/blob/a6544e3c7354c4b798fe96ed47d8a537bac03ca3/p2p/server.go#L134-L136) encode error messages. So `body` should be empty `Vec`.

node/src/exchange.rs
```diff
@@ -231,6 +233,7 @@ where
                 return Poll::Ready(ev);
             }
         }
+        let _ = self.server_handler.poll(cx, &mut self.req_resp);
```
oblique, line 236: Based on the [`poll` convention](https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md#conventions-for-poll-implementations), the body of this method should be like this:

```rust
        loop {
            if let Poll::Ready(ev) = self.req_resp.poll(cx, params) {
                if let Some(ev) = self.on_to_swarm(ev) {
                    return Poll::Ready(ev);
                }

                continue;
            }

            if let Poll::Ready(()) = self.server_handler.poll(cx, &mut self.req_resp) {
                continue;
            }

            return Poll::Pending;
        }
```

node/src/exchange/server.rs
fl0rek: BoxFuture is indeed more readable, changed 👍

node/src/exchange/server.rs
fl0rek: Good point, changed

node/src/exchange/server.rs
fl0rek: switche to functional style 👍

node/src/exchange/server.rs
fl0rek: done 👍

node/src/exchange/utils.rs
fl0rek: changed 👍

node/src/exchange.rs
fl0rek: Thanks for the link, havent found that info yet  👍 changed according to guidelines.

node/src/exchange/server.rs
fl0rek: I've started with named utility functions to better convey what's being done to the store, but this indeed results in cleaner code and less generics, thanks!

node/src/exchange/server.rs
fl0rek: switched to BoxFuture and `futures` 👍

node/src/exchange/server.rs
```diff
@@ -1,52 +1,389 @@
...
+impl<S, R> ExchangeServerHandler<S, R>
 where
     S: Store + 'static,
+    R: ResponseSender,
 {
     pub(super) fn new(store: Arc<S>) -> Self {
-        ExchangeServerHandler { _store: store }
+        ExchangeServerHandler {
+            store,
+            futures: FuturesUnordered::new(),
```
oblique, line 54: I know `FuturesUnordered` suggestion was mine, but for some reason I didn't think at that time that we could just use `spawn` and an mpsc channel instead. There are some pros and cons for both cases, we can discuss them and decide.

fl0rek: switched to `spawn`ing and receiving the result over channel as discussed :+1:

node/src/exchange/server.rs
```diff
@@ -1,52 +1,382 @@
...
+    }
+
+    fn handle_invalid_request(&self, channel: R::Channel) {
+        let tx = self.tx.clone();
+
+        spawn(async move {
+            let _ = tx.send((channel, vec![HeaderResponse::invalid()])).await;
+        })
+    }
+}
```
oblique, line 171: Small optimization: We can `try_send` first and if it fails, then we do it thought a new task.

```rust
    fn handle_invalid_request(&self, channel: R::Channel) {
        if let Err(TrySendError::Full(val)) =
            self.tx.try_send((channel, vec![HeaderResponse::invalid()]))
        {
            let tx = self.tx.clone();

            spawn(async move {
                let _ = tx.send(val).await;
            });
        }
    }
```

node/src/exchange/server.rs
```diff
@@ -1,52 +1,382 @@
...
 
-    #[instrument(level = "trace", skip(self))]
     pub(super) fn on_failure(
         &mut self,
         peer: PeerId,
         request_id: RequestId,
         error: InboundFailure,
     ) {
-        // TODO
+        // TODO: cancel job if libp2p already failed it?
```
oblique, line 95: I think the failure here only happens if `send_response` fails. If that is the case, then job is already done.

node/src/exchange/server.rs
```diff
@@ -1,52 +1,382 @@
...
+                let response = store
+                    .get_by_height(i)
+                    .await
+                    .map(|head| head.to_header_response())
+                    .unwrap_or_else(|_| HeaderResponse::not_found());
+                responses.push(response);
+            }
+
+            let _ = tx.send((channel, responses)).await;
+        });
```
oblique, line 161: The Go implementation behaves differently: If `origin` does not exist then `NotFound` is send, otherwise if send only up to the height it managed to retrieve. Also I suggest to have a maximum amount, to avoid malicious requests.

Also as optimization we can get at the beginning if `origin` exists.

```rust
        if !store.has_at(origin) {
            self.send_not_found(response);
            return;
        }

        let store = self.store.clone();
        let tx = self.tx.clone();

        spawn(async move {
            let amount = amount.min(512);
            let mut responses = Vec::with_capacity(amount  as usize);

            for i in origin..origin + amount {
                match store.get_by_height(i).await {
                    Ok(header) => responses.push(header.to_header_response()),
                    Err(_) => break,
                }
            }

            if responses.is_empty() {
                responses.push(HeaderResponse::not_found());
            }

            let _ = tx.send((channel, responses)).await;
        });
```

node/src/exchange/server.rs
```diff
@@ -1,52 +1,382 @@
...
+        let (mut sender, receiver) = create_test_response_sender();
+
+        handler.on_request_received(
+            PeerId::random(),
+            "test",
+            HeaderRequest::with_origin(1, 1),
+            (),
+        );
+
+        poll_fn(move |cx| -> Poll<()> { handler.poll(cx, &mut sender) }).await;
```
oblique, line 235:
```suggestion
        poll_fn(move |cx| handler.poll(cx, &mut sender)).await;
```

oblique: fix it in other places too

node/src/exchange/server.rs
```diff
@@ -1,52 +1,382 @@
...
         error: InboundFailure,
     ) {
-        // TODO
+        // TODO: cancel job if libp2p already failed it?
+        trace!("on_failure; request_id: {request_id}, peer: {peer}, error: {error:?}");
+    }
+
+    pub fn poll(&mut self, cx: &mut Context<'_>, sender: &mut R) -> Poll<()>
+    where
+        <R as ResponseSender>::Channel: std::fmt::Debug,
```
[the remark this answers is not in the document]
fl0rek: yes, debugging leftover, removed :+1:

node/src/exchange/server.rs
```diff
@@ -1,52 +1,382 @@
...
+                    .await
+                    .map(|head| head.to_header_response())
+                    .unwrap_or_else(|_| HeaderResponse::not_found())
+            } else {
+                HeaderResponse::invalid()
+            };
+
+            let _ = tx.send((channel, vec![response])).await;
+        });
+    }
```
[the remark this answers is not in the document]
fl0rek: got rid of the tendermint dependency, thanks!

node/src/exchange/server.rs
fl0rek: nice, added the `try_send`

node/src/exchange/server.rs
fl0rek: Fixed the algorithm. As for the optimisation,`store.has_at` is also async, so I don't think we can skip `spawn`ing, I've added `has_at` check to at least avoid allocation at the expense of one extra check (we'd break the loop on first iteration anyway, so not much of a time save here).

node/src/exchange/server.rs
fl0rek: [`InboundFailure`](https://docs.rs/libp2p/latest/libp2p/request_response/enum.InboundFailure.html) can be sent even if we never call `send_response`. I've added artificial delay to spawned task, so that libp2p times out and I see we get 
```
...  WARN celestia_node::exchange::server: on_failure; request_id: 264, peer: 12D3KooWNJqKMTTH6Tmj5CNNksS1dkcMQzkqfAuoz98gdNQb895U, error: ConnectionClosed
```
But I'm still not sure if tracking necessary to manage jobs here is worth the potential resource save.

oblique: I think it doesn't worth it in our scenario because we don't have any heavy computation. 

Resources will became an issue when a malicious client is spamming us with requests and the channel gets full, in that case resources will be eaten by huge amount of spawned tasks. We need to somehow restrict on how many requests we can handle per time on a higher level.

node/src/exchange/server.rs
```diff
@@ -1,52 +1,393 @@
...
+
+    pub fn poll(&mut self, cx: &mut Context<'_>, sender: &mut R) -> Poll<()> {
+        if let Poll::Ready(Some((channel, response))) = self.rx.poll_recv(cx) {
+            sender.send_response(channel, response);
+            return Poll::Ready(());
+        }
+
+        Poll::Pending
+    }
```
oblique, line 107: Since this method doesn't return an actual result, the loop pattern needs to be used

node/src/exchange/server.rs
oblique: Maybe we can use this: https://docs.rs/libp2p/latest/libp2p/swarm/struct.SwarmBuilder.html#method.max_negotiating_inbound_streams

Also by dropping the `ResponseChannel` the stream will be closed. So, we may track the number of active requests and drop the rest.

node/src/exchange/server.rs
oblique: hmm. maybe in this case we can do something like:

```rust
            let mut responses = vec![];

            for i in origin..origin + amount {
                match store.get_by_height(i).await {
                    Ok(header) => {
                        if responses.is_empty() {
                            responses.reserve(amount as usize);
                        }
                        responses.push(header.to_header_response());
                    }
                    Err(_) => break,
                }
            }
```

fl0rek: Nice! Removed the unneeded double check.

node/src/exchange/server.rs
fl0rek: Fixed properly according to libp2p guidelines :+1:

node/src/exchange/server.rs
```diff
@@ -1,52 +1,408 @@
...
+        let expected_head = store.get_head().unwrap();
+        let mut handler = ExchangeServerHandler::new(Arc::new(store));
+        let (mut sender, receiver) = create_test_response_sender();
+
+        handler.on_request_received(PeerId::random(), "test", HeaderRequest::head_request(), ());
+
+        let received = select! {
+            _ = poll_fn(move |cx| handler.poll(cx, &mut sender)) => panic!("shouldn't return"),
+            r = receiver => { r.unwrap() }
+        };
```
oblique, line 230: Is this needed because of the loop?

fl0rek: Yes. Previously, returned `Ready` signaled `poll_fn` to finish, but now, whether is done is opaque from the caller side, so I'm relying on `receiver` to know when to stop.

oblique: Can you add a helper function that does this? Also add a comment in the function why it is needed.

node/src/exchange/server.rs
```diff
@@ -1,52 +1,408 @@
...
+    async fn request_range_beyond_head_test() {
+        let (store, _) = gen_filled_store(5);
+        let expected_hashes = [store.get_by_height(5).ok()];
+        let expected_status_codes = [StatusCode::Ok];
+        assert_eq!(expected_hashes.len(), expected_status_codes.len());
+
+        let mut handler = ExchangeServerHandler::new(Arc::new(store));
+        let (mut sender, receiver) = create_test_response_sender();
+
+        let request = HeaderRequest::with_origin(5, u64::try_from(expected_hashes.len()).unwrap());
```
zvolin, line 370: I don't think this requests beyond, it looks like a request for a single header

node/src/exchange/server.rs
fl0rek: created https://github.com/eigerco/celestia-node-rs/issues/90 for future tracking

node/src/exchange/server.rs
fl0rek: true, missed that request uses len from expected response. Fixed to request const `10` :+1:

node/src/exchange/server.rs
fl0rek: Added `poll_handler_for_result` with explanation at the bottom

oblique (approved):
lgtm

zvolin (approved):

PR #73, opened 2023-09-18
Implement header Store

### LINKED ISSUE #57: Implement the Store trait and in memory

fl0rek:
closes  #57

### DIFF
--- a/node/src/store.rs
+++ b/node/src/store.rs
@@ -3,9 +3,9 @@ use std::sync::atomic::{AtomicU64, Ordering};
 
 use async_trait::async_trait;
 use celestia_types::ExtendedHeader;
+use celestia_types::Hash;
 use dashmap::mapref::entry::Entry;
 use dashmap::DashMap;
-use celestia_types::Hash;
 use thiserror::Error;
 use tracing::{debug, instrument};
 
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -7,7 +7,6 @@ license = "Apache-2.0"
 [dependencies]
 celestia-proto = { workspace = true }
 celestia-types = { workspace = true }
-tendermint = { workspace = true }
 tendermint-proto = { workspace = true }
 
 async-trait = "0.1.73"


types/src/extended_header.rs
```diff
@@ -28,6 +28,13 @@ pub struct ExtendedHeader {
     pub dah: DataAvailabilityHeader,
 }
 
+use core::fmt::{Display, Formatter};
+impl Display for ExtendedHeader {
```
fl0rek, line 32: I found those useful, avoids printing entire Header in logs, not sure if we want this merged

node/src/exchange/utils.rs
```diff
@@ -58,11 +64,63 @@ impl HeaderResponseExt for HeaderResponse {
     }
 }
 
-pub(super) trait ExtendedHeaderExt {
-    fn to_header_response(&self) -> HeaderResponse;
+pub trait ExtendedHeaderExt {
+    fn with_height(height: u64) -> ExtendedHeader;
```
fl0rek, line 68: I've canibalised this trait and moved conversion to header response to `ToHeaderResponse` as I want to implement that for easier Result transformation.
This Header generation is used currently only in tests, not sure what's a good place for it.

oblique: I believe this shouldn't be here because it generates an invalid `ExtendedHeader` which its current purpose is ti be used in tests. I'm in favor of something like:

```rust
fn gen_extended_header(height: u64) -> ExtendedHeader
```

Also I prefer `to_header_response` in `ExtendedHeaderExt`. I can not see why it had to be changed.

node/src/store.rs
```diff
@@ -1,9 +1,136 @@
...
+    }
+
+    pub fn get_head_height(&self) -> usize {
+        self.height_to_hash.len()
+    }
+
+    pub fn append_header(&self, header: ExtendedHeader) -> Result<(), InsertError> {
+        let hash = header.hash();
+        let height = header.height(); // TODO: should Store be responsible for making sure we have
+                                      // all 1..=head headers?
```
oblique, line 54: It shouldn't. Ideally `Syncer` should be able to parallelize the filling of the `Store`, so it will never be able to have `1..=head` headers. What store should be responsible is returning all the continues ranges we have in the store and the missing ranges.

node/src/store.rs
```diff
@@ -1,9 +1,136 @@
...
+    #[error("Store in inconsistent state, lost head")]
+    LostStoreHead,
+    #[error("Store in inconsistent state, height->hash mapping exists, but hash doesn't")]
+    LostHash,
+    #[error("Failed to convert height from usize to u64, should not happen")]
+    HeadHeightConversionError,
+
+    #[error("Header not found in store")]
+    NotFound,
+}
```
oblique, line 37: Merge these in `StoreError`

node/src/store.rs
```diff
@@ -1,9 +1,136 @@
...
+        let head_height = match self.height_to_hash.len() {
+            0 => return Err(ReadError::NotFound),
+            height => match u64::try_from(height) {
+                Ok(h) => h,
+                Err(e) => {
+                    error!("Failed to convert {height} from usize to u64, should not happen: {e}");
+                    return Err(ReadError::HeadHeightConversionError);
+                }
+            },
+        };
```
oblique, line 86: This can be unnested like this:

```rust
        let head_height = match u64::try_from(self.height_to_hash.len()) {
            Ok(0) => return Err(ReadError::NotFound),
            Ok(height) => height,
            Err(e) => {
                error!("Failed to convert {height} from usize to u64, should not happen: {e}");
                return Err(ReadError::HeadHeightConversionError);
            }
        };
```

node/src/store.rs
```diff
@@ -1,9 +1,136 @@
...
+
+        info!("Will insert {hash} at {height}");
+        hash_entry.insert(header);
+        height_entry.insert(hash);
+
+        Ok(())
+    }
+
+    pub fn get_head(&self) -> Result<ExtendedHeader, ReadError> {
+        let head_height = match self.height_to_hash.len() {
```
oblique, line 77: Based on my comment about `1..=head`, this can not be done. The highest know height number must be save in a variable and used directly. If you are going the use the continues range idea, it will be the `end` of the latest range.

node/src/store.rs
```diff
@@ -1,9 +1,136 @@
...
+
+        let Some(head_hash) = self.height_to_hash.get(&head_height).map(|v| *v) else {
+            error!("height_to_hash[height_to_hash.len()] not found, store is inconsistent");
+            return Err(ReadError::LostStoreHead);
+        };
+
+        match self.headers.get(&head_hash) {
+            Some(v) => Ok(v.clone()),
+            None => {
+                error!("Header with hash {head_hash} for height {head_height} missing");
```
oblique, line 96: I'm not sure if I like the error logging and then returning of the error. It produces a noise to read code. I suggest to implement the error messages in the error type itself and the caller should log it. Also in this case, instead of giving to the caller the responsibility, you can use [`#[instrument(err)]`](https://docs.rs/tracing/latest/tracing/attr.instrument.html) from tracing crate.

node/src/store.rs
```diff
@@ -1,9 +1,136 @@
...
+                Err(ReadError::LostHash)
+            }
+        }
+    }
+
+    pub fn get_by_hash(&self, hash: &Hash) -> Result<ExtendedHeader, ReadError> {
+        match self.headers.get(hash) {
+            Some(v) => Ok(v.clone()),
+            None => Err(ReadError::NotFound),
+        }
```
oblique, line 106:
```suggestion
        self.headers.get(hash).cloned().ok_or(ReadError::NotFound)
```

node/src/store.rs
```diff
@@ -1,9 +1,136 @@
...
+            Some(h) => Ok(h.clone()),
+            None => {
+                error!("Lost hash {hash} at height {height}");
+                Err(ReadError::LostHash)
+            }
+        }
+    }
+
+    #[doc(hidden)]
+    pub fn test_filled_store(height: u64) -> Self {
```
oblique, line 124: This should be in `mod tests` instead.

types/src/extended_header.rs
```diff
@@ -28,6 +28,13 @@ pub struct ExtendedHeader {
     pub dah: DataAvailabilityHeader,
 }
 
+use core::fmt::{Display, Formatter};
```
oblique, line 31: Move this at the beginning of the file

types/src/extended_header.rs
oblique: I think it's ok

oblique (changes requested):

node/src/exchange.rs
```diff
@@ -21,11 +21,13 @@ use tracing::instrument;
 mod client;
 mod server;
 mod utils;
+pub use utils::ExtendedHeaderExt;
```
oblique, line 24: why this?

oblique:
I think to correctly close #57 we need to implement a `Storage` trait, to make store/load operations generic. Also since this is planned to be used with non-memory databases, the operations should be async.

@zvolin Should we create another issue in the milestone for the trait or implement it now?

node/src/p2p.rs
```diff
@@ -327,6 +330,10 @@ impl Worker {
             SwarmEvent::ConnectionClosed { peer_id, .. } => {
                 self.peer_tracker.remove(peer_id);
             }
+            #[cfg(debug_assertions)]
+            SwarmEvent::NewListenAddr { address, .. } => {
+                info!("listening on: {address}");
+            }
```
oblique, line 336: why the `#[cfg(debug_assertions)]`?

node/src/store.rs
fl0rek: Neat, haven't thought of that here. 
Entire thing got removed though as we store head height separately.

node/src/store.rs
fl0rek: switched to functional style here and elsewhere

types/src/extended_header.rs
fl0rek: moved :+1:

node/src/exchange/utils.rs
fl0rek: My idea was to have `ToHeaderResponse` solely for converting stuff to HeaderResponse, as I want to use that to [convert Result from store into respose ](https://github.com/eigerco/celestia-node-rs/pull/72/files#diff-93f633bcbe6080e868774f979d9cae6ffa477bcc909c263b66d42372172176d6R138-R153). 
I see your point this not generating valid headers, so moved this to test as a freestanding function.

node/src/exchange.rs
fl0rek: since utils was private I had to pry that out somehow :P
`gen_extended_header` and `gen_filled_store` live in `node::store::tests` now, but it isn't ideal as it's used in other tests too.. Should we spin out `test-utils` or similar for non-localised test utils?

node/src/store.rs
```diff
@@ -1,9 +1,214 @@
+use celestia_types::ExtendedHeader;
+use dashmap::mapref::entry::Entry;
+use dashmap::DashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
+use tendermint::Hash;
+use thiserror::Error;
+use tracing::{error, info, instrument};
+
+trait StoreTrait {
```
zvolin, line 9: please rename it to `Store` and then add a meaningful name for implementation, like eg. `InMemoryStore`. We will have a few stores and likely will refer to them by the trait name in most places

zvolin, line 9: I'm also wondering if that shouldn't be an async trait. web storage api looks synchronous, but we don't know what db are we gonna use there

node/src/exchange/utils.rs
oblique: I don't like the convertion of Result into response. I would prefer an explicit function if it going to be used in many places. If it's just one place then I prefer `match`.  If your intention was to have a cleaner logic before sending the result, one idea is to add this constructor in `HeaderResponseExt`:

```rust
fn not_found() -> HeaderResponse;
```

Also, we should never send Invalid. That was added intetionally because of Protobuf convertion rules.

node/src/node.rs
```diff
@@ -54,7 +53,7 @@ where
     SyncerSrv: SyncerService<P2pSrv>,
 {
     pub async fn new(config: NodeConfig) -> Result<Self, NodeError<P2pSrv, SyncerSrv>> {
-        let store = Arc::new(RwLock::new(Store::new()));
```
zvolin, line 57: I think store should be used by the trait, not directly

node/src/exchange/utils.rs
oblique: I think Invalid actually can be send back if the Request is invalid. We can try to send an empty HeaderRequest to the network and see if it behaves that way.

node/src/exchange.rs
oblique: Yes we can have a `test_utils.rs`. I was also thinking about it.

node/src/exchange.rs
```diff
@@ -21,11 +21,11 @@ use tracing::instrument;
 mod client;
 mod server;
 mod utils;
-
```
oblique, line 24: restore this empty line

types/src/extended_header.rs
```diff
@@ -1,3 +1,4 @@
...
```
oblique, line 1: use `std` for consistency.

types/src/extended_header.rs
```diff
@@ -28,6 +29,12 @@ pub struct ExtendedHeader {
     pub dah: DataAvailabilityHeader,
 }
 
+impl Display for ExtendedHeader {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
```
oblique, line 33: use `std` for consistency

node/src/store.rs
```diff
@@ -1,14 +1,361 @@
...
+use dashmap::DashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use tendermint::Hash;
+use thiserror::Error;
+use tracing::{error, info, instrument};
+
+type Result<T, E = StoreError> = std::result::Result<T, E>;
+
+pub type WrappedStore = Arc<Box<dyn Store + Sync + Send>>;
```
oblique, line 14: Usually this pattern has `Boxed` as suffix and without the `Arc`.

```suggestion
pub type StoreBoxed = Box<dyn Store + Sync + Send>;
```

node/src/store.rs
```diff
@@ -1,14 +1,361 @@
...
+
+#[async_trait]
+pub trait Store: Send + Sync + Debug {
```
oblique, line 17: If you have `Send + Sync` in the boxed type, then it is not needed to be bound also here. Keep one of the two.

I'm not sure what I prefer.

oblique: Usually the `Send + Sync` is not bound to the trait itself, unless if it is a hard requirement.

node/src/store.rs
fl0rek: Per our yesterday's discussion it seems we want Store to keep continuous range from genesis to current verified head and keep syncer responsible for catching up and feeding the store headers in ordered fasion. I'll change the store code and add tests.

node/src/store.rs
fl0rek: Good point, switched to `instrument` suggestion and removed explicit error logs.

node/src/lib.rs
```diff
@@ -7,6 +7,8 @@ pub mod p2p;
 pub mod peer_tracker;
 pub mod store;
 pub mod syncer;
+#[cfg(test)]
+pub mod test_utils;
```
oblique, line 11:
```suggestion
mod test_utils;
```

node/src/store.rs
fl0rek: Created `test_utils` and moved this and `gen_extended_header` there.

node/src/exchange.rs
fl0rek: Created `test_utils` and moved this and `gen_extended_header` there.

node/src/store.rs
fl0rek: renamed :+1:

node/src/node.rs
```diff
@@ -54,7 +53,7 @@ where
...
+        let store: Arc<BoxedStore> = Arc::new(Box::new(InMemoryStore::new()));
```
oblique, line 56: Store should be configurable, so add the following in the `NodeConfig`, without the `Arc`. Wrap it in `Arc` in this method.

Also since this type went all the way up to the `Node` constructor, I started thinking that maybe it shouldn't be boxed and be generic. Box is nice if you want to hide the abstraction from the upper layer, but we are going to expose it because we want it configurable.

node/src/store.rs
```diff
@@ -1,14 +1,296 @@
+use async_trait::async_trait;
+use celestia_types::ExtendedHeader;
+use core::fmt::Debug;
```
oblique, line 3:
```suggestion
use std::fmt::Debug;
```

node/src/store.rs
```diff
@@ -1,14 +1,296 @@
...
+use dashmap::mapref::entry::Entry;
+use dashmap::DashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
```
oblique, line 6: split the imports to three groups, like we do in other files

node/src/store.rs
fl0rek: renamed :+1:

node/src/store.rs
```diff
@@ -1,14 +1,296 @@
...
+            if let Err(e) = self.append_single(header).await {
+                error!("error appending: {e}");
+                return Err(StoreError::ContinuousAppendFailedAt(idx));
+            }
+        }
+
+        Ok(())
+    }
+}
+*/
```
oblique, line 53: Any reason you want an extended trait?

node/src/store.rs
```diff
@@ -1,14 +1,296 @@
...
+        info!("Will insert {hash} at {height}");
+        hash_entry.insert(header);
+        height_entry.insert(hash);
+
+        self.head_height.store(height.value(), Ordering::Release);
+
+        Ok(())
+    }
+
+    //pub fn add(&self, header: ExtendedHeader) -> Result<(), StoreError> { }
```
oblique, line 165: remove commented code

node/src/store.rs
```diff
@@ -1,14 +1,296 @@
...
+        // https://github.com/xacrimon/dashmap/issues/233
+        let hash_entry = self.headers.entry(hash);
+        let height_entry = self.height_to_hash.entry(height.into());
+
+        if matches!(hash_entry, Entry::Occupied(_)) {
+            return Err(StoreError::HashExists(hash));
+        }
+        if matches!(height_entry, Entry::Occupied(_)) {
+            return Err(StoreError::HeightExists(height.into()));
+        }
```
oblique, line 151: We need to check about this. Some amount of the most current blocks can have the same height but different hashes.

Should the Syncer decide which chain wins?

node/src/node.rs
fl0rek: `Store` implementation is now passed via `NodeConfig`

node/src/store.rs
fl0rek: fixed :+1:

node/src/p2p.rs
fl0rek: to lessen the impact :P I added that for testing connection between nodes manually, I'll remove that now

node/src/exchange/utils.rs
fl0rek: Yep! I'm sending `Invalid`, if there's something wrong with the request and `NotFound`, if, well, not found or if there's an issue retrieving the value from store (I think it makes more sense than returning Invalid, if request is correct but there might be issues on our side).

This seems consistent with go node, I see `Invalid` back for empty or nonsensical requests.

node/src/store.rs
```diff
@@ -1,14 +1,281 @@
...
+}
+
+#[derive(Error, Debug, PartialEq)]
+pub enum StoreError {
+    // TODO: error consolidation again
+    #[error("Hash {0} already exists in store")]
+    HashExists(Hash),
+    #[error("Height {0} already exists in store")]
+    HeightExists(u64),
+
```
oblique, line 83: Leave 1 empty line between all the errors

node/src/store.rs
```diff
@@ -1,14 +1,281 @@
...
+    }
+
+    async fn has_at(&self, height: u64) -> bool {
+        self.exists_by_height(height)
+    }
+
+    async fn append_single(&mut self, header: ExtendedHeader) -> Result<()> {
+        self.append_continuous(header)
+    }
+}
```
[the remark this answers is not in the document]
fl0rek: moved 👍

node/src/store.rs
```diff
@@ -1,14 +1,281 @@
...
+#[async_trait]
+pub trait Store: Send + Sync + Debug {
+    async fn get_by_hash(&self, hash: &Hash) -> Result<ExtendedHeader>;
+    async fn get_by_height(&self, height: u64) -> Result<ExtendedHeader>;
+
+    async fn height(&self) -> u64;
+    async fn has(&self, hash: &Hash) -> bool;
+    async fn has_at(&self, height: u64) -> bool;
+
+    async fn append_single(&mut self, header: ExtendedHeader) -> Result<()>;
```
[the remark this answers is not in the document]
fl0rek: yes, we also don't need it, removing.
 I seems to have cleaned that up only on `feat/exchange-server1` branch, not here.

node/src/store.rs
fl0rek: merged 👍

node/src/store.rs
```diff
@@ -1,14 +1,283 @@
...
+    async fn get_by_hash(&self, hash: &Hash) -> Result<ExtendedHeader>;
+    async fn get_by_height(&self, height: u64) -> Result<ExtendedHeader>;
+
+    async fn height(&self) -> u64;
+    async fn has(&self, hash: &Hash) -> bool;
+    async fn has_at(&self, height: u64) -> bool;
+
+    async fn append_single(&self, header: ExtendedHeader) -> Result<()>;
+
+    async fn append<I: IntoIterator<Item = ExtendedHeader>>(&self, headers: I) -> Result<()>
```
oblique, line 25: `I` is already defined in the where-clause
```suggestion
    async fn append<I>(&self, headers: I) -> Result<()>
```

node/src/store.rs
```diff
@@ -1,14 +1,283 @@
...
+    where
+        I: IntoIterator<Item = ExtendedHeader> + Send,
+        <I as IntoIterator>::IntoIter: Send,
+    {
+        let headers = headers.into_iter();
+
+        for (idx, header) in headers.enumerate() {
+            if self.append_single(header).await.is_err() {
+                return Err(StoreError::ContinuousAppendFailedAt(idx));
```
oblique, line 34: An index will not make sense in the error message. Use `height`

oblique: Or a better idea: In `append_continus` return this:

```
    return NonContinuousAppend(self.get_head_height(), height.value());
```

And here just propagate the error.

fl0rek: I think the reference Store interface had it return number of headers successfuly applied, but I like your suggestion better, it simplifies `StoreError` and implementation.

node/src/store.rs
fl0rek: removed, didn't notice that, thanks. 👍  
Weird, I'd expect clippy to catch me adding obviously redundant bounds

node/src/store.rs
oblique: After talking with the Celestia team, they told me that it is impossible to have a block with same height but different hashes. So we can say that this if-statement is valid.

node/src/store.rs
```diff
@@ -1,14 +1,326 @@
...
+        let Some(hash) = self.height_to_hash.get(&height).as_deref().copied() else {
+            return false;
+        };
+
+        self.headers.get(&hash).is_some()
+    }
+
+    #[instrument(err)]
+    pub fn get_by_height(&self, height: u64) -> Result<ExtendedHeader, StoreError> {
+        let Some(hash) = self.height_to_hash.get(&height).as_deref().copied() else {
```
oblique, line 165: Small optimization, after you apply the suggestion on `exists_by_height` :

```suggestion
        if !exists_by_height(height) {
                return Err(StoreError::NotFound);        
        }

        let Some(hash) = self.height_to_hash.get(&height).as_deref().copied() else {
```

node/src/store.rs
```diff
@@ -1,14 +1,326 @@
...
+        };
+
+        self.headers
+            .get(&head_hash)
+            .as_deref()
+            .cloned()
+            .ok_or(StoreError::LostHash(head_hash))
+    }
+
+    pub fn exists_by_hash(&self, hash: &Hash) -> bool {
```
oblique, line 142: As a name I prefer `has_hash` or `contains_hash`. In std library they use `contains` in these cases.

node/src/store.rs
```diff
@@ -1,14 +1,326 @@
...
+    #[instrument(err)]
+    pub fn get_by_hash(&self, hash: &Hash) -> Result<ExtendedHeader, StoreError> {
+        self.headers
+            .get(hash)
+            .as_deref()
+            .cloned()
+            .ok_or(StoreError::NotFound)
+    }
+
+    pub fn exists_by_height(&self, height: u64) -> bool {
```
oblique, line 155: Same as my other comment: `has_height` or `contains_height`

node/src/store.rs
```diff
@@ -1,14 +1,326 @@
...
+    pub fn with_genesis(genesis: ExtendedHeader) -> Self {
+        let genesis_hash = genesis.hash();
+        let genesis_height = genesis.height().value();
+
+        InMemoryStore {
+            headers: DashMap::from_iter([(genesis_hash, genesis)]),
+            height_to_hash: DashMap::from_iter([(genesis_height, genesis_hash)]),
+            head_height: AtomicU64::new(genesis_height),
+        }
+    }
```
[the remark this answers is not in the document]
fl0rek: I'm not sure if we can do that. Notion page mentions maintaining continuous range from genesis to head, not necessarily from height zero. Go implementation even has even tests for initialising store with genesis header of arbitrary height.
https://github.com/celestiaorg/go-header/blob/main/store/store_test.go#L153

node/src/store.rs
```diff
@@ -1,14 +1,326 @@
...
+            return Err(StoreError::HeightExists(height.into()));
+        }
+        if self.get_head_height() + 1 != height.value() {
+            return Err(StoreError::NonContinuousAppend(
+                self.get_head_height(),
+                height.value(),
+            ));
+        }
+
+        info!("Will insert {hash} at {height}");
```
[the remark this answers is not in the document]
fl0rek: Thought we'd want to be notified about this, or do you think it would be too chatty?

node/src/store.rs
```diff
@@ -1,14 +1,326 @@
...
+            .cloned()
+            .ok_or(StoreError::NotFound)
+    }
+
+    pub fn exists_by_height(&self, height: u64) -> bool {
+        let Some(hash) = self.height_to_hash.get(&height).as_deref().copied() else {
+            return false;
+        };
+
+        self.headers.get(&hash).is_some()
```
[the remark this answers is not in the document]
fl0rek: [My comment](https://github.com/eigerco/celestia-node-rs/pull/73#discussion_r1332621621) about genesis head also applies here, but I think we can just keep genesis height and apply your optimisation

node/src/store.rs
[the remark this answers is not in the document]
oblique: It is too chatty. You can have it as `debug!` or `trace!`, but also reworded to something like: `Inserting header {hash} with height {height}`

node/src/store.rs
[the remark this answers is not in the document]
oblique: So practically we can say if HEAD is 30000, we can set genesis to 29000?

node/src/store.rs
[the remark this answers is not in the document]
fl0rek: changed log level and message 👍

node/src/store.rs
```diff
@@ -1,14 +1,326 @@
...
+            head_height: AtomicU64::new(genesis_height),
+        }
+    }
+
+    pub fn get_head_height(&self) -> u64 {
+        self.head_height.load(Ordering::Acquire)
+    }
+
+    #[instrument(err)]
+    pub fn append_continuous(&self, header: ExtendedHeader) -> Result<(), StoreError> {
```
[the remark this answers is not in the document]
fl0rek: Added `_unverified` suffix as well as doc comment with function expectations about responsibilities + TODO so that we can put verification/validation in the right place once syncer crystalises.

node/src/store.rs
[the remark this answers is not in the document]
fl0rek: Yes, that's how I understand it.

node/src/store.rs
fl0rek: went with `contains_height`

node/src/store.rs
```diff
@@ -1,14 +1,354 @@
...
+        self.headers
+            .get(hash)
+            .as_deref()
+            .cloned()
+            .ok_or(StoreError::NotFound)
+    }
+
+    pub fn contains_height(&self, height: u64) -> bool {
+        height != 0
+            && height >= self.genesis_height.load(Ordering::Acquire)
```
oblique, line 166:
```suggestion
            && height >= self.get_genesis_height()
```

node/src/store.rs
```diff
@@ -1,14 +1,354 @@
+use std::fmt::Debug;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use async_trait::async_trait;
+use celestia_types::ExtendedHeader;
+use dashmap::mapref::entry::Entry;
+use dashmap::DashMap;
+use tendermint::Hash;
```
zvolin, line 8:
```suggestion
use celestia_types::Hash;
```

oblique: In this case remove also tendermint from `[dependencies]`

zvolin (approved):
Nice

node/src/store.rs
fl0rek: nice, removed 👍

oblique (approved):
lgtm

PR #74, opened 2023-09-21
chore: switch to upstream nmt-rs

zvolin:


fl0rek (approved):
:guitar:

types/src/nmt/namespace_proof.rs
```diff
@@ -68,8 +68,7 @@ impl TryFrom<RawProof> for NamespaceProof {
         let mut proof = NmtNamespaceProof::PresenceProof {
             proof: NmtProof {
                 siblings,
-                start: value.start as u32,
-                end: value.end as u32,
+                range: value.start as u32..value.end as u32,
```
oblique, line 71: was the previous range `[start, end)`?

oblique (commented):
lgtm

types/src/nmt/namespace_proof.rs
oblique: actually range is strongly typed so we are ok

PR #75, opened 2023-09-21
feat(proto): add cosmos Tx and celestia MsgPayForBlobs

zvolin:


proto/build.rs
```diff
@@ -86,12 +87,15 @@ fn main() -> Result<()> {
...
                 "vendor/celestia/da/data_availability_header.proto",
+                "vendor/celestia/blob/v1/tx.proto",
                 "vendor/header/pb/extended_header.proto",
                 "vendor/share/p2p/shrexnd/pb/share.proto",
                 "vendor/share/eds/byzantine/pb/share.proto",
                 "vendor/cosmos/base/v1beta1/coin.proto",
                 "vendor/cosmos/base/abci/v1beta1/abci.proto",
+                "vendor/cosmos/crypto/multisig/v1beta1/multisig.proto",
                 "vendor/cosmos/staking/v1beta1/query.proto",
+                "vendor/cosmos/tx/v1beta1/tx.proto",
```
fl0rek, line 98: to confirm, these are the new definitions we need and the rest are protobuf imported dependencies?

fl0rek (approved):
protobuf :rocket:

oblique (approved):

proto/build.rs
oblique: Previously we vendored partially Cosmos protobuf. For this PR we need `MsgPayForBlobs` and it's dependencies.

zvolin: yeah, from cosmos we need the `Tx` and from celestia we need `MsgPayForBlobs`

PR #78, opened 2023-09-21
feat(types): expose commitment creation for shares

zvolin:


oblique (approved):

types/src/blob/commitment.rs
```diff
@@ -25,40 +25,43 @@ impl Commitment {
...
         namespace: Namespace,
         share_version: u8,
         blob_data: &[u8],
     ) -> Result<Commitment> {
         let shares = split_blob_to_shares(namespace, share_version, blob_data)?;
+        Self::for_shares(namespace, &shares)
+    }
 
+    /// Generate the commitment for given shares.
+    pub fn for_shares(namespace: Namespace, mut shares: &[impl AsRef<[u8]>]) -> Result<Commitment> {
```
zvolin, line 38: question, should we require namespace there and check if all shares have this namespace inside?

zvolin: Another one, should I check if only the first one has sequence len and all are together taking a sequence len bytes +  optional padding?

zvolin: Would it be wrong to just have arbitrary bytes there and not verify them as shares?

zvolin: I added a comment that we treat the shares there as arbitrary byte blobs and perform no checks. Please resolve this issue if you agree

oblique (approved):

PR #79, opened 2023-09-22
feat: Peer discovery with Kademlia

oblique:


zvolin (commented):
This looks super cool! I'm wondering, is it possible to write a test where you spin up 2 nodes that don't know of each other but both connect to the same celestia node and test if they eventually discover themselves?

oblique:
We can start two `celestia` instances, but both need to have a listening address that can be reached from the outside. I can test this manually. But I'm not sure if it is possible to do this in the CI.

zvolin (approved):
coooool

PR #80, opened 2023-09-25
fix(node/exchange): Forward handling of pending connections to req_resp

oblique:


zvolin (commented):
I'm not sure I understand why do we need that. It seems those 2 are provided by the trait when we implement the rest, is something wrong with the provided implementations?

oblique:
Yes, the provided ones do nothing ([inbound](https://github.com/libp2p/rust-libp2p/blob/636966697ea89c333f15148220d88a5f46e91f09/swarm/src/behaviour.rs#L133-L140), [outbound](https://github.com/libp2p/rust-libp2p/blob/636966697ea89c333f15148220d88a5f46e91f09/swarm/src/behaviour.rs#L169-L177)). However when you wrap a `NetworkBehaviour` the correct and future-proof way is to forward then to the underlying type because it might implement them. I found about this when I was navigating the source code of `Kademlia`, which implement those methods ([inbound](https://github.com/libp2p/rust-libp2p/blob/636966697ea89c333f15148220d88a5f46e91f09/protocols/kad/src/behaviour.rs#L2065-L2086), [outbound](https://github.com/libp2p/rust-libp2p/blob/636966697ea89c333f15148220d88a5f46e91f09/protocols/kad/src/behaviour.rs#L2111-L2143))

zvolin (approved):

zvolin:
thanks for explanation!

PR #81, opened 2023-09-25
chore: Split gossipsub code in smaller functions

oblique:


node/src/p2p.rs
```diff
@@ -420,4 +404,40 @@ where
...
+
+    #[instrument(skip_all)]
+    fn on_header_sub_message(&mut self, data: &[u8]) {
+        let Ok(header) = ExtendedHeader::decode(data) else {
+            trace!("Malformed header from header-sub");
+            return;
+        };
+
+        if let Err(e) = header.validate() {
+            trace!("Invalid header from header-sub ({e})");
```
zvolin, line 416: could we also have TODO there to not propagate it further?

zvolin (approved):

zvolin (approved):

PR #82, opened 2023-09-25
feat(node): Add state of peer in PeerTracker

oblique:


node/src/peer_tracker.rs
```diff
@@ -1,56 +1,163 @@
...
+use libp2p::{identify, Multiaddr, PeerId};
 
+#[derive(Debug)]
 pub struct PeerTracker {
-    peers: DashSet<PeerId>,
+    peers: DashMap<PeerId, InnerState>,
+}
+
+#[derive(Debug)]
+struct InnerState {
```
zvolin, line 14: maybe some name like `PeerInfo` would do?

oblique: yes sounds much better

node/src/peer_tracker.rs
```diff
@@ -1,56 +1,163 @@
...
 
-    pub fn remove(&self, peer: PeerId) {
-        self.peers.remove(&peer);
+    pub fn clear_addresses(&self, peer: PeerId) {
+        let mut state = self.get(peer);
+
+        state.addrs.clear();
+        state.addrs.shrink_to_fit();
+
+        // Downgrade state. Note that we do not downgrade a connected peer.
```
zvolin, line 90: Is it even okay to remove the addresses of connected peer, or maybe the one that we are connected through?

oblique: Not sure. I don't use this method yet

oblique: I also had an idea that the connected address can be stored in a different field as a primary address. But in general I'm not sure if we really need to keep track of the addresses

zvolin (approved):
looks good

PR #85, opened 2023-09-28
feat: Improve verification and implement verification in Exchange client

oblique:


node/src/exchange/client.rs
```diff
@@ -232,25 +232,31 @@ where
...
-                    }
+                let first = headers.first().expect("empty headers vector");
+
+                if first.height().value() != *start {
+                    return Err(ExchangeError::InvalidResponse);
                 }
+
+                first
+                    .verify_adjacent_range(&headers[1..])
+                    .map_err(|_| ExchangeError::InvalidResponse)?;
```
oblique, line 259: I'm still not sure about this

oblique: After a discussion with Celestia team, the verification must only be done in `get_verified_headers_range` and Syncer will use that method

types/src/lib.rs
```diff
@@ -14,11 +15,13 @@ pub(crate) mod serializers;
...
 pub mod state;
 mod sync;
+#[cfg(feature = "test-utils")]
+pub mod test_utils;
 pub mod trust_level;
 mod validate;
 mod validator_set;
 
-pub use tendermint::hash::Hash;
+pub type Height = tendermint::block::Height;
```
zvolin, line 24: please move it to our `block` mod

types/src/block.rs
```diff
@@ -30,10 +30,19 @@ impl ValidateBasic for Header {
             bail_validation!("height == 0")
         }
 
-        if self.last_block_id.is_none() && self.height.value() != GENESIS_HEIGHT {
+        if self.height.value() == GENESIS_HEIGHT && self.last_block_id.is_some() {
```
zvolin, line 33: please add a testcase

types/src/data_availability_header.rs
```diff
@@ -29,6 +29,17 @@ impl DataAvailabilityHeader {
     pub fn column_root(&self, column: usize) -> Option<NamespacedHash> {
         self.column_roots.get(column).cloned()
     }
+
+    pub fn hash(&self) -> Hash {
+        let all_roots: Vec<_> = self
+            .row_roots
+            .iter()
+            .chain(self.column_roots.iter())
+            .map(|root| root.to_vec())
```
zvolin, line 38:
```suggestion
            .map(|root| root.to_array())
```

types/src/data_availability_header.rs
```diff
@@ -18,7 +19,6 @@ use crate::{bail_validation, Error, Result, ValidateBasic, ValidationError};
 pub struct DataAvailabilityHeader {
     pub row_roots: Vec<NamespacedHash>,
     pub column_roots: Vec<NamespacedHash>,
-    pub hash: Hash,
```
zvolin, line 21: why do we remove `hash` there?

zvolin: I mean, it has to be computed at least once during validate (which we perform on each decoding) so it will just result in additional computations when someone needs it

types/src/test_utils.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::nmt::{NamespacedHash, NamespacedHashExt};
+use crate::{DataAvailabilityHeader, ExtendedHeader, ValidatorSet};
+
+pub type SigningKey = ed25519_consensus::SigningKey;
+
+pub fn generate_ed25519_key() -> SigningKey {
+    SigningKey::new(rand::thread_rng())
+}
+
+impl ExtendedHeader {
```
zvolin, line 23: I'm wondering if it wouldn't be more convenient to use if we'd introduce a newtype like `TestExtendedHeader` that would combine the EH and key in itself, so that the key doesn't need to be passed everywhere. It could also `impl Deref<Target = ExtendedHeader>` so it wouldn't be less convenient to use

types/src/data_availability_header.rs
oblique: There are multiple reasons:

* Since `row_roots` and `column_roots` are public, if they are modified, the `hash` gets invalidated. This was confusing when I was implementing the test utils of `ExtendedHeader`.
* I think I prefer to keep it consistent with `tendermint` types.
* After validation it can be retrieved from `extended_header.header.data_hash`, we can add `ExtendedHeader::dah_hash` method that will return the validated one. We did the same for `Header::hash()`.

types/src/test_utils.rs
oblique: I think If I will add a new type, I prefer `ExtendedHeaderGenerator`. and have the methods there with additional utilities too. Generator will also initialize the SigningKey internally and user doesn't need to know about it.

types/src/block.rs
oblique: done

types/src/lib.rs
oblique: done

types/src/test_utils.rs
oblique: implemented `ExtendedHeaderGenerator`

types/src/test_utils.rs
```diff
@@ -0,0 +1,449 @@
...
+
+    /// Creates new `ExtendedHeaderGenerator` and skip an amount of headers.
+    ///
+    /// This is equivalent to:
+    ///
+    /// ```ignore
+    /// let mut gen = ExtendedHeaderGenerator::new();
+    /// gen.skip(amount);
+    /// ```
+    pub fn new_skipped(amount: u64) -> ExtendedHeaderGenerator {
```
zvolin, line 50: I'd rather prefer something like `new_from_height(height: u64)` that generates the first header of `height` height. I found this `new_skipped` a bit weird to understand

oblique: > generates the first header of `height` height.

I don't want to return a `(ExtendedHeaderGenerator, ExtendedHeader)`, `new_from_height` as a name is ok with me.

types/src/test_utils.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// let mut gen = ExtendedHeaderGenerator::new();
+    /// let header1 = gen.next();
+    /// let header2 = gen.next();
+    /// let another_header2 = gen.next_of(&header1);
+    /// ```
+    ///
+    /// # Note
+    ///
+    /// This method does not change the state of `ExtendedHeaderGenerator`.
+    pub fn next_of(&self, header: &ExtendedHeader) -> ExtendedHeader {
```
zvolin, line 105: as this doesn't advance the generator I'd name it `another_of`

oblique: `another_of` would be nice if we provided the header with height 5 and generates another with height 5. In our case it will create header with height 6. So `next` needs to be in the name. Do you have any other suggestion?

types/src/test_utils.rs
oblique: done

types/src/test_utils.rs
oblique: implemented `another_of` too

zvolin (approved):
Looks good

PR #88, opened 2023-10-02
feat: Store improvements

oblique:
* Return `EmptyStore` error on empty store.
* Optimize failure on appending already known header.
* Remove custom genesis height. Now genesis is always height 1.
* Rename `*_unverified` methods to `*_unchecked`.
* Implement `append_single` and `append`.

### DIFF
--- a/node/src/store.rs
+++ b/node/src/store.rs
@@ -59,7 +59,8 @@ pub trait Store: Send + Sync + Debug {
             Ok(head) => {
                 head.verify(&header)?;
             }
-            Err(StoreError::EmptyStore) => {}
+            // Empty store, we can not verify
+            Err(StoreError::NotFound) => {}
             Err(e) => return Err(e),
         }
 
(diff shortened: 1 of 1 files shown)


oblique:
After a discussion with Celestia team we came into conclusion that custom genesis height must be removed for now.

oblique:
This PR must be merged after #87

node/src/store.rs
```diff
@@ -227,8 +268,8 @@ pub mod tests {
     #[test]
     fn test_empty_store() {
         let s = InMemoryStore::new();
-        assert_eq!(s.get_head_height(), 0);
-        assert!(matches!(s.get_head(), Err(StoreError::NotFound)));
+        assert!(matches!(s.get_head_height(), Err(StoreError::EmptyStore)));
+        assert!(matches!(s.get_head(), Err(StoreError::EmptyStore)));
         assert!(matches!(s.get_by_height(1), Err(StoreError::NotFound)));
```
fl0rek, line 273: If we have separate store error for `EmptyStore`, would it make sense for it to be returned for all operations and let the caller decide whether to handle that differently?

oblique: I think if I was a user I wouldn't like to get `EmptyStore` when I do `get_by_height`. We can remove `EmptyStore` and  `NotFound` only, but we should document this in `get_head`

fl0rek (approved):
:guitar:

PR #89, opened 2023-10-02
feat: Forward only verified and new HeaderSub messages

oblique:


fl0rek (approved):
no notes :guitar:

node/src/p2p.rs
```diff
@@ -547,17 +576,45 @@ where
...
+
+            debug!("New header from header-sub ({header})");
+            *state = Some(header);
+            true
+        });
+
+        if updated {
+            gossipsub::MessageAcceptance::Accept
+        } else {
+            gossipsub::MessageAcceptance::Ignore
```
zvolin, line 616: I think after the `verify` succeeds we should `Accept` unconditionally
https://github.com/celestiaorg/go-header/blob/main/p2p/subscriber.go#L84

oblique: Wouldn't this spam the network?

oblique: Hmm. If you notice the `val` is a callback function and ends up [here](https://github.com/celestiaorg/go-header/blob/e75d00df4c66d32f8528185d9e91273d10db94b5/sync/sync_head.go#L137). Then they [verify it with the current subjective head](https://github.com/celestiaorg/go-header/blob/e75d00df4c66d32f8528185d9e91273d10db94b5/sync/sync_head.go#L168) instead of the `Store` head. Which means older heights are ignored.

I can change the verification to be done on the current subjective head.

zvolin: hmm true. yeah, I think we should verify with syncer's subjective head when syncer is introduced

PR #92, opened 2023-10-04
feat: Add trusted peers and keep track of multiple connections per peer

oblique:


node/src/peer_tracker.rs
```diff
@@ -81,49 +107,81 @@ impl PeerTracker {
...
+            increase_connected_peers(&self.info_tx, peer_info.trusted);
+        }
     }
 
-    pub fn disconnected(&self, peer: PeerId) {
-        let mut state = self.get(peer);
+    /// Inform PeerTracker that a connection of a peer got disconnected
+    ///
+    /// Returns `true` if there are no other connections by the peer
+    pub fn maybe_disconnected(&self, peer: PeerId, connection_id: ConnectionId) -> bool {
```
zvolin, line 142: I think I'd prefer having the name convention in peer tracker like:
- `connect`
- `maybe_connect`
- `discover`
- `maybe_discover`
- `disconnect`
- `maybe_disconnect`

those `ed` suffixes for me looks like I'm querying for the information, not changing the state

zvolin: also `identify`

 + we could then drop `is_` in some places

zvolin: or we could keep the convention from the swarm and then have `on_` prefix

node/src/peer_tracker.rs
```diff
@@ -157,10 +215,40 @@ impl PeerTracker {
...
+    }
 }
 
 impl Default for PeerTracker {
     fn default() -> Self {
         PeerTracker::new()
     }
 }
+
+fn increase_connected_peers(info_tx: &watch::Sender<PeerTrackerInfo>, trusted: bool) {
```
fl0rek, line 236: naming nit: maybe `increment/decrement`, drives home the small, definite (1) amount of change. 
Or maybe change entirely to `notify_peer_connected/disconnected` to focus on the intent rather than implementation.

node/src/peer_tracker.rs
oblique: If we drop the `ed` it will look like we are commanding to connect or disconnect etc. What about `set_` prefix? Like `set_connected`.

node/src/peer_tracker.rs
oblique: Done. I chose `increment/decrement`.

node/src/peer_tracker.rs
```diff
@@ -127,26 +135,30 @@ impl PeerTracker {
...
+        if !peer_info.is_connected() {
+            peer_info.state = PeerState::Connected;
+        }
+
         peer_info.connections.push(connection_id);
 
         // If this is the first connection from the peer
         if peer_info.connections.len() == 1 {
-            increase_connected_peers(&self.info_tx, peer_info.trusted);
+            increment_connected_peers(&self.info_tx, peer_info.trusted);
```
zvolin, line 147: I think this could be merged into the previous `if`

oblique: done

fl0rek (approved):

zvolin (approved):
looks good

PR #95, opened 2023-10-06
fix(node): Remove keep_alive::Behaviour

oblique:
This was preventing `ping::Behaviour` to close a broken connections.

### DIFF
--- a/node/src/p2p.rs
+++ b/node/src/p2p.rs
@@ -17,8 +17,8 @@ use libp2p::{
     multiaddr::Protocol,
     ping,
     swarm::{
-        keep_alive, ConnectionId, DialError, NetworkBehaviour, NetworkInfo, Swarm, SwarmBuilder,
-        SwarmEvent, THandlerErr,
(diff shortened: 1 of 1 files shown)


fl0rek (approved):
:rocket:

zvolin (commented):
Now I'm wondering what to do with the peer tracker. If nodes will be disconnecting in idle, we shouldn't treat the `Identified` state as an upgrade of connected but rather the separate thing, like `trusted`? Also `Identified` shouldn't affect `is_connected`, am I correct?

oblique:
`Identified` means we received what protocols the node supports after the connection. If it gets disconnected the identification get invalidated because we don't know if process was stopped manually to be started in a different version. Versions can support different protocols.

On the other hand, we don't make any use of this information and I think we will never. So we can remove the `Identified` state.

zvolin:
Doesn't identify also provide us listening addresses we put into PeerInfo?

oblique:
True. But we can use `add_addresses` for that.

PR #96, opened 2023-10-06
fix: temporary backward compatibility for json proofs

zvolin:
The  proof representation had breaking change in node v0.11.0-rc14. Previously we were always setting the `ignore_max_ns` to true by hand, but since the rc14, we just use the value provided from json (which hadn't existed before). So to keep the compatibility with an order proofs, we should provide a `true` for this value if missing (when deserializing proofs in older versions)


oblique (approved):
For now we can accept this workaround until testnets are upgraded.

fl0rek (approved):

PR #97, opened 2023-10-06
Implement persistent storage for native builds using sled 

fl0rek:
Implement persistent storage using [`sled`](https://github.com/spacejam/sled) database. 

By default store is created in OS specific cache directory, `--store` flag added to specify path manually.
Add new store errors to reflect errors coming from arbitrary backing storage, representing either recoverable errors (where suspected problem is with data contained) or problems with the backend itself (which are less recoverable).
split `store.rs` into `store` dir for particular implementations and `store.rs` with traits and common code.


node/src/store/sled_store.rs
```diff
@@ -0,0 +1,392 @@
...
+    pub fn new<P>(path: P) -> sled::Result<Self>
+    where
+        P: AsRef<Path>,
+    {
+        let db = sled::open(path)?;
+        Self::init(db)
+    }
+
+    pub fn new_in_memory() -> sled::Result<Self> {
+        let db = sled::Config::default().temporary(true).open()?;
```
zvolin, line 37: this is interesting. So with such config sled will use the `/dev/shm`. I have mixed feelings about that behavior. It sounds cool but I'm not sure if the `/tmp` wouldn't be better match as we expect this to be of non-trivial size.
This on the other hand sounds like what @oblique would love the `InMemoryStore` to behave by still serializing the headers

node/src/store/sled_store.rs
```diff
@@ -0,0 +1,392 @@
...
+        self.db.flush_async().await?;
+
+        Ok(())
+    }
+}
+
+// we can report contained StoreError directly, otherwise transpose Sled error as StoreError
+impl From<TransactionError<StoreError>> for StoreError {
+    fn from(error: TransactionError<StoreError>) -> StoreError {
+        type T = TransactionError<StoreError>;
```
zvolin, line 142: nice, I'd do `E` tho

node/tests/node.rs
zvolin, line 1: please add a test that persistent store persists

zvolin (changes requested):
nice, left some thoughts

node/src/store/sled_store.rs
oblique: I think we shouldn't touch the default dehavior of sled, because it is [implemented in a cross-platform way](https://github.com/spacejam/sled/blob/005c023ca94d424d8e630125e4c21320ed160031/src/config.rs#L412-L417). `/tmp` is not really cross-platform in UNIX systems. For example in macOS it is cleaned periodically, and programs need to use `$TMPDIR` that points to a different directory to avoid that.

node/Cargo.toml
```diff
@@ -30,6 +30,7 @@ tokio = { version = "1.32.0", features = ["macros", "sync"] }
 tracing = "0.1.37"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+sled = "0.34" # upgrade requires migration https://docs.rs/sled/latest/sled/struct.Db.html#examples-1
```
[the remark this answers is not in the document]
oblique: > we will need a convenience script for that

Or we can have a migration procedure at the beginning of `main` and if it fails, we start fresh.

node/src/store/sled_store.rs
zvolin: yeah I didn't want to go with a raw `/tmp` but rather some `tempdir` crate anyway. I'm wondering if ram is what we want

node/Cargo.toml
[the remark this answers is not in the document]
zvolin: guess that could make sense in `celestia` crate, hopefully sled isn't slow to build

celestia/src/native.rs
```diff
@@ -27,6 +28,9 @@ struct Args {
     /// Bootnode
     #[arg(short, long = "bootnode")]
     bootnodes: Vec<Multiaddr>,
+
+    #[arg(short, long = "store")]
```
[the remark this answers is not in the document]
fl0rek: Added doc, but in light of [conversation here](https://github.com/eigerco/celestia-node-rs/pull/97#discussion_r1353263196) maybe this should be removed altogether?

node/src/store/sled_store.rs
fl0rek: done

node/src/store/sled_store.rs
fl0rek: I've added `new_in_memory` mostly with tests in mind, but with platform discrepancies we should probably avoid that as Yiannis mentioned. Changed the API to suggested :+1:

node/tests/node.rs
fl0rek: Added simple persistance tests. 

I was wondering whether we can have reliable crash recovery testing.[ sled does them internally](https://github.com/spacejam/sled/blob/main/tests/test_crash_recovery.rs), but the setup is quite involved (and we'd be mostly re-testing the same). 
I've found also [closefds](https://docs.rs/closefds/latest/closefds/) crate with which I think we could close existing db connection to simulate unexpected close, but this seems finicky especially once more platforms are involved.

Simple solution I initially hoped for (mem::forget on db) doesn't work because we still have db file exclusively open until test process exits.

node/src/store/sled_store.rs
```diff
@@ -0,0 +1,497 @@
...
+            )));
+        };
+        let mut db_path = project_dirs.cache_dir().to_owned();
+        db_path.push(network_id);
+
+        Self::new_in_path(db_path)
+    }
+
+    pub fn new_temp() -> sled::Result<Self> {
+        let tmp_path = TempDir::new("celestia")?.into_path();
```
oblique, line 45: This line and in general all the operations done in `SledStore` are IO operations, which means they are blocking and they will block executor. They need to be wrapped with `spawn_blocking` or `block_in_place`.

`spawn_blocking` is more preferred because if will use a new thread. With `block_in_place` we will restrict user to use only multithreaded Tokio executor.

You will need to have something like the following because `spawn_blocking` needs `'static`:

```rust
struct SledStore {
    inner: Arc<Inner>,
}

struct Inner {
    db: Db,
    headers: Tree,
    height_to_hash: Tree,
}
```

and you will use this like this:

```rust
    pub async fn get_by_height(&self, height: u64) -> Result<ExtendedHeader> {
        let inner = self.inner.clone();
        
        spawn_blocking(move || {
            let hash = read_hash_by_db_key(&inner.height_to_hash, &height_to_key(height))?;
            get_by_hash(&inner.headers, &hash)
        }).await
    }
```

node/src/store/sled_store.rs
```diff
@@ -0,0 +1,497 @@
...
+    pub fn get_by_height(&self, height: u64) -> Result<ExtendedHeader> {
+        let hash = read_hash_by_db_key(&self.height_to_hash, &height_to_key(height))?;
+
+        self.get_by_hash(&hash)
+    }
+
+    pub fn get_head(&self) -> Result<ExtendedHeader> {
+        let head_height = self.head_height()?;
+        self.get_by_height(head_height)
+    }
```
oblique, line 91: Does sled do any caching? If not, then we need to cache lasted appended header and use it as head.

node/src/store/sled_store.rs
```diff
@@ -0,0 +1,497 @@
...
+
+        Self::new_in_path(db_path)
+    }
+
+    pub fn new_temp() -> sled::Result<Self> {
+        let tmp_path = TempDir::new("celestia")?.into_path();
+
+        let db = sled::Config::default()
+            .path(tmp_path)
+            .temporary(true)
```
oblique, line 49: Does `temporary` cleans up the directory in drop? If not, then we need to do it, or keep the `TempDir` in the struct, which does this.

oblique (changes requested):

node/src/store/sled_store.rs
fl0rek: Yes, sled cleans up the directory: https://docs.rs/sled/latest/sled/struct.Config.html#method.temporary

node/src/store/sled_store.rs
```diff
@@ -0,0 +1,497 @@
...
+
+#[derive(Debug)]
+pub struct SledStore {
+    db: Db,
+    headers: Tree,
+    height_to_hash: Tree,
+}
+
+impl SledStore {
+    pub fn new(network_id: &str) -> sled::Result<Self> {
```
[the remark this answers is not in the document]
fl0rek: Eliminated public `sled::Error`s, replacing them with existing, plus added `IoError`

node/src/store/sled_store.rs
fl0rek: Thanks for pointing that out, there's even [note](https://github.com/spacejam/sled#interaction-with-async) on the repo about async :sweat: 
Switched to `spawn_blocking`, reorganising code a bit to avoid spawning threads multiple times per single call.

oblique: Sorry I didn't read their note before. To my understanding all read and write are cached and it is up to a point async friendly. I read some of their code just to make sure.

Based on this facts we will can not trust that blocking will never happen, but because it is going to a rare event (I think), we can switch to `block_in_place` instead. I'm trying to figure out how rare will be in our case. For example when another client tries to sync from us.

fl0rek: Oh, not sweat, I appreciate the comment and agree we should have that fixed. 
In my testing I didn't notice blocking since default cache size of 1GB, which, by the looks of it, translates to roughly 10k headers cached.
Shame there doesn't seem to be a way to get a hint whether header we want is cached, but would it be worthwhile to do some kind of heuristic `block_in_place/spawn_blocking` decision? E.g. we hope last N headers are cached and block in place there, but do spawn_blocking for the rest to accomodate full syncs.
I'll switch to `block_in_place` for now.

celestia/src/native.rs
```diff
@@ -76,7 +87,12 @@ pub async fn run() -> Result<()> {
     node.p2p().wait_connected_trusted().await?;
 
     // We have nothing else to do, but we want to keep main alive
+    let header = node.p2p().get_header_by_height(26625).await?;
+
     loop {
+        let now = std::time::Instant::now();
+        node.p2p().get_verified_headers_range(&header, 512).await?;
+        println!("Took {:?}", now.elapsed());
```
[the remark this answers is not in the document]
fl0rek: `git add` strikes again 😞 removed 👍

celestia/src/native.rs
```diff
@@ -55,6 +58,14 @@ pub async fn run() -> Result<()> {
     let network_id = network_id(args.network).to_owned();
     let genesis_hash = network_genesis(args.network)?;
 
+    let store = if let Some(db_path) = args.store {
+        SledStore::new_in_path(db_path)?
+    } else {
+        SledStore::new(&network_id)?
+    };
+    //let store = celestia_node::store::in_memory_store::InMemoryStore::new();
+    //info!("Initialised store with head: {:?}", store.head_height());
```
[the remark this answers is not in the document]
fl0rek: Accidentaly commited test code when comparing in memory vs sled, removed 👍

node/src/store/sled_store.rs
fl0rek: Per call, leaving it with `block_in_place`. Fixed remaining possible places where we could block.

node/src/store/sled_store.rs
```diff
@@ -0,0 +1,565 @@
...
+    let serialized = tree.get(db_key)?.ok_or(StoreError::NotFound)?;
+
+    ExtendedHeader::decode(serialized.as_ref()).map_err(|e| StoreError::CelestiaTypes(e.into()))
+}
+
+#[inline]
+fn height_to_key(height: u64) -> [u8; 8] {
+    // sled recommends BigEndian representation for ints since it preserves expected int order
+    // when sorted lexicographically
+    height.to_be_bytes()
```
oblique, line 312: interesting. good to know

node/src/store/sled_store.rs
```diff
@@ -0,0 +1,565 @@
...
+                .create_new(true) // make sure we fail if db is already there
+                .open()?;
+            Self::init(db)
+        })
+        .await?
+    }
+
+    pub async fn new_in_path<P>(path: P) -> Result<Self>
+    where
+        P: Into<PathBuf> + Send + 'static,
```
oblique, line 70: The convention is `AsRef<Path>`. After that you can convert it to PathBuf within the method. Remove `Send + 'static` also.

fl0rek: 👍 reverted back to AsRef<Path>

oblique (approved):
lgtm

zvolin (approved):

PR #102, opened 2023-10-12
Implement persistent header storage in browser using IndexedDB

### LINKED ISSUE #60: implement persistent store

fl0rek:
closes #60 
Implement persistent header storage using [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API). 
Also setup brower CI tests with [wasm-pack](https://rustwasm.github.io/wasm-pack/book/). I've had to disable some of the tests for wasm for now as fixing them was bloating this PR.


node/tests/exchange.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(not(target_arch = "wasm32"))]
```
fl0rek, line 1: I've disabled these (and other tests) for wasm as they panic (or don't compile) and fixing them became a bit scope-creepy for this PR.
I'll addresss that separately next.

types/src/time.rs
```diff
@@ -0,0 +1,17 @@
+use instant::SystemTime;
+use tendermint::Time;
+
+// rely on instant to provide time for wasm32
+pub(crate) fn now() -> Time {
```
fl0rek, line 5: Usage of time in tendermint was causing panics so I've added this compatibility layer. Is it the right place? Or maybe this should be moved to our tendermint fork entirely?

node/Cargo.toml
```diff
@@ -4,6 +4,9 @@ version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
 
+[lib]
+crate-type = ["cdylib", "rlib"]
```
oblique, line 8: Is `cdylib` needed because of the tests?

node/src/syncer.rs
```diff
@@ -366,6 +366,7 @@ where
 }
 
 #[cfg(test)]
+#[cfg(not(target_arch = "wasm32"))]
```
oblique, line 369: In which place Syncer tests panic under WASM. I think they shouldn't

fl0rek: It's the sleep in `MockP2pHandle::expect_cmd`, they panic when accessing time: 
```
        panicked at library/std/src/sys/wasm/../unsupported/time.rs:13:9:
        time not implemented on this platform
```

node/Cargo.toml
fl0rek: It's needed for `wasm-pack build`. I see that tests run without it fine, so I'll remove it since we'll need to `build` `celestia`, not `node`.

node/src/syncer.rs
oblique: I see. This is because of `timeout` I used in expect_cmd. You need to create the equivalent function for wasm the same signature and have it under `executor.rs`

types/src/time.rs
oblique: I think since we do not wrap `Time` and replace it in a new `Header`. The best option is to fix it in tendermint fork. Users may want to construct their own `ExtendedHeader` and they will use it.

node/src/syncer.rs
fl0rek: That fixed the `time` panic, additionally I've changed command receiving part of syncer: https://github.com/eigerco/celestia-node-rs/pull/102/files#diff-4d0e726dde3f3ba66c0cd97c8f61c7aa1fc2104ca66e3d04d2bda1e488909930R174-R179

Worker thread wasn't being killed at the end of tests and causing them to never finish. I've changed it so that if command channel is dropped, worked exists.

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,514 @@
...
+}
+
+// SendWrapper usage is safe in wasm because we're running on single thread
+#[derive(Debug)]
+pub struct IndexedDbStore(Arc<SendWrapper<Rexie>>);
+
+impl IndexedDbStore {
+    pub async fn new_with_name(name: &str) -> rexie::Result<Self> {
+        let rexie = Rexie::builder(name)
+            .version(DB_VERSION)
```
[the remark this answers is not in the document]
fl0rek: Yes, rexie has `onupgradeneeded` handler, which then updates db schema https://github.com/devashishdxt/rexie/blob/c6bce1961d101b74b92a01e87f128cbfafa43655/src/rexie_builder.rs#L86

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,514 @@
...
+        Ok(Self(Arc::new(SendWrapper::new(rexie))))
+    }
+
+    pub async fn new() -> rexie::Result<Self> {
+        Self::new_with_name("CELESTIA").await
+    }
+
+    pub async fn clear(self) -> rexie::Result<()> {
+        Rexie::delete(&self.0.name()).await
+    }
```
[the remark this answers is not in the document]
fl0rek: Renamed to `delete_db` since what's that it does. And [indeed, we should](https://github.com/devashishdxt/rexie/issues/32). Fixed this method and added tests for it.

node/src/executor.rs
```diff
@@ -67,7 +78,7 @@ mod imp {
         pub(crate) async fn new(dur: Duration) -> Self {
             // If duration was less than a millisecond, then make
             // it 1 millisecond.
-            let millis = u32::try_from(dur.as_millis().max(1)).unwrap_or(u32::MAX);
+            let millis = u32::try_from(dur.as_millis().max(100)).unwrap_or(u32::MAX);
```
oblique, line 79: Why did you change this to 100?

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,514 @@
...
+            .0
+            .transaction(&[HEADER_STORE_NAME], TransactionMode::ReadOnly)?;
+
+        let header_store = tx.store(HEADER_STORE_NAME)?;
+        let header_get_result = header_store.get_all(None, Some(1), None, Some(Direction::Prev));
+
+        let header_get_result = header_get_result.await?;
+        let key_value_result = header_get_result.first().ok_or(StoreError::NotFound)?;
+
+        let header_entry: ExtendedHeaderEntry = from_value(key_value_result.1.clone())?;
```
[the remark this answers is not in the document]
fl0rek: switched to functional style here 👍

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,514 @@
...
+    pub async fn clear(self) -> rexie::Result<()> {
+        Rexie::delete(&self.0.name()).await
+    }
+
+    pub async fn get_head(&self) -> Result<ExtendedHeader> {
+        let tx = self
+            .0
+            .transaction(&[HEADER_STORE_NAME], TransactionMode::ReadOnly)?;
+
+        let header_store = tx.store(HEADER_STORE_NAME)?;
```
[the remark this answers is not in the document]
fl0rek: I think this (and Arc) is leftover from trying to fix most of the types here not being `Send`. `SendWrapper` dealt with that, so it's not necessary now.

node/src/syncer.rs
```diff
@@ -170,8 +171,12 @@ where
...
                     self.fetch_next_batch().await;
                 }
-                Some(cmd) = self.cmd_rx.recv() => {
-                    self.on_cmd(cmd).await;
+                cmd = self.cmd_rx.recv() => {
+                    if let Some(cmd) = cmd {
+                        self.on_cmd(cmd).await;
+                    } else {
+                        return;
+                    }
```
oblique, line 179: I prefer [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) for graceful shutdown. We should use that in `Drop` impls etc. Also for each task. I have a separate issue for that (https://github.com/eigerco/celestia-node-rs/issues/70).

node/src/executor.rs
fl0rek: That was change done for testing I've pushed inadvertently, reverted.

node/src/store/indexed_db_store.rs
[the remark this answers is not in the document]
oblique: nice!

node/src/exchange/.client.rs.swo
oblique, line 1: leftover file from your editor

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,533 @@
...
+                    .add_index(Index::new(HEIGHT_INDEX_NAME, "height").unique(true)),
+            )
+            .build()
+            .await?;
+        Ok(Self(SendWrapper::new(rexie)))
+    }
+
+    pub async fn new(network_id: &str) -> rexie::Result<Self> {
+        Self::new_with_name(&network_id).await
+    }
```
oblique, line 45: I think this needs to be removed and rename `new_with_name` to `new`

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,533 @@
...
+    hash: Hash,
+    header: ExtendedHeader,
+}
+
+// SendWrapper usage is safe in wasm because we're running on single thread
+#[derive(Debug)]
+pub struct IndexedDbStore(SendWrapper<Rexie>);
+
+impl IndexedDbStore {
+    pub async fn new_with_name(name: &str) -> rexie::Result<Self> {
```
oblique, line 28: Return `StoreError`. Maybe something like `OpenFailed` to be generic and also adjust the constructor of `SledStore` to use it.

node/src/exchange/.client.rs.swo
fl0rek: removed 👍

node/Cargo.toml
```diff
@@ -48,10 +48,21 @@ libp2p = { version = "0.52.3", features = [
...
 send_wrapper = { version = "0.6.0", features = ["futures"] }
+serde = { version = "1.0.164", features = ["derive"] }
+serde-wasm-bindgen = "0.6.0"
+wasm-bindgen = "0.2.87"
+wasm-bindgen-futures = "0.4.37"
 libp2p = { version = "0.52.3", features = ["webtransport-websys"] }
+pin-project = "1.1.3"
+
+[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+tokio = { version = "1.32.0", features = ["time" ] }
```
oblique, line 62: This line is not needed

types/src/time.rs
fl0rek: Moved the fix to tendermint: https://github.com/eigerco/celestia-tendermint-rs/pull/13

node/src/store/indexed_db_store.rs
fl0rek: Removed, left just single `fn new(network_id: &str)` constructor.

node/src/store/indexed_db_store.rs
fl0rek: Replaced crate specific errors with `StoreError::OpenFailed` with description here and used that in `SledStore`.

node/Cargo.toml
fl0rek: yes, removed

node/src/syncer.rs
fl0rek: Removed this closed channel detection and replaced it with `CancellationToken` and `DropGuard` in `Syncer`

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,514 @@
...
+        let key_value_result = header_get_result.first().ok_or(StoreError::NotFound)?;
+
+        let header_entry: ExtendedHeaderEntry = from_value(key_value_result.1.clone())?;
+
+        Ok(header_entry.header)
+    }
+
+    pub async fn get_head_height(&self) -> Result<u64> {
+        self.get_head().await.map(|h| h.height().value())
+    }
```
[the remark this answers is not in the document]
fl0rek: I don't think we can rely on that, added head header caching in `IndexedDbStore` 👍 
I've added few optimisations for cases where we don't need to look into the db, like `contains_height`, similar to what we have in `InMemoryStore`.

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,550 @@
...
+            )
+            .build()
+            .await
+            .map_err(|e| StoreError::OpenFailed(e.to_string()))?;
+
+        let db_head = match get_head_from_database(&rexie).await {
+            Ok(v) => Ok(Some(v)),
+            Err(StoreError::NotFound) => Ok(None),
+            Err(e) => Err(e),
+        }?;
```
oblique, line 53:
```suggestion
        let db_head = match get_head_from_database(&rexie).await {
            Ok(v) => Ok(Some(v)),
            Err(StoreError::NotFound) => Ok(None),
            Err(e) => return Err(e),
        };
```

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,550 @@
...
+
+const DB_VERSION: u32 = 1;
+const HEADER_STORE_NAME: &str = "headers";
+const HASH_INDEX_NAME: &str = "hash";
+const HEIGHT_INDEX_NAME: &str = "height";
+
+#[derive(Debug, Serialize, Deserialize)]
+struct ExtendedHeaderEntry {
+    height: u64,
+    hash: Hash,
```
oblique, line 23: As I understood these two values are never used. Am I correct? If yes, remove them.

node/src/store/indexed_db_store.rs
```diff
@@ -0,0 +1,550 @@
...
+    header: Vec<u8>,
```
oblique, line 24: Can you test if the space is reduced if we use base64? Because if the value is save as a string instead of blob it will take much more space.

types/src/time.rs
oblique, line 1: This file is not needed anymore

node/src/store/indexed_db_store.rs
fl0rek: It is being saved as an array of bytes

types/src/time.rs
fl0rek: yes, thanks; removed 👍

PR #104, opened 2023-10-13
feat: Improve performance of Exchange

oblique:
We sacrifice memory usage for performance.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -591,7 +591,6 @@ dependencies = [
  "tendermint-proto",
  "thiserror",
  "tokio",
- "tokio-util",
  "tracing",
  "wasm-bindgen-futures",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -532,9 +532,9 @@ checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
 
 [[package]]
 name = "bytes"
-version = "1.4.0"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


zvolin (approved):
cool :fire:

zvolin (changes requested):
Miri has no issues with it but I'm concerned about the transmute.

In the MaybeUninit docs there is this example (3rd in section)
https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initialization-invariant
```rs
let x: i32 = unsafe { MaybeUninit::uninit().assume_init() }; // undefined behavior! ⚠️
```
so I think it would also do so for
```rs
let x: MaybeUninit<u8> = MaybeUninit::uninit();
let x: u8 = std::mem::transmute(x); // ub
```

also I found this `MaybeUninit::slice_assume_init_mut` which happens to do the same thing and it says that calling is unsafe on uninitialized data
https://doc.rust-lang.org/src/core/mem/maybe_uninit.rs.html#976

Want to open a discussion what do we want to do with it. It seems it's safe by *current* compiler's behavior and used in some popular crates but causes UB per documentation.

oblique:
I posted a question about this in the Rust forum ([link](https://users.rust-lang.org/t/is-transmuting-mut-maybeuinit-u8-to-mut-u8-an-ub-in-my-code/101164)) and the code is safe as long as `Stream` only fills the slice. If `Stream::poll_read` try to read a byte from the slice then it is UB.

I have checked all the `Stream` types we use (tokio-tcp, noise, yamux, quic) and they only fill the slice. But for the sake of peace of mind in case future releases of the mentioned types, I'm going to remove the unsafe. By removing a ~1ms of overhead will be added.

fl0rek (approved):
great work! 🚀

PR #112, opened 2023-10-20
feat(node): Implement running node in browser

### LINKED ISSUE #109: Run a node with syncer and header store in browser
Now that we have:
- exchange client/server
- IndexedDb store
- syncer

make sure this "very light node" compiles and runs in browser as a whole

fl0rek:
closes #109 

Add `wasm-node` which implements similar interface to `Node` but in JS.

One can run local server to serve it with command:
```
wasm-pack build --target web node-wasm --debug && cargo run -- --browser --network mocha
```

Then node can be accessed on `localhost:9876`. There's a preliminary web gui interface, but you'll probably want to look at console logs (F12 on chrome) to monitor node's progress.


types/src/network.rs
oblique, line 1: I prefer this under `node` crate instead

types/src/network.rs
```diff
@@ -0,0 +1,94 @@
...
+            .map(|s| s.parse().unwrap())
+            .collect()
+        ),
+        Network::Mocha => Ok(
+            [
+            "/ip4/40.85.94.176/tcp/2121/p2p/12D3KooWNJ3Nf1DTQTz8JZogg2eSvPKKKv8itC6fxxspe4C6bizs",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/p2p/12D3KooWNJ3Nf1DTQTz8JZogg2eSvPKKKv8itC6fxxspe4C6bizs",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/webtransport/certhash/uEiBf-OX4HzFK9owOpjdCifsDIWRO0SoD3j3vGKlq0pAXKw/certhash/uEiCx1md1BATJ_0NXAjp3KOuwRYG1535E7kUzFdMq8aPaWw/p2p/12D3KooWNJ3Nf1DTQTz8JZogg2eSvPKKKv8itC6fxxspe4C6bizs",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/p2p/12D3KooWQUYAApYb4DJnhS1QmAwRr5HRvUeHJYocchCpwEhCtDGu",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/webtransport/certhash/uEiBr4-sr95BpqfA-ttpjiLdjbGABhTvX8oxrTXf3Ubfibw/certhash/uEiBSVgyze9xG1UbbNuTwyEUWLPq7l2N9pyeQSs3OtEhGRg/p2p/12D3KooWQUYAApYb4DJnhS1QmAwRr5HRvUeHJYocchCpwEhCtDGu",
```
oblique, line 82: Add a comment that these are not official bootnotes, and there are going to be removed later on. Or remove them from here, and append them to the vec after `canonical_network_bootnodes` is used.

types/src/network.rs
fl0rek: Right, probably shouldn't pollute `types` with wasm dependencies, moved.

types/src/network.rs
fl0rek: Kept only official nodes here and moved additional addresses to `wasm-node/utils` as these are currently required in browser only.

celestia/src/lib.rs
```diff
@@ -1,8 +1,3 @@
-#[cfg(target_arch = "wasm32")]
-mod wasm;
-#[cfg(target_arch = "wasm32")]
-pub use crate::wasm::*;
-
 #[cfg(not(target_arch = "wasm32"))]
```
zvolin, line 1: maybe in this case just add this at the top?
```rs
#![cfg(not(target_arch = "wasm32"))]
```

node/Cargo.toml
```diff
@@ -10,8 +10,10 @@ celestia-types = { workspace = true }
 tendermint-proto = { workspace = true }
 
 async-trait = "0.1.73"
+clap = { version = "4.4.4", features = ["derive"] }
```
zvolin, line 13: not sure if I'm happy with clap in a node unconditionally? wdyt

wasm-node/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.34"
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+axum = "0.6.20"
+clap = { version = "4.4.4", features = ["derive"] }
+dotenvy = "0.15.7"
+mime_guess = "2.0"
+rust-embed="8.0.0"
```
zvolin, line 35:
```suggestion
rust-embed = "8.0.0"
```

wasm-node/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+serde = "1.0.189"
+serde-wasm-bindgen = "0.6.0"
+tracing = "0.1.37"
+tracing-wasm = "0.2.1"
+wasm-bindgen = "0.2.87"
+wasm-bindgen-futures = "0.4.37"
+web-sys = { version = "0.3.64", features = ["console"]}
+serde_json = "1.0.107"
+
+[dev-dependencies]
```
zvolin, line 27: can we have those after non-dev deps?

wasm-node/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+celestia-node = { workspace = true, features = ["test-utils"] }
+celestia-types = { workspace = true, features = ["test-utils"] }
+
+anyhow = "1.0.71"
+console_error_panic_hook = "0.1.7"
+futures-timer = { version = "3.0.2", features = ["wasm-bindgen"]}
```
zvolin, line 15:
```suggestion
futures-timer = { version = "3.0.2", features = ["wasm-bindgen"] }
```

wasm-node/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+js-sys = "0.3.64"
+libp2p = "0.52.3"
+serde = "1.0.189"
+serde-wasm-bindgen = "0.6.0"
+tracing = "0.1.37"
+tracing-wasm = "0.2.1"
+wasm-bindgen = "0.2.87"
+wasm-bindgen-futures = "0.4.37"
+web-sys = { version = "0.3.64", features = ["console"]}
```
zvolin, line 24:
```suggestion
web-sys = { version = "0.3.64", features = ["console"] }
```

wasm-node/rsc/index.html
zvolin, line 1: what does rsc mean?

wasm-node/rsc/run_node.js
```diff
@@ -0,0 +1,44 @@
+Error.stackTraceLimit = 99;
+
+import init, { setup, Network, WasmNode, WasmNodeConfig, canonical_network_bootnodes, network_genesis } from "/wasm/wasm_node.js";
+
+// initialize wasm
+await init();
+// setup logging and console panic hook
+await setup();
```
zvolin, line 8: can we rename `setup` to `configure_logging` or setup_logging or whatever that will be more precise in what it does? I think we can treat `console_panic_hook` as part of `logging` in this case

node/src/network.rs
```diff
@@ -0,0 +1,97 @@
...
+use serde_repr::{Deserialize_repr, Serialize_repr};
+use thiserror::Error;
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen::prelude::*;
+
+#[cfg_attr(not(target_arch = "wasm32"), derive(ValueEnum))]
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize_repr, Deserialize_repr)]
+#[repr(u8)]
+pub enum Network {
```
zvolin, line 18: btw celestia introduced mainnet already, but we can be slow to the party I guess as it hasn't started yet
https://github.com/celestiaorg/celestia-node/blob/main/nodebuilder/p2p/bootstrap.go#L40

wasm-node/rsc/run_node.js
```diff
@@ -0,0 +1,44 @@
...
+const response = await fetch('/cfg.json');
+const json = await response.json();
+
+console.log("Received config:", json);
+
+const network = json.network;
+let bootnodes = json.bootnodes
+if (bootnodes.length === 0) {
+    bootnodes = canonical_network_bootnodes(network);
+}
```
zvolin, line 19:
```suggestion
const bootnodes = json.bootnodes
if (bootnodes.length === 0) {
    bootnodes.push(...canonical_network_bootnodes(network));
}
```

wasm-node/rsc/run_node.js
```diff
@@ -0,0 +1,44 @@
...
+    const genesis = document.getElementById("genesis").value;
+    const bootnodes  = document.getElementById("bootnodes").value.split("\n");
+
+    console.log("starting with:", network, bootnodes, genesis);
+
+    const config = new WasmNodeConfig(network, genesis, bootnodes);
+    window.node = await new WasmNode(config);
+
+    setInterval(async function() {
+        document.getElementById("peer_id").innerText = JSON.stringify(await window.node.local_peer_id());
```
zvolin, line 39: this could be set once before the interval I think

wasm-node/rsc/run_node.js
```diff
@@ -0,0 +1,44 @@
...
+    bootnodes = canonical_network_bootnodes(network);
+}
+const genesis = network_genesis(network);
+
+document.getElementById("network_id").value = network;
+document.getElementById("genesis").value = genesis;
+document.getElementById("bootnodes").value = bootnodes.join("\n");
+
+document.getElementById("start").addEventListener("click", async function(ev) {
+    document.getElementById("start").setAttribute("disabled", "disabled");
```
zvolin, line 27: wouldn't this do?
```suggestion
    document.getElementById("start").disabled = true;
```

wasm-node/rsc/run_node.js
```diff
@@ -0,0 +1,44 @@
...
+
+    const network = Number(document.getElementById("network_id").value);
+    const genesis = document.getElementById("genesis").value;
+    const bootnodes  = document.getElementById("bootnodes").value.split("\n");
```
zvolin, line 31: I think I'd prefer having this values set on `window.node_config` object, which would provide values for both divs and the script, rather than setting and parsing just from divs

wasm-node/src/axum_server.rs
zvolin, line 1: I think `axum_` prefix is not relevant

zvolin: let's not have to rename the file if we change our backend

wasm-node/src/axum_server.rs
```diff
@@ -0,0 +1,96 @@
...
+        .route("/js/*script", get(serve_embedded_path::<StaticResources>))
+        .route("/wasm/*binary", get(serve_embedded_path::<WasmPackage>))
+        .route("/cfg.json", get(serve_config))
+        .with_state(state);
+
+    spawn(axum::Server::bind(&args.listen_addr).serve(app.into_make_service()));
+
+    loop {
+        time::sleep(time::Duration::from_secs(1)).await;
+    }
```
zvolin, line 68: why do we spawn + infinite loop instead of just awaiting the serve?

wasm-node/src/axum_server.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    let state = WasmNodeArgs {
+        network: args.network,
+        bootnodes: args.bootnode,
+    };
+
+    let app = Router::new()
+        .route("/", get(serve_index_html))
+        .route("/js/*script", get(serve_embedded_path::<StaticResources>))
+        .route("/wasm/*binary", get(serve_embedded_path::<WasmPackage>))
```
zvolin, line 60: why do we have those `*script` and `*binary`?

wasm-node/src/axum_server.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+async fn serve_config(state: State<WasmNodeArgs>) -> Result<Response, StatusCode> {
+    let args = serde_json::to_string(&state.0).map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    Response::builder()
+        .header(header::CONTENT_TYPE, "application/json")
+        .body(body::boxed(body::Full::from(args)))
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)
```
zvolin, line 95: there should be a `Json` response wrapper in the `axum`

wasm-node/src/axum_server.rs
```diff
@@ -0,0 +1,96 @@
...
+        .with_state(state);
+
+    spawn(axum::Server::bind(&args.listen_addr).serve(app.into_make_service()));
+
+    loop {
+        time::sleep(time::Duration::from_secs(1)).await;
+    }
+}
+
+async fn serve_index_html() -> Result<impl IntoResponse, StatusCode> {
```
zvolin, line 71:
```suggestion
async fn serve_index_html() -> Result<Response, StatusCode> {
```

wasm-node/src/node.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    pub async fn listeners(&self) -> Array {
+        self.node
+            .p2p()
+            .listeners()
+            .await
+            .unwrap_throw()
+            .iter()
+            .map(ToString::to_string)
```
zvolin, line 111: what would happen if we skip to_string?

wasm-node/src/node.rs
```diff
@@ -0,0 +1,153 @@
...
+        let syncer_info = self.node.syncer().info().await.unwrap_throw();
+        to_value(&syncer_info).unwrap_throw()
+    }
+}
+
+#[wasm_bindgen]
+impl WasmNodeConfig {
+    #[wasm_bindgen(constructor)]
+    pub fn new(network: Network, genesis_hash: JsString, bootnodes: Vec<JsString>) -> Self {
+        let p2p_local_keypair = identity::Keypair::generate_ed25519(); // TODO
```
zvolin, line 138: what has to be done for this todo?

node/Cargo.toml
fl0rek: True, especially since it's just a single `enum Network`. I'm thinking about duplicate `Network` dedicated just for arg-parsing in `celestia` with `From` and `Into` to keep the one in `node` pure. Alternatively we could feature gate it behind `native-node` or similar.

zvolin: I second the duplication of this enum, that was my first thought but didn't want to push it right away :)

wasm-node/rsc/index.html
zvolin: I have never seen that in a wild, what would you say for just `static/`? It seems pretty common

wasm-node/rsc/run_node.js
zvolin: ah you missunderstood me I thought about bringing the `peer_id` out of the loop completely as it doesn't change through the node's lifetime. I'm fine with having other stuff updated after the interval for simplicity but up to you

wasm-node/rsc/run_node.js
zvolin: I actually think that setting .disabled to "disabled" was a bit hacky as it worked because the string "disabled" is truthy :)

wasm-node/rsc/run_node.js
zvolin: > example/proof of concept
that's correct, however I think we will be working with it for some time so the cleaner the better anyway

I don't think you will need any synchronization. You can have the function that will read the configuration and put it in the object and call it in the beginning of the script. It's fine relying on the fact that the config never changes after that.

in general in regards to javascript, I'd personally still keep such scripts structured so that you have:
```js
import init, { ... } from "/wasm/wasm_node.js";

async function fetch_config() {
    config = await fetch(...)
    ...
    return new WasmNodeConfig(config)
}

async function display_config(config) {
    // here you update dom with config values so that user can see them
    // also only here will it be translated to plain strings, stuff like join("\n") etc
}

async function start_node(config) { // here it will take ready WasmNodeConfig
    ...
}

async fn main() {
    await init()
    await setup()
    const config = fetch_config()
    show_config(config)
    await start_node(config)
}

await main()
```
more or less?

wasm-node/src/node.rs
zvolin: currently I'm fine with rolling a new one each boot, we can improve later on if needed but feel free to open an issue for discussion

wasm-node/rsc/run_node.js
zvolin: ah, I missed the fact that as a user I can change values in those divs

wasm-node/rsc/run_node.js
fl0rek: Thanks for clarification, it makes sense now. PeerId indeed isn't changing so we can set once 👍

wasm-node/src/node.rs
fl0rek: Opened https://github.com/eigerco/celestia-node-rs/issues/116 for further tracking

wasm-node/rsc/index.html
fl0rek: `static` works for me, changed

wasm-node/src/axum_server.rs
fl0rek: Route without wildcard isn't interpreted as prefix, but rather as an [entire path](https://docs.rs/axum/latest/axum/routing/struct.Router.html#static-paths), so `/js/run_node.js` path wouldn't match `/js/` route here.
Ultimately I went with using `*path` as a wildcard name for both, seemed most generic and fitting, since we then we deconstruct it with axum's `Path` extractor.

node/Cargo.toml
fl0rek: Separated the two, so that `ArgNetwork` is responsible for the argument parsing and passing side, while `Network` in `node` references the network in node's context.

celestia/static/run_node.js
```diff
@@ -0,0 +1,74 @@
...
+        document.querySelectorAll('.config').forEach(function(element) {
+            element.disabled = true
+        });
+
+        start_node(window.config);
+    });
+
+    await show_stats(window.node);
+    setInterval(async function() { await show_stats(window.node) }, 1000)
+}(document, window);
```
zvolin, line 60: does this call main right away? if so, are the rest of functions interpreted? I'd rather prefer regular call at the end of file to not wonder

wasm-node/src/node.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub async fn syncer_info(&self) -> JsValue {
+        let syncer_info = self.node.syncer().info().await.unwrap_throw();
+        to_value(&syncer_info).unwrap_throw()
+    }
+}
+
+#[wasm_bindgen]
+impl WasmNodeConfig {
+    #[wasm_bindgen(constructor)]
+    pub fn new(network: Network, genesis_hash: JsString, bootnodes: Vec<JsString>) -> Self {
```
oblique, line 136: I feel these are not needed since we have setters. What is the idiomatic way in JS?

celestia/static/run_node.js
```diff
@@ -0,0 +1,74 @@
...
+
+    document.getElementById("peers").replaceChildren(peers_ul);
+}
+
+
+function bind_config() {
+    // TODO two way binding between window.config and input values
+}
+
++async function main(document, window, undefined) {
```
oblique, line 42: What is the `+` here?

celestia/static/run_node.js
oblique: I was trying to find the closing brackets of the function 😅, then I realized that it is actually calling `main`

node/src/network.rs
```diff
@@ -0,0 +1,93 @@
...
+
+impl FromStr for Network {
+    type Err = UnknownNetworkError;
+
+    fn from_str(network_id: &str) -> Result<Self, Self::Err> {
+        match network_id {
+            "arabica-10" => Ok(Network::Arabica),
+            "mocha-4" => Ok(Network::Mocha),
+            "private" => Ok(Network::Private),
+            network => Err(UnknownNetworkError(network.to_string())),
```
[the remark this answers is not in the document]
zvolin: this should be separate change. We can use basically any name for the "private" but we need to persist this information for correct protocol's naming in node

wasm-node/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+libp2p = "0.52.3"
+serde = "1.0.189"
+serde-wasm-bindgen = "0.6.0"
+serde_repr = "0.1"
+tracing = "0.1.37"
+tracing-wasm = "0.2.1"
+wasm-bindgen = "0.2.87"
+wasm-bindgen-futures = "0.4.37"
+web-sys = { version = "0.3.64", features = ["console"] }
+serde_json = "1.0.107"
```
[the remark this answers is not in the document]
zvolin: we should have ci check for that :+1:

wasm-node/src/node.rs
```diff
@@ -0,0 +1,179 @@
...
+        let p2p_local_keypair = identity::Keypair::generate_ed25519();
+
+        let p2p_bootnodes: Vec<Multiaddr> = bootnodes
+            .iter()
+            .map(|addr| addr.as_string().unwrap_throw().parse().unwrap_throw())
```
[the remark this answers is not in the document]
zvolin: I think returning the `Result<T, JsValue>` works the same and we could use `?` too, but more type conversions could be needed

wasm-node/src/node.rs
zvolin: this is for the `new WasmNodeConfig(...)` syntax
I'm not sure what the difference is with wasm bindgen, but in javascript it will have a separation of object and instance, eg. the methods will be stored once in the object and shared by instances, if you would just return objects without new then they would allocate a separated function each

celestia/static/run_node.js
zvolin: I think the `run_node.js` is a bit drafty now :)

node/src/network.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let bytes = hex::decode(hex)?;
+    let array = bytes
+        .try_into()
+        .map_err(|b: Vec<u8>| NetworkGenesisError::InvalidLengthError(b.len()))?;
+
+    Ok(Some(Hash::Sha256(array)))
+}
+
+pub fn canonical_network_bootnodes(network: Network) -> Result<Vec<Multiaddr>, MultiaddrError> {
```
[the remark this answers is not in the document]
fl0rek: I think that's a good idea, it'll cut down on unnecessary error handling downstream. Removed `Result` from return type and added tests.

wasm-node/Cargo.toml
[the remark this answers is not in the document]
fl0rek: Thanks for noting that!
For tools, I've tried [`udep`](https://lib.rs/cargo-udeps) and discovered [`machete`](https://github.com/bnjbvr/cargo-machete) this morning. Former doesn't seem to flag dependencies we could remove but that are used by our dependencies though. Second one gave me more accurate results and has convenient option for flagging false positives (which they both have..).
Created https://github.com/eigerco/celestia-node-rs/issues/118 to add CI check

wasm-node/src/node.rs
fl0rek: Constructor here will add, as I understand it, JS's idomatice `new ...` way to create class instance, as @zvolin mentioned.
I've added getters and setters so that js code can interact with config like a normal js object. Most of the fields aren't directly representable in js, so I had to `#[wasm_bindgen(skip)]` them, which resulted in opaque class without members.

celestia/Cargo.toml
```diff
@@ -13,13 +13,18 @@ celestia-rpc = { workspace = true, features = ["p2p"] }
...
+libp2p = "0.52.3"
+mime_guess = "2.0"
+rust-embed = "8.0.0"
+serde = "1.0.189"
+serde_repr = "0.1"
 tracing = "0.1.37"
 tracing-appender = "0.2.2"
-tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
```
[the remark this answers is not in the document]
fl0rek: Switched so that dependencies are declared for not-wasm only  👍 + removed the lib

node/src/syncer.rs
```diff
@@ -73,7 +74,7 @@ enum SyncerCmd {
     },
 }
 
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
 pub struct SyncingInfo {
```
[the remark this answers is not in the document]
fl0rek: Yes, we basically have 2 ways of passing custom types back to js:
- `#[wasm_bindgen]` the type and then we can return it directly
- serialize it with `wasm-bindgen-serde` and return it as `JsValue`, but this requires it to be `Serialize`able
analogous for getting data from js

celestia/static/run_node.js
fl0rek: https://github.com/eigerco/celestia-node-rs/pull/112#discussion_r1374147836 
yeah, this seems more trouble than it's worth, switching to normal fn invocation 👍

wasm-node/rsc/run_node.js
oblique: I was also thinking about a single `init`, that will do more stuff in the future.

wasm-node/Cargo.toml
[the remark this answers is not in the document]
fl0rek: Ah, actually - I've added the `futures-timer` not directly for our sake, but because it's used in `libp2p_autonat::behaviour`, but libp2p doesn't set `wasm-bindgen` feature. This causes futures-timer to use `std::time`, which panics in browser.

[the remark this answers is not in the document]
zvolin: that sounds like an issue in libp2p, you could go for the first contribution :)

celestia/static/run_node.js
oblique: 🤯

wasm-node/Cargo.toml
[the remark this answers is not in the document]
fl0rek: actually nr2: looks like libp2p has `wasm-bindgen` feature that should be set so that it can forward it to its dependencies. Added that and removed `futures-timer` 👍

wasm-node/src/lib.rs
```diff
@@ -2,3 +2,4 @@
 
 pub mod node;
 pub mod utils;
+pub mod wrapper;
```
oblique, line 5: I don't think user needs to know that it is wrapped or from which library is wrapped. We can reexport the types in `node` module

fl0rek: true, removed pub here. 

wasm-bindgen does seem to care only about type visibility and doesn't evaluate whether it's entire path is reachable, so we don't need to re-export it for js sake, and I hope we won't use them in other rust crates :p 
Also note that in exported wasm module everything is dumped in top scope, so our internal layout isn't reflected.

celestia/src/common.rs
```diff
@@ -0,0 +1,126 @@
...
+use serde_repr::{Deserialize_repr, Serialize_repr};
+
+use crate::{native, server};
+
+const SERVER_DEFAULT_BIND_ADDR: &str = "127.0.0.1:9876";
+
+#[derive(Debug, Parser)]
+// disallow specifying store path when running in browser
+#[clap(group(ArgGroup::new("native_xor_browser")
+             .args(&["store", "browser"])))]
```
zvolin, line 18: you can do this instead. I think it's a bit cleaner
```rs
    /// Persistent header store path.
    #[arg(short, long = "store", group = "browser_xor_store")]
    pub(crate) store: Option<PathBuf>,

    /// Serve wasm node which can be accessed with web browser
    #[arg(long, group = "browser_xor_store")]
    browser: bool,
```

celestia/src/common.rs
```diff
@@ -0,0 +1,126 @@
...
+            0 => SERVER_DEFAULT_BIND_ADDR.parse().unwrap(),
+            1 => multiaddr_to_socketaddr(args.listen_addrs.pop().unwrap())?,
+            _ => Args::command()
+                .error(
+                    ErrorKind::TooManyValues,
+                    "expected single listenting address when serving wasm node",
+                )
+                .exit(),
+        };
+        server::run(args.network, args.bootnodes, listen_addr).await
```
zvolin, line 68: tbh I'm not a big fan of this and multiaddr_to_socketaddr parsing. What would you say for subcommands? It's fine for me for duplicating some arguments but I think there is also a way to pass common ones.

```rs
#[derive(Parser)]
#[command(author, version, about, long_about = None)]
enum TestSuite {
    Node(NodeArgs),
    Browser(BrowserArgs),
}

#[derive(Args)]
struct NodeArgs {
    ...
}
```

It could then be used as
```bash
celestia node -n mocha
celestia browser -n mocha
```

and you can then have `listen_addr` as `Vec<Multiaddress>` for node and `SocketAddrV4` for browser

celestia/src/common.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// Serve wasm node which can be accessed with web browser
+    #[arg(long)]
+    browser: bool,
+}
+
+#[derive(
+    Debug, Default, Clone, Copy, PartialEq, Eq, ValueEnum, Serialize_repr, Deserialize_repr,
+)]
+#[repr(u8)]
```
zvolin, line 44: why do we need `repr(u8)` there now and `Serialize_repr`, `Deserialize_repr`? Previously it was working fine with
```rs
#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, ValueEnum)]
enum Network {
    Arabica,
    Mocha,
    #[default]
    Private,
}
```

celestia/src/native.rs
```diff
@@ -3,59 +3,35 @@ use std::path::PathBuf;
...
-use clap::{Parser, ValueEnum};
 use libp2p::{identity, multiaddr::Protocol, Multiaddr};
 use tokio::time::sleep;
 use tracing::info;
 
-#[derive(Debug, Parser)]
-struct Args {
-    /// Network to connect.
-    #[arg(short, long, value_enum, default_value_t)]
+pub(crate) async fn run(
```
zvolin, line 14: with subcommands this could be a method on `NodeArgs` and avoid passing arguments by hand too

celestia/static/run_node.js
```diff
@@ -0,0 +1,77 @@
...
+function show_config(config) {
+    document.getElementById("network_id").value = config.network;
+    document.getElementById("genesis").value = config.genesis_hash;
+    document.getElementById("bootnodes").value = config.bootnodes.join("\n");
+}
+
+async function start_node(config_json) {
+    const config = new NodeConfig(config_json.network, config_json.genesis, config_json.bootnodes);
+
+    window.node = await new Node(config);
```
zvolin, line 51: this is already a `NodeConfig`
```suggestion
async function start_node(config) {
    window.node = await new Node(config);
```

celestia/static/run_node.js
```diff
@@ -0,0 +1,77 @@
...
+
+    document.getElementById("start").addEventListener("click", async function(ev) {
+        document.querySelectorAll('.config').forEach(function(element) {
+            element.disabled = true
+        });
+
+        start_node(window.config);
+    });
+
+    await show_stats(window.node);
```
zvolin, line 72: I think this isn't needed because it'll likely start executing the interval before the click event is sent to the listener

wasm-node/pkg/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
zvolin, line 1: I think the correct way is to do
```suggestion
*
!/.gitignore
```
I think this works because you somehow already had this `.gitignore` file in git's index. However if you'd try to do
```bash
mkdir foo
echo '*' > foo/.gitignore
# those won't work
git add foo
git add foo/.gitignore
```

wasm-node/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+crate-type = ["cdylib", "rlib"]
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+celestia-node = { workspace = true }
+celestia-types = { workspace = true }
+
+anyhow = "1.0.71"
+console_error_panic_hook = "0.1.7"
+js-sys = "0.3.64"
+libp2p = "0.52.3"
```
zvolin, line 16: should we maybe specify libp2p version on workspace level?

wasm-node/src/node.rs
```diff
@@ -0,0 +1,179 @@
...
+
+#[wasm_bindgen(js_name = NodeConfig)]
+pub struct WasmNodeConfig {
+    pub network: Network,
+    #[wasm_bindgen(skip)]
+    pub genesis_hash: Option<Hash>,
+    #[wasm_bindgen(skip)]
+    pub p2p_local_keypair: Keypair,
+    #[wasm_bindgen(skip)]
+    pub p2p_bootnodes: Vec<Multiaddr>,
```
zvolin, line 29: wouldn't `Keypair` and `Multiaddr` be supported with some `wasm` feature in libp2p?
If so, then we could use the `[u8; HASH_SIZE]` for hash and remove the wasm_bindgen stuff.

We can still translate to and from some types in `WasmNode::new`

wasm-node/src/node.rs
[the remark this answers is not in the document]
zvolin: @fl0rek could you check if we could have the `-> Result<T, JsValue>` instead of those unwrap throws? they scary me a bit :D

here we could do eg:
```suggestion
            .map(|addr| addr.as_string().and_then(|val| val.parse()))
            .collect::<Result<_, _>()?;
```

wasm-node/src/utils.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[wasm_bindgen]
+pub fn setup_logging() {
+    console_error_panic_hook::set_once();
+
+    tracing_wasm::set_as_global_default();
+}
+
+#[wasm_bindgen]
+pub fn canonical_network_bootnodes(network: Network) -> JsValue {
```
zvolin, line 21: we could not export this function to js and call it from within the `WasmNodeConfig::new`

wasm-node/src/utils.rs
```diff
@@ -0,0 +1,64 @@
...
+        ].iter().map(|s| s.parse().unwrap_throw());
+
+        bootnodes.extend(webtransport_bootnode_addrs);
+    }
+
+    to_value(&bootnodes).unwrap_throw()
+}
+
+#[wasm_bindgen]
+pub fn network_genesis(network: Network) -> JsValue {
```
zvolin, line 42: the same as with `canonical_network_bootnodes`

wasm-node/src/utils.rs
```diff
@@ -0,0 +1,64 @@
...
+#[wasm_bindgen]
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub enum Network {
+    Arabica,
+    Mocha,
+    Private,
+}
+
+#[wasm_bindgen]
+pub fn setup_logging() {
```
zvolin, line 14: Instead of requiring user to call this function, we could annotate it with `#[wasm_bindgen(start)]` to get it executed on wasm initialization

wasm-node/src/wrapper/libp2p.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl From<SwarmConnectionCounters> for ConnectionCounters {
+    fn from(info: SwarmConnectionCounters) -> ConnectionCounters {
+        ConnectionCounters(info)
+    }
+}
+
+#[wasm_bindgen]
+impl ConnectionCounters {
+    #[wasm_bindgen(getter)]
```
zvolin, line 38: do those getters support printing this object as a whole with eg `console.log`?

zvolin (changes requested):
nice, left some additional ideas

wasm-node/pkg/.gitignore
zvolin: wasm-pack generates a .gitignore?

fl0rek: Yes. `pkg` directory itself is needed for `rust-embed` to compile at all so I've added it to tracking so that `clippy` et al doesn't fail in CI.

wasm-node/src/node.rs
fl0rek: Unfortunately, no. Multiaddr comes from `multiaddr` crate and isn't wasm-bindgen aware. `Keypair` comes from `libp2p` but also isn't.

wasm-node/src/wrapper/libp2p.rs
fl0rek: yes 
Screenshot 2023-10-27 at 17 42 22

the `(...)` is a side effect of this being only a getter not primitive value - one has to click on it for chrome to pull the actual value as I did for the first 3

wasm-node/src/utils.rs
fl0rek: Thanks for the suggestion, I like the idea. I've changed the constructor of `WasmNodeConfig` too. It now takes `Network` only and fills out `bootnodes` and `genesis_hash` with official defaults, obviating the need for `canonical_network_bootnodes` and `network_genesis` to be exported.
If user wants to change the defaults, or connect to private network, they'd use appropriate setters to adjust the config.

wasm-node/src/utils.rs
fl0rek: please see response https://github.com/eigerco/celestia-node-rs/pull/112#discussion_r1374786229

celestia/static/run_node.js
fl0rek: right, removed

wasm-node/src/node.rs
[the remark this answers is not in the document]
fl0rek: this part of code got refactored, but to the original question - both `unwrap_throw` and returning `Result` give similar effect - from the js pointview, exception is thrown. Result way requires type conversion, but is slightly more friendly as we can control the error getting passed, for the former it's just "Uncaught Error: `unwrap_throw` failed" with a backtrace. I'll switch to `Result` where possible.

celestia/src/common.rs
oblique: How does this work? Is `_xor` special?

celestia/src/common.rs
oblique: In general I mostly use subcommands. It is much easier to split the actions. Duplication is fine is this case.

wasm-node/src/utils.rs
oblique: The bad thing about this, is that user can not set its own tracing logger.

celestia/src/common.rs
fl0rek: This uses clap grups https://docs.rs/clap/latest/clap/struct.ArgGroup.html. 
By default `multiple` is set to false there, which means that only one argument from the group can be specified and `xor` in the name has not special meaning or use.

celestia/src/common.rs
fl0rek: I thought simple flag would do, but probably it's good idea to switch to subcommands here, changed 👍

wasm-node/src/node.rs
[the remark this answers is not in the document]
fl0rek: Found additional resources about passing errors from Rust to Js:
https://github.com/rustwasm/wasm-bindgen/issues/1742
https://github.com/rustwasm/wasm-bindgen/issues/2970
So I've ended up defining `Result<T,`[`JsError`](https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsError.html)`>`, which plays nicely with existing rust errors. Long term we could export our error type into js for more granular control.

celestia/src/common.rs
fl0rek: Serialize is used by axum as this field is returned as part of the configuration in `WasmNodeArgs` returned as JSON, deleted Deserialize as it wasn't needed.
`repr(u8)` is required by `Serialize_repr. I'm using that because it matches how wasm-bindgen represents rust enums in js, as numbers, so I can use them directly without conversion.

wasm-node/Cargo.toml
fl0rek: It's getting pretty ubiquitous so I think it's a good idea.
I've made the change in b3176aad0df9d866650af6c224893cc97051739b, can split it out if this PR is getting too big.

wasm-node/src/utils.rs
fl0rek: Currently set this to be run unconditionally, created https://github.com/eigerco/celestia-node-rs/issues/121 for future tracking.

wasm-node/rsc/run_node.js
fl0rek: Fixed the overall structure a bit, `window.node` and `window.config` is where the running node and its config are kept respectively and config is kept up to date with UI widgets 👍

.github/workflows/ci.yml
```diff
@@ -68,6 +68,9 @@ jobs:
     - name: Test node crate
       run: wasm-pack test --headless --chrome node
 
+    - name: Build and pack wasm-node
+      run: wasm-pack build wasm-node && wasm-pack pack wasm-node
```
zvolin, line 72:
```suggestion
      run: wasm-pack build --release --target web wasm-node && wasm-pack pack wasm-node
```

celestia/static/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <body>
+        <h1>Celestia Node</h1>
+        <h2>Network</h2>
+        <select id="network_id" class="config">
+            <option value="0">Arabica</option>
+            <option value="1">Mocha</option>
+            <option value="2">Private</option>
+        </select>
+        <h2>Bootnodes</h2>
+        <textarea id="bootnodes" class="config" cols=80 rows=10>
```
zvolin, line 17: may we get this wider? I don't really care about overflows but this way it can fit all bootstrap mas without splitting
```suggestion
        <textarea id="bootnodes" class="config" cols=220 rows=10>
```

celestia/static/run_node.js
```diff
@@ -0,0 +1,100 @@
+Error.stackTraceLimit = 99; // rust stack traces can get pretty big, increase the default
+
+import init, { setup_logging, Network, Node, NodeConfig} from "/wasm/wasm_node.js";
```
zvolin, line 3:
```suggestion
import init, { Node, NodeConfig} from "/wasm/wasm_node.js";
```
unused

celestia/static/run_node.js
```diff
@@ -0,0 +1,100 @@
...
+        window.config.bootnodes = event.target.value.split("\n");
+    });
+
+
+    document.getElementById("network_id").value = config.network;
+    document.getElementById("genesis_hash").value = config.genesis_hash;
+    document.getElementById("bootnodes").value = config.bootnodes.join("\n");
+}
+
+async function start_node(proxy_config) {
```
zvolin, line 77:
```suggestion
async function start_node(config) {
```

celestia/static/run_node.js
```diff
@@ -0,0 +1,100 @@
...
+    window.node = await new Node(config);
+
+    document.getElementById("peer_id").innerText = JSON.stringify(await window.node.local_peer_id());
+}
+
+async function main(document, window, undefined) {
```
zvolin, line 83: doesn't this create a local variable with the name `undefined`?
```suggestion
async function main(document, window) {
```

wasm-node/src/node.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+#[wasm_bindgen(js_class = Node)]
+impl WasmNode {
+    #[wasm_bindgen(constructor)]
+    pub async fn new(config: WasmNodeConfig) -> Self {
+        let network_id = network_id(config.network.into());
+        let store = IndexedDbStore::new(network_id).await.unwrap();
+        info!(
+            "Initialised store with head height: {:?}",
```
zvolin, line 42: this prints

> Initialised store with head height: Err(NotFound)

celestia/static/run_node.js
fl0rek: with guaranteed `undefined` value, since we've passed 2 parameters in the call :D
it's not used though, removing

Cargo.toml
```diff
@@ -1,12 +1,14 @@
...
 resolver = "2"
-members = ["celestia", "node", "proto", "rpc", "types"]
+members = ["celestia", "node", "proto", "rpc", "types", "wasm-node"]
 
 [workspace.dependencies]
 celestia-node = { version = "0.1.0", path = "node" }
 celestia-proto = { version = "0.1.0", path = "proto" }
 celestia-rpc = { version = "0.1.0", path = "rpc", default-features = false }
 celestia-types = { version = "0.1.0", path = "types", default-features = false }
+celestia-wasm-node = { version = "0.1.0", path = "wasm-node" }
```
zvolin, line 10: nitpick, I think I'd prefer `celestia-node-wasm` and `node-wasm` but I have no strong opinion, wdyt? @oblique @fl0rek

celestia/static/run_node.js
```diff
@@ -0,0 +1,100 @@
...
+        window.config.genesis_hash = event.target.value;
+    });
+    document.getElementById("bootnodes").addEventListener("change", (event) => {
+        window.config.bootnodes = event.target.value.split("\n");
+    });
+
+
+    document.getElementById("network_id").value = config.network;
+    document.getElementById("genesis_hash").value = config.genesis_hash;
+    document.getElementById("bootnodes").value = config.bootnodes.join("\n");
```
zvolin, line 74: there is no `config` here, but as mentioned offline, the current approach doesn't play well with changing the network

celestia/static/run_node.js
```diff
@@ -0,0 +1,107 @@
...
+
+
+function bind_config(data) {
+  const network_div = document.getElementById("network_id");
+  const genesis_div = document.getElementById("genesis_hash");
+  const bootnodes_div = document.getElementById("bootnodes");
+
+  const update_config_elements = () => {
+    network_div.value = window.config.network;
+    genesis_div.value = window.config.genesis_div || "";
```
zvolin, line 47: I had typo there

```suggestion
    genesis_div.value = window.config.genesis_hash || "";
```

wasm-node/src/node.rs
fl0rek: Switched to printing head only,if it exists, same as native.

wasm-node/src/node.rs
```diff
@@ -0,0 +1,187 @@
...
+        let header = from_value::<ExtendedHeader>(from)?;
+        let verified_headers = self
+            .node
+            .p2p()
+            .get_verified_headers_range(&header, amount)
+            .await?;
+
+        Ok(Array::from_iter(
+            verified_headers.iter().map(|v| to_value(v).unwrap_throw()),
+        ))
```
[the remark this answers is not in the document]
fl0rek: I've wanted to avoid `collecting` here, because `Array` has constructor `from_iter`. Other ways I've tried to avoid `unwrap_throw`resulted in additional copy of the entire array getting created, as `Vec` cannot be returned directly.

Cargo.toml
fl0rek: I think wasm-node is less popular overall,  so renamed to `node-wasm` and `celestia-node-wasm`.

wasm-node/src/node.rs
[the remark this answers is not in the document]
zvolin: iirc `from_iter` comes from `FromIterator` trait and is what allows you to actually call `.collect()`. I see them equivalent except that `.collect()` is rather preferred unless some less common use cases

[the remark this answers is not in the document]
zvolin: https://doc.rust-lang.org/stable/std/iter/trait.FromIterator.html

node/src/network.rs
```diff
@@ -0,0 +1,116 @@
...
+            .collect()
+        ,
+        Network::Mocha => [
+            // 40.85.94.176 is a node set up for testing QUIC/WebTransport since official nodes
+            // don't have that enabled currently
+            "/ip4/40.85.94.176/tcp/2121/p2p/12D3KooWNJ3Nf1DTQTz8JZogg2eSvPKKKv8itC6fxxspe4C6bizs",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/p2p/12D3KooWNJ3Nf1DTQTz8JZogg2eSvPKKKv8itC6fxxspe4C6bizs",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/webtransport/certhash/uEiBf-OX4HzFK9owOpjdCifsDIWRO0SoD3j3vGKlq0pAXKw/certhash/uEiCx1md1BATJ_0NXAjp3KOuwRYG1535E7kUzFdMq8aPaWw/p2p/12D3KooWNJ3Nf1DTQTz8JZogg2eSvPKKKv8itC6fxxspe4C6bizs",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/p2p/12D3KooWQUYAApYb4DJnhS1QmAwRr5HRvUeHJYocchCpwEhCtDGu",
+            "/ip4/40.85.94.176/udp/2121/quic-v1/webtransport/certhash/uEiBr4-sr95BpqfA-ttpjiLdjbGABhTvX8oxrTXf3Ubfibw/certhash/uEiBSVgyze9xG1UbbNuTwyEUWLPq7l2N9pyeQSs3OtEhGRg/p2p/12D3KooWQUYAApYb4DJnhS1QmAwRr5HRvUeHJYocchCpwEhCtDGu",
```
oblique, line 73: These need to be updated every two weeks. I'm suggesting the following (it can be done in another task):

Implement some kind of lookup function that will return all the addresses of the specified node. The `serve_config` will use the TCP multiaddr to fetch the addresses and serve the WebTransport addresses via `/cfg.json`.

celestia/static/run_node.js
```diff
@@ -0,0 +1,107 @@
...
+        var li = document.createElement("li");
+        li.innerText = peer;
+        peers_ul.appendChild(li);
+    });
+
+    document.getElementById("peers").replaceChildren(peers_ul);
+}
+
+
+
```
oblique, line 39: Reduce this to single empty line

oblique:
Build notes for `celestia` executable must be updated

zvolin:
I'm picking this up as @fl0rek is on a vacations

node-wasm/pkg/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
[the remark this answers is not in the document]
zvolin: Any change to this file would result in dirty workspace after building with `wasm-pack`. I switched from this `.gitignore` to a `.gitkeep` which does the job while not being overwritten by `wasm-pack` at the same time

zvolin:
> Build notes for `celestia` executable must be updated

@oblique what do you think about having a buildscript for that in `celestia` crate? if so, what do you think about when should it be rebuilt. I am thinking about it in regards to convenience and usage in cargo install, I'll still update the docs

oblique:
We can have a `Justfile`

node/src/network.rs
zvolin: > /dns4/celestia1.nodes.eiger.co/udp/2121/quic-v1/webtransport/p2p/12D3KooW9z4jLqwodwNRcSa5qgcSgtJ13kN7CYLcwZQjPRYodqWx

please use this one for now

node-wasm/src/node.rs
```diff
@@ -0,0 +1,186 @@
...
+        let verified_headers = self
+            .node
+            .p2p()
+            .get_verified_headers_range(&header, amount)
+            .await?;
+
+        Ok(Array::from_iter(
+            verified_headers.iter().map(|v| to_value(v).unwrap_throw()),
+        ))
+    }
```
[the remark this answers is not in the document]
zvolin: we've reworked the Node api in #127, this should be aligned to it

node/src/node.rs
```diff
@@ -167,8 +167,8 @@ where
     }
 
     /// Get a synced header for the block with a given height.
-    pub async fn get_header_by_height(&self, hash: u64) -> Result<ExtendedHeader> {
-        Ok(self.store.get_by_height(hash).await?)
+    pub async fn get_header_by_height(&self, height: u64) -> Result<ExtendedHeader> {
```
zvolin, line 170: :upside_down_face:

node-wasm/src/node.rs
```diff
@@ -0,0 +1,254 @@
...
+        let hash: Hash = hash.parse().js_context("parsing hash failed")?;
+        let eh = self.0.get_header_by_hash(&hash).await?;
+        Ok(to_value(&eh)?)
+    }
+
+    /// Get a synced header for the block with a given height.
+    pub async fn get_header_by_height(&self, height: u64) -> Result<JsValue> {
+        let eh = self.0.get_header_by_height(height).await?;
+        Ok(to_value(&eh)?)
+    }
```
zvolin, line 178:
```suggestion
    }
    
```

node-wasm/src/node.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Get synced headers from the given heights range.
+    ///
+    /// If start of the range is undefined (None), the first returned header will be of height 1.
+    /// If end of the range is undefined (None), the last returned header will be the last header in the
+    /// store.
+    ///
+    /// # Errors
+    ///
+    /// If range contains a height of a header that is not found in the store or [`RangeBounds`]
+    /// cannot be converted to a valid range.
```
zvolin, line 188:
```suggestion
    /// If range contains a height of a header that is not found in the store.
```

node-wasm/src/node.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Get a synced header for the block with a given height.
+    pub async fn get_header_by_height(&self, height: u64) -> Result<JsValue> {
+        let eh = self.0.get_header_by_height(height).await?;
+        Ok(to_value(&eh)?)
+    }
+    /// Get synced headers from the given heights range.
+    ///
+    /// If start of the range is undefined (None), the first returned header will be of height 1.
+    /// If end of the range is undefined (None), the last returned header will be the last header in the
+    /// store.
```
zvolin, line 183: I was wondering for a moment if we shouldn't here mimic [`slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice), however I'm not really sure about that tho so feel free to just resolve you or @oblique.

node-wasm/src/node.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /// Get all the multiaddresses on which the node listens.
+    pub async fn listeners(&self) -> Result<Array> {
+        let listeners = self.0.listeners().await?;
+
+        Ok(listeners
+            .iter()
+            .map(js_value_from_display)
+            .collect::<Array>())
```
zvolin, line 96: does it return an empty array or Err? We could as well just return here the or empty array directly as it's not possible to listen in browser afaik

zvolin (approved):
naisu :ninja:

node-wasm/src/node.rs
oblique: I'm fine with the current implementation

celestia/src/common.rs
```diff
@@ -0,0 +1,57 @@
+use anyhow::Result;
+use celestia_node::network::Network;
+use clap::{Parser, ValueEnum};
+use serde_repr::Serialize_repr;
+
+use crate::{native, server};
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, ValueEnum, Serialize_repr)]
+#[repr(u8)]
+pub(crate) enum ArgNetwork {
```
zvolin, line 10: this should support Mainnet and as default

celestia/static/run_node.js
```diff
@@ -0,0 +1,105 @@
...
+  update_config_elements();
+
+  network_div.addEventListener("change", (event) => {
+    window.config.network = Number(event.target.value);
+  });
+  genesis_div.addEventListener("change", (event) => {
+    window.config.genesis_hash = event.target.value;
+  });
+  bootnodes_div.addEventListener("change", (event) => {
+    window.config.bootnodes = event.target.value.split("\n");
```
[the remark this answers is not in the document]
zvolin: this is more forgiving for a user, doesn't produce failures with newlines etc.

node-wasm/src/node.rs
fl0rek: Currently this returns empty array `[]`, as expected. Is there a harm to leaving it as is? guess we could gain listening capabilities some day?

celestia/static/run_node.js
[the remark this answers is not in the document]
zvolin: actually I have even better idea

node-wasm/src/node.rs
zvolin: we can leave it

node-wasm/Cargo.toml
```diff
@@ -2,6 +2,9 @@
 name = "celestia-node-wasm"
 version = "0.1.0"
 edition = "2021"
+description = "celestia node in wasm"
+repository = "https://github.com/eigerco/celestia-node-rs/"
```
oblique, line 6:
```suggestion
repository = "https://github.com/eigerco/celestia-node-rs"
```

node-wasm/Cargo.toml
```diff
@@ -2,6 +2,9 @@
 name = "celestia-node-wasm"
 version = "0.1.0"
 edition = "2021"
+description = "celestia node in wasm"
```
oblique, line 5:
```suggestion
description = "Celestia node in WASM"
```

zvolin (approved):
lgtm

oblique (approved):
lgtm

PR #114, opened 2023-10-25
feat(rpc): create wrappers for jsonrpsee clients

### LINKED ISSUE #100: rpc: re-export jsonrpsee HttpClient and WsClient
`celestia-rpc` provides convenience functions to return a `HttpClient` and `WsClient`, but no way to actually name them other than importing `jsonrpsee` directly.

celestia-rpc should reexport these (and potentially all of `jsonrpsee_http_client`, `jsonrpsee_ws_client` and maybe others, haven't checked) so that downstream users need not directly depend on celestia-rpc's version of jsonrpsee.

zvolin:
An alternative way to fix #100 that avoids re-exporting jsonrpsee directly

### DIFF
--- a/rpc/src/client.rs
+++ b/rpc/src/client.rs
@@ -5,45 +5,37 @@
 //! one using [`jsonrpsee`] crate directly.
 
 #[cfg(not(target_arch = "wasm32"))]
-pub use self::imp::{HttpClient, WsClient};
+pub use self::native::Client;
 
 #[cfg(not(target_arch = "wasm32"))]
-mod imp {
-    use crate::Result;
+mod native {
+    use std::fmt;
+    use std::marker::Send;
+    use std::result::Result as StdResult;
+
+    use crate::{Error, Result};
+    use async_trait::async_trait;
     use http::{header, HeaderValue};
-    use jsonrpsee::http_client::{HeaderMap, HttpClient as JrpcHttpClient, HttpClientBuilder};
-    use jsonrpsee::ws_client::{WsClient as JrpcWsClient, WsClientBuilder};
+    use jsonrpsee::core::client::{BatchResponse, ClientT, Subscription, SubscriptionClientT};
+    use jsonrpsee::core::params::BatchRequestBuilder;
+    use jsonrpsee::core::traits::ToRpcParams;
+    use jsonrpsee::core::Error as JrpcError;
+    use jsonrpsee::http_client::{HeaderMap, HttpClient, HttpClientBuilder};
+    use jsonrpsee::ws_client::{WsClient, WsClientBuilder};
+    use serde::de::DeserializeOwned;
 
-    /// Json RPC client using http protocol.
-    pub struct HttpClient {
-        inner: JrpcHttpClient,
+    /// Json RPC client.
+    pub enum Client {
+        Http(HttpClient),
+        Ws(WsClient),
     }
 
-    impl HttpClient {
-        /// Create a new http client.
-        pub fn new(conn_str: &str, auth_token: Option<&str>) -> Result<Self> {
-            let mut headers = HeaderMap::new();
-
-            if let Some(token) = auth_token {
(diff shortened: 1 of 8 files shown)


oblique:
In this case, IMO we should have only `Client` and internally choose which one to use based on the uri.

rpc/src/client.rs
```diff
@@ -1,45 +1,150 @@
...
+            if let Some(token) = auth_token {
+                let val = HeaderValue::from_str(&format!("Bearer {token}"))?;
+                headers.insert(header::AUTHORIZATION, val);
+            }
 
-        let client = HttpClientBuilder::default()
-            .set_headers(headers)
-            .build(conn_str)?;
+            let client = match conn_str.split_once(':') {
+                Some(("http", _)) => Client::Http(
```
oblique, line 47: add `"https"`

rpc/src/client.rs
```diff
@@ -1,45 +1,150 @@
...
+                    HttpClientBuilder::default()
+                        .set_headers(headers)
+                        .build(conn_str)?,
+                ),
+                Some(("ws", _)) => Client::Ws(
```
oblique, line 52: add `"wss"`

rpc/tests/utils/client.rs
```diff
@@ -54,12 +53,13 @@ pub async fn new_test_client(auth_level: AuthLevel) -> Result<WsClient> {
 }
 
 // This can be used if you want to inspect the requests from `mitmproxy`.
-pub async fn new_test_client_http(auth_level: AuthLevel) -> Result<HttpClient> {
+pub async fn new_test_client_http(auth_level: AuthLevel) -> Result<Client> {
```
oblique, line 56: Actually we can remove this now. `Client` will take the decision based on `CELESTIA_RPC_URL` env var.

rpc/src/client.rs
```diff
@@ -1,45 +1,150 @@
...
 
-        Ok(client)
+            Ok(client)
+        }
     }
 
-    pub async fn new_websocket(conn_str: &str, auth_token: Option<&str>) -> Result<WsClient> {
-        let mut headers = HeaderMap::new();
+    #[async_trait]
+    impl ClientT for Client {
```
oblique, line 66: I thought we wanted to hide anything related to `jsonrpsee`. Am I missing something?

rpc/src/client.rs
zvolin: https and ws are not supported by celestia-node currently

rpc/src/client.rs
zvolin: as with https

rpc/src/client.rs
zvolin: yes, so the #100 was about having to manually include `jsonrpsee` in dependencies for typing the client we returned from `celestia_rpc::client::` helpers. We still want to have all the `(Blob|Share|Header|P2p|etc)Client` in our API and allow for interacting with them through custom `jsonrpsee` clients I think.

If we were to hide the `jsonrpsee` completely then we'd need to write a boilerplate for all the rpc modules in a `Client` and also provide more flexible interface allowing eg. setting max_request / response sizes or timeouts?

rpc/src/client.rs
oblique: Someone can use reverse proxy and have an https in front of it. I think we should support it

rpc/src/client.rs
oblique: Ok got it.

rpc/src/client.rs
zvolin: added

rpc/src/client.rs
zvolin: added

oblique (approved):
lgtm

fl0rek (approved):
nice!

PR #115, opened 2023-10-25
chore: Make sure we run clippy for wasm and fix wasm build/lints

fl0rek:
Added running clippy for wasm32.

Disabled `p2p` feature gated test for rpc, as there's no obvious way to set up test libp2p nodes in wasm (which is why tiny-node is also compiled out).

`rpc/tests/utils/client.rs` - reorganised code in anticipation of implementing jsonrpc using wasm-client. It seems like we need a way to get auth token (for native it we get it from the environment) and send it alongside the request in header.

Fixed missed wasm clippy issues


oblique:
Converted it to draft since it has many `todo!()`

rpc/src/client.rs
```diff
@@ -150,5 +153,80 @@ mod native {
...
+    use jsonrpsee::core::traits::ToRpcParams;
+    use jsonrpsee::core::Error as JrpcError;
+    use serde::de::DeserializeOwned;
+
     // TODO: implement HttpClient with `fetch`
+    pub struct Client;
+
+    impl Client {
+        pub async fn new(_conn_str: &str, _auth_token: Option<&str>) -> crate::Result<Self> {
+            // TODO: jsonrpsee_wasm_client doesn't support setting headers
```
zvolin, line 170: wasm-client and more specifically websockets are not an option for us as due to this reason. This should be done with `fetch` and that's why we didn't implement it yet :) Can we just revert changes on the `toml` and `src` there and just properly guard tests for now?

fl0rek: Ah, I thought we'd be able to achieve single interface with just jsonrpsee - removed the stubs for wasm here and added `#!cfg(..)` to rpc tests so that we can run clippy.

rpc/Cargo.toml
```diff
@@ -16,6 +16,9 @@ tracing = "0.1.37"
 http = "0.2.9"
 jsonrpsee = { version = "0.20", features = ["http-client", "ws-client"] }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+jsonrpsee = { version = "0.20", features = ["wasm-client"] }
```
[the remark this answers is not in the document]
fl0rek: removed this dependency for wasm as it's unused now 👍

.github/workflows/ci.yml
```diff
@@ -17,9 +17,18 @@ jobs:
         version: "23.3"
         repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-    - name: Run clippy
+    - name: Install wasm Rust
+      uses: ATiltedTree/setup-rust@v1
```
zvolin, line 21: hmmm, that's an exotic pick :) I guess that's because the actions-rs got archived? I'd rather think about going with dtolnay's one in that case but it should also be then changed in other steps

zvolin: I'm fine with continuing to use actions-rs too tho until we decide what we migrate too, as there are also new players in the game like:
https://github.com/moonrepo/setup-rust

fl0rek: It being archived scared me off, but it seems like it installs wasm clippy just fine, switched to `actions-rs` 👍

zvolin (approved):
lgtm

oblique (approved):

PR #117, opened 2023-10-26
chore: refactore compose to support multiple nodes

zvolin:
the -v flag is no longer needed when stopping compose setup


zvolin:
fixed ci and gitignored whole credentials directory

ci/run-validator.sh
```diff
@@ -34,35 +46,59 @@ wait_for_block() {
...
+  for node_idx in $(seq 0 "$last_node_idx"); do
+    # TODO: create an issue in celestia-app and link it here
+    # we need to transfer the coins for each node in separate
+    # block, or the signing of all but the first one will fail
+    wait_for_block $((start_block + node_idx))
+
+    local bridge_name="bridge-$node_idx"
+    local bridge_address
+
+    bridge_address=$(node_address "$bridge_name")
```
fl0rek, line 93:
```suggestion
    local bridge_address=$(node_address "$bridge_name")
```
or was there a reason you've put them separately?

tools/gen_auth_tokens.sh
```diff
@@ -5,6 +5,17 @@ DOTENV=".env"
...
 DOCKER_COMPOSE_FILE="./ci/docker-compose.yml"
 
+wait_for_docker_setup() {
+ # we follow the logs with -f and then kill it with awk exit so we
+ # need to suppress the exit status of the docker compose command
+ # but we check the correctness with the grep instead
+ set +o pipefail
+  docker compose -f "$DOCKER_COMPOSE_FILE" logs -f |
+    awk '/Configuration finished. Running a bridge/ {print; exit}' |
+    grep -o Configuration >/dev/null
```
fl0rek, line 15: I'm not quite clear what grep is checking for? As I understand it, awk will run until it matches the `Configuration finished...`, returning successfuly or, if service isn't running or is stopped, exit unsuccessfuly when docker compose exits

ci/docker-compose.yml
```diff
@@ -5,19 +5,52 @@ services:
...
+  bridge-0:
     image: bridge
     platform: "linux/amd64"
     build:
       context: .
       dockerfile: Dockerfile.bridge
-    volumes:
-      - shared:/shared
+    # uncomment to provide the id of the bridge node (default: 0)
+    # command: ["/opt/entrypoint.sh", "0"]
```
fl0rek, line 21: I'm wondering whether providing node id with env variable would be less awkward?
Initially I wanted to suggest extracting it from container name, but there doesn't seem to be any good way to get to that from inside.

ci/run-validator.sh
zvolin: yeah, so if you do a local binding and command substitiution at once, the error from command substitution will be suppressed by the exit code of local binding

tools/gen_auth_tokens.sh
zvolin: I'm not sure how to explain that correctly :D feel free to suggest the improvement to the comment.

So here we have 2 possible scenarios:
 - if service is running, we will wait for the 'Configuration finished...' log entry and then exit with awk
 - if service is not running, it will return immediately not matching in awk
 
 So the awk itself will result in `0` status code either way, so I added the `grep -o` to report correct return code depending on whether we matched the correct log line or not.
 
We are also using `-o pipefail` which propagates error codes through pipes, eg:
 ```bash
 set +o pipefail # the default
 command_that_exits_with_non_zero | command_that_exits_with_zero
 $? # evaluates to 0
 
 set -o pipefail # turn on error propagation
 command_that_exits_with_non_zero | command_that_exits_with_zero
 $? # evaluates to non_zero
 ```
and we need to remove it for this part as when the `docker compose logs -f` is aborted by awk's `exit` it will return non-zero exit code that would be always propagated to the end, regardless of further comands

ci/run-validator.sh
fl0rek: til! thanks

ci/docker-compose.yml
zvolin: let's see, switched to env

tools/gen_auth_tokens.sh
fl0rek: > if service is not running, it will return immediately not matching in awk
I don't think that's true - when I run `docker compose logs` with our setup stopped I just get logs from previous iteration. 
I later get `service "bridge-0" is not running`, but that's coming from docker command in `generate_token`.

zvolin: hmmm not sure how did I test that because I was executing those in terminal, but you're right, I also see logs from previous invocation now

zvolin: refactored this part

fl0rek: Oh and if this part remains, I think you could also do somethink like 
```
! docker compose -f "$DOCKER_COMPOSE_FILE" logs -f |
    awk '/Configuration finished. Running a bridge/ {print; exit 1}'
```
instead of testing with grep, not sure what's more hacky though :P

zvolin: so it remained but I used `|| true` at the end instead. It's a bit more clear about the intentions I think as it enforces positive status code no matter what

zvolin: oops forgot to push

oblique: Just curious. Instead of parsing the logs, can't we use `curl` in a loop to check when service response to us?

oblique: Actually it works:

```bash
  while :; do
      curl http://127.0.0.1:26658 > /dev/null 2>&1 && break
      sleep 1
  done
```

zvolin: switched to curl

oblique (approved):
lgtm

PR #123, opened 2023-10-31
feat: unused deps workflow

### LINKED ISSUE #118: Add CI job to check for unused dependencies

Evalir:
Closes #118 

Adds an unused deps workflow that will trigger at UTC 00:00 each day, or when the workflow is dispatched manually, using cargo udeps.


oblique:
Thanks for your contribution!

@zvolin Do we want this once a day or as a CI job? I think it is safe to do it once a day. What do you think?

zvolin:
I think I'd prefer this on PRs job, as it's where such unused deps can be introduced and the action could be then taken immediately. Once a day would be more suited for things like cargo audit. Nevertheless thanks for the contrib @Evalir and I'm also okay with having this periodic now

Evalir:
happy to change it to run per-pr—it's a quick one

oblique:
@Evalir Yes please :)

Evalir:
done :) @oblique

.github/workflows/unused-deps.yml
```diff
@@ -0,0 +1,34 @@
...
+  # Once per day at 00:00 UTC
+  push:
+    branches: [main]
+  pull_request:
+
+env:
+  RUSTFLAGS: -D warnings
+
+jobs:
+    unused-deps:
```
oblique, line 12: Move the job in `ci.yml`

.github/workflows/unused-deps.yml
```diff
@@ -0,0 +1,34 @@
...
+        runs-on: ubuntu-latest
+        name: unused dependencies
+        steps:
+        - name: Checkout sources
+          uses: actions/checkout@v3
+        - name: Install toolchain
+          uses: dtolnay/rust-toolchain@nightly
+
+        - name: Install cargo-udeps
+          run: cargo install cargo-udeps --locked
```
oblique, line 22: Use `taiki-e/cache-cargo-install-action@v1` for this. Check how we install `wasm-pack` in `ci.yml`

.github/workflows/unused-deps.yml
```diff
@@ -0,0 +1,34 @@
...
+          run: cargo +nightly udeps --all-features --all-targets
+
+        - uses: JasonEtco/create-an-issue@v2
+          if: ${{ failure() }}
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            WORKFLOW_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          with:
+            update_existing: true
+            filename: .github/SANITY_UNUSED_DEPS_ISSUE_TEMPLATE.md
```
oblique, line 34: Since the job will run in the CI of each PR, this is not needed

.github/workflows/unused-deps.yml
```diff
@@ -0,0 +1,34 @@
+on:
+  workflow_dispatch: {}
+  # Once per day at 00:00 UTC
+  push:
+    branches: [main]
+  pull_request:
+
+env:
+  RUSTFLAGS: -D warnings
```
oblique, line 9: Instead of having this, you can use `-- -D warnings` at the end `cargo udeps` line

oblique (commented):
Thanks! I have some feedback

.github/workflows/unused-deps.yml
zvolin: done

.github/workflows/unused-deps.yml
zvolin: done

.github/workflows/unused-deps.yml
zvolin: removed

.github/workflows/unused-deps.yml
zvolin: this doesn't work but moved it to step-specific env

oblique (approved):
lgtm

fl0rek (approved):
nice, thanks for the contribution  🎸

PR #126, opened 2023-11-06
chore: Upgrade libp2p to v0.53.0

oblique:


zvolin:
if you want to do it at once, there is this timeout related complexity we added to header-ex that is resolved by the libp2p changes

zvolin (approved):
looks good

PR #127, opened 2023-11-07
refactor: hide p2p and syncer components from public node api

zvolin:


node/src/node.rs
```diff
@@ -66,14 +76,89 @@ where
             p2p: p2p.clone(),
         })?);
 
-        Ok(Node { p2p, syncer })
+        Ok(Node { p2p, store, syncer })
     }
 
+    /// Grab a reference to the [`P2p`] component of the node.
+    #[doc(hidden)]
```
oblique, line 83: why hidden?

zvolin: do we want to expose them?

oblique: In the Rust library I believe we should.

zvolin: exposed under `Node` with `request_` prefix as discussed offline

node/src/node.rs
```diff
@@ -137,28 +137,73 @@ where
...
+
+    /// Request a header for the block with a given height from the network.
+    pub async fn request_header_by_height(&self, hash: u64) -> Result<ExtendedHeader> {
+        Ok(self.p2p.get_header_by_height(hash).await?)
+    }
+
+    /// Request headers in range (from, from + amount] from the network.
+    ///
+    /// The headers will be verified with the `from` header.
+    pub async fn request_verified_headers_range(
```
oblique, line 158: After our discussion, I feel `_range` shouldn't be in the name. `request_verified_headers` feels nicer

node/src/node.rs
```diff
@@ -137,28 +137,73 @@ where
...
     }
 
-    /// Get a header for the block with a given height.
+    /// Get a synced header for the block with a given height.
     pub async fn get_header_by_height(&self, hash: u64) -> Result<ExtendedHeader> {
         Ok(self.store.get_by_height(hash).await?)
     }
+
+    /// Get synced headers in range (from, from + amount].
+    pub async fn get_verified_headers_range(
```
oblique, line 192: Also here, we can still do `get_header_range(range: Range)`. I don't see it as an issue.

BTW, if you are going to change it, check out [RangeBounds](https://doc.rust-lang.org/stable/core/ops/trait.RangeBounds.html)

zvolin: yeah I know of `RangeBounds`. I didn't touch them as the way std implements them is not complete (unergonomic), but I think I can nail them still with some util trait, will look into tomorrow

node/src/node.rs
zvolin: done

node/src/node.rs
zvolin: switched to `RangeBounds`

node/src/store.rs
```diff
@@ -135,4 +163,68 @@ pub enum StoreError {
...
+    let start = match bounds.start_bound() {
+        Bound::Unbounded => 1,
+        Bound::Included(&x) => x,
+        Bound::Excluded(&u64::MAX) => return Err(StoreError::InvalidHeadersRange),
+        Bound::Excluded(&x) => x + 1,
+    };
+    let end = match bounds.end_bound() {
+        Bound::Unbounded => last_index,
+        Bound::Included(&x) => x,
+        Bound::Excluded(&0) => return Err(StoreError::InvalidHeadersRange),
```
oblique, line 182: I think instead of this (and the one in `start_bound` match). We should do `if start > end`

node/src/node.rs
```diff
@@ -66,14 +77,133 @@ where
...
+    #[doc(hidden)]
     pub fn syncer(&self) -> &Syncer<S> {
         &self.syncer
     }
+
+    /// Grab a reference to the [`Store`] component of the node.
+    #[doc(hidden)]
+    pub fn store(&self) -> &S {
+        &self.store
+    }
```
oblique, line 99: I'm not sure on why we should keep this

node/src/node.rs
```diff
@@ -66,14 +77,133 @@ where
...
+    /// If start of the range is unbounded, the first returned header will be of height 1.
+    /// height.
+    /// If end of the range is unbounded, the last returned header will be the last header in the
+    /// store.
+    ///
+    /// # Errors
+    ///
+    /// If range contains a height of a header that is not found in the store or [`RangeBounds`]
+    /// cannot be converted to a valid range.
+    pub async fn get_verified_headers<R>(&self, range: R) -> Result<Vec<ExtendedHeader>>
```
oblique, line 203: Why do we need `verified` in the name? Is it to avoid confusing the users?

node/src/store.rs
```diff
@@ -33,6 +34,33 @@ pub trait Store: Send + Sync + Debug {
...
+    ///
+    /// If range contains a height of a header that is not found in the store or [`RangeBounds`]
+    /// cannot be converted to a valid range.
+    async fn get_range<R>(&self, range: R) -> Result<Vec<ExtendedHeader>>
+    where
+        R: RangeBounds<u64> + Send,
+    {
+        let head_height = self.head_height().await?;
+        let range = to_headers_range(range, head_height)?;
+        let mut headers = Vec::with_capacity(range.clone().count());
```
oblique, line 54: Is `count` optimized in this case? Or does it actually iterate all the items in the range?

node/tests/node.rs
```diff
@@ -1,113 +1,85 @@
...
-        if head.height().value() >= 3 {
-            break;
+    // check local head
+    let head = node.get_local_head_header().await.unwrap();
+    let expected_head = node.get_header_by_height(100).await.unwrap();
+    assert_eq!(head, expected_head);
+
+    // check getting existing headers
+    for _ in 0..100 {
+        let height = rand::thread_rng().gen_range(1..100);
```
oblique, line 43: Why random?

node/src/store.rs
zvolin: but then we will already hit overflow

oblique: Ok true, didn't thought it before.

Can you add the `if start > end` after them?

node/src/node.rs
zvolin: I was wondering whether to remove it. There are some left end to end tests that make the raw header-ex requests but we can just delete them. Up to you

node/src/node.rs
zvolin: we don't need it, I have it kind of symmetrical to the `request_verified_headers` but now with the signature also different, I can rename it. up to you again

node/src/store.rs
zvolin: no, and it's even worse, it can [overflow on wasm](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#overflow-behavior) :exploding_head: 

I went through the possible options and found out about [`TrustedLen`](https://doc.rust-lang.org/stable/std/ops/struct.RangeInclusive.html#impl-TrustedLen-for-RangeInclusive%3CA%3E) and about [ExactSizeIterator](https://doc.rust-lang.org/stable/std/ops/struct.RangeInclusive.html#impl-ExactSizeIterator-for-RangeInclusive%3Cu16%3E) however both doesn't seem well suited due to the possible quirks with usize overflows.

node/tests/node.rs
zvolin: previously I was doing only 10 iterations there and picking random, now it's just a leftover

node/src/store.rs
zvolin: refactored this part. I haven't included this condition as per offline discussion

node/src/store.rs
zvolin: optimized that by hand

node/tests/node.rs
zvolin: removed

node/src/node.rs
zvolin: removed

node/src/node.rs
zvolin: removed _verified

node/src/store.rs
```diff
@@ -135,4 +174,144 @@ pub enum StoreError {
...
+        // valid start indexes
+        Bound::Included(&x) => x,
+        Bound::Excluded(&x) => x + 1, // can't overflow thanks to last_index check
+    };
+    let end = match bounds.end_bound() {
+        // in case of unbounded, default to the last index
+        Bound::Unbounded => last_index,
+        // range ends after the last index
+        Bound::Included(&x) if x > last_index => return Err(StoreError::NotFound),
+        Bound::Excluded(&x) if x > last_index + 1 => return Err(StoreError::NotFound),
```
oblique, line 199:
```suggestion
        Bound::Excluded(&x) if x >= last_index => return Err(StoreError::NotFound),
```

oblique (approved):
lgtm

node/src/store.rs
zvolin: this is not equivalent :) in fact this re-introduces the bug I had before

zvolin: Excluded(last_index + 1) == last_index, the logic here is correct

oblique: actually this is wrong. i dont know what i was thinking

node/src/store.rs
```diff
@@ -33,6 +34,44 @@ pub trait Store: Send + Sync + Debug {
     /// Returns the header of a specific height.
     async fn get_by_height(&self, height: u64) -> Result<ExtendedHeader>;
 
+    /// Returns the headers from the given heights range.
+    ///
+    /// If start of the range is unbounded, the first returned header will be of height 1.
+    /// height.
```
fl0rek, line 40:
```suggestion
```

node/src/node.rs
```diff
@@ -66,14 +77,115 @@ where
...
+    pub async fn get_header_by_height(&self, hash: u64) -> Result<ExtendedHeader> {
+        Ok(self.store.get_by_height(hash).await?)
     }
 
-    pub fn syncer(&self) -> &Syncer<S> {
-        &self.syncer
+    /// Get synced headers from the given heights range.
+    ///
+    /// If start of the range is unbounded, the first returned header will be of height 1.
+    /// height.
```
fl0rek, line 177:
```suggestion
```

fl0rek (approved):
new API looks nice, minor notes in comments, otherwise lgtm 🚀

PR #128, opened 2023-11-07
fix(node): Allow HeaderSub reinitialization

oblique:
When `Syncer` was recovering from all peers disconnection, it was failing on HeaderSub initilization.

### DIFF
--- a/node/src/p2p.rs
+++ b/node/src/p2p.rs
@@ -74,9 +74,6 @@ pub enum P2pError {
     #[error("HeaderEx: {0}")]
     HeaderEx(#[from] HeaderExError),
 
-    #[error("HeaderSub already initialized")]
-    HeaderSubAlreadyInitialized,
-
     #[error("Bootnode multiaddrs without peer ID: {0:?}")]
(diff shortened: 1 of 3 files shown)


node/src/p2p.rs
```diff
@@ -672,22 +662,8 @@ where
...
-        });
-
-        if updated {
-            trace!("HeaderSub initialized");
-            Ok(())
-        } else {
-            Err(P2pError::HeaderSubAlreadyInitialized)
-        }
+    fn on_init_header_sub(&mut self, head: ExtendedHeader) {
+        self.header_sub_watcher.send_replace(Some(head));
```
zvolin, line 666: can we keep the trace?

zvolin: hmm okay ignore me

zvolin: I don't think it's needed

zvolin (approved):
looks good

node/src/p2p.rs
oblique: i added it

zvolin (approved):

Issue #62, opened 2023-09-08
stretch goals: Introduce Session-based Request Distribution


zvolin:

oblique:
https://github.com/eigerco/celestia-node-rs/pull/67#discussion_r1327926574

oblique:
Since partial responses are allowed:
* [Here](https://github.com/eigerco/celestia-node-rs/blob/9f0bdc8990222fdbd8f2822131e0fa54ed0d3ba1/node/src/exchange.rs#L336-L356) we should return the error *only* if vector is empty.
* [Here](https://github.com/eigerco/celestia-node-rs/blob/8df7715b7c79452610d5e3a85ba1b12bac1828b5/node/src/exchange/client.rs#L248-L254) in case of failure if some headers passed verification, we should return them instead of an error.

Session should retry requesting the remaining range.

oblique:
We need to figure out this: Should requests that failed because of a connection error be retried?

PR #130, opened 2023-11-10
feat(node): Implement sessions

oblique:


zvolin:
Closes #62

node/src/header_ex.rs
```diff
@@ -33,14 +34,21 @@ use crate::utils::{protocol_id, OneshotResultSender};
 
 /// Max request size in bytes
 const REQUEST_SIZE_MAXIMUM: usize = 1024;
+/// Time limit on reading/writing a request
+const REQUEST_TIME_LIMIT: Duration = Duration::from_secs(1);
 /// Max response size in bytes
 const RESPONSE_SIZE_MAXIMUM: usize = 10 * 1024 * 1024;
+/// Time limit on reading/writing a response
```
zvolin, line 41: can we align names for the size, like `REQUEST_SIZE_LIMIT`?

oblique: done

node/src/syncer.rs
```diff
@@ -827,4 +851,33 @@ mod tests {
...
+            let start = (height - first_start) as usize;
+            let end = (height - first_start + amount) as usize;
+            respond_to
+                .send(Ok(remaining_headers[start..end].to_vec()))
+                .map_err(|_| format!("headers [{}, {}]", height, height + amount - 1))
+                .unwrap();
+        }
+
+        // Remove already sent batch
+        remaining_headers.drain(..512);
```
fl0rek, line 881: Is this `512` coming from `MAX_HEADERS_IN_BATCH` in syncer? maybe you could use the const here

node/src/header_ex.rs
```diff
@@ -320,7 +383,14 @@ impl Codec for HeaderCodec {
         }
 
         if msgs.is_empty() {
-            return Err(io::Error::new(io::ErrorKind::Other, "invalid response"));
+            // We have two cases that can reach here:
```
fl0rek, line 386:
```suggestion
            // There are two cases we can reach here:
```
or maybe "We have two cases that we can reach here"

node/src/syncer.rs
```diff
@@ -827,4 +851,33 @@ mod tests {
...
     }
+
+    async fn handle_session_batch(
+        p2p_mock: &mut MockP2pHandle,
+        remaining_headers: &mut Vec<ExtendedHeader>,
+        mut requests: Vec<(u64, u64)>,
+    ) {
+        let first_start = requests[0].0;
+
+        for _ in 0..8 {
```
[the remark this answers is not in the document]
oblique: made it more flexible instead

node/src/syncer.rs
oblique: it the the sum of all amounts. i changed it to be calculated

zvolin (approved):
nice

fl0rek (approved):
🎸

PR #147, opened 2023-11-16
feat: Return bootstrap peers as iterator and filter relevant ones

zvolin:


oblique:
Is this for minimizing the "not supported" logs? Or it fixes something else too?

zvolin:
it's mostly the UX change I'd say, you are presented with many bootstrap peers for each network but they are all not working. That way we can show the true state to the user

celestia/src/server.rs
```diff
@@ -53,6 +53,8 @@ pub(crate) async fn run(args: Params) -> Result<()> {
     let genesis_hash = network_genesis(network);
     let bootnodes = if args.bootnodes.is_empty() {
         canonical_network_bootnodes(network)
+            .filter(|addr| addr.protocol_stack().any(|proto| proto == "webtransport"))
```
oblique, line 56: There is a type-safe way for this check:

```suggestion
            .filter(|addr| addr.iter().any(|proto| proto == Protocol::WebTransport))
```

node-wasm/src/node.rs
```diff
@@ -197,7 +197,7 @@ impl WasmNodeConfig {
             network,
             genesis_hash: network_genesis(network.into()).map(|h| h.to_string()),
             bootnodes: canonical_network_bootnodes(network.into())
-                .iter()
+                .filter(|addr| addr.protocol_stack().any(|proto| proto == "webtransport"))
```
oblique, line 200: same here

celestia/src/server.rs
zvolin: done

oblique (approved):

fl0rek (approved):
🎸

PR #160, opened 2023-12-06
feat: Add in-memory blockstore

### LINKED ISSUE #162: Implement Blockstore trait and in-memory blockstore
subtask of #133 
https://docs.rs/fvm_ipld_blockstore/latest/fvm_ipld_blockstore/ is close to our requirements, but is entangled in fvm workspace (we'd need to update cid/multihash crates everywhere there for example). We also need blockstore to be async in anticipation of adding persistance to it.

fl0rek:
closes #162


blockstore/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+    CidTooLong,
+}
+
+type Result<T> = std::result::Result<T, BlockstoreError>;
+
+#[async_trait]
+pub trait Blockstore {
+    async fn get<const S: usize>(&self, cid: &CidGeneric<S>) -> Result<Vec<u8>>;
+    async fn insert<const S: usize>(&self, cid: &CidGeneric<S>, data: &[u8]) -> Result<()>;
+}
```
oblique, line 21: I prefer if we had the same methods as [fvm_ipld_blockstore::Blockstore](https://docs.rs/fvm_ipld_blockstore/0.2.0/fvm_ipld_blockstore/trait.Blockstore.html).

oblique (approved):
lgtm

blockstore/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+type Result<T> = std::result::Result<T, BlockstoreError>;
+
+/// A IPLD blockstore capable of holding arbitrary data indexed by CID. Implementations can impose
+/// limit on supported CID length, and any operations on longer CIDs will fail with [`CidTooLong`]
+///
+/// [`CidTooLong`]: BlockstoreError::CidTooLong
+
+#[cfg_attr(not(docs_rs), async_trait::async_trait)]
```
fl0rek, line 39: I've taken this from https://github.com/dtolnay/async-trait/issues/213
otherwise generated docs show de-sugared async trait with ton of additional lifetimes.
Unfortunately, to get the same effect locally, one has to generate docs with `RUSTDOCFLAGS="--cfg docs_rs" cargo +nightly doc`, not sure if there's neater way to get the same result.

fl0rek: Per daily, I've tried compiling with beta toolchain and without async-trait crate and yeah, there's a warning that using async fn in public traits is discouraged, because it doesn't generate auto traits. Seems like we could stick with the crate or follow the suggestion of desugaring the `fn` ourselves 
```
42 -     async fn get<const S: usize>(&self, cid: &CidGeneric<S>) -> Result<Option<Vec<u8>>>;
42 +     fn get<const S: usize>(&self, cid: &CidGeneric<S>) -> impl std::future::Future<Output = Result<Option<Vec<u8>>>> + Send;
```
wdyt?

zvolin: I am okay with desugaring until this is corrected in std, it is the approach suggested by async wg

zvolin: > warning that using async fn in public traits is discouraged

wait, so it is discouraged in public traits? then I'm leaning towards `async_trait` :sweat_smile:

fl0rek: yeah, we might want to stick with `async_trait`. This is the full warning:
```
warning: use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
  --> blockstore/src/lib.rs:42:5
   |
42 |     async fn get<const S: usize>(&self, cid: &CidGeneric<S>) -> Result<Option<Vec<u8>>>;
   |     ^^^^^
   |
   = note: you can suppress this lint if you plan to use the trait only in your own code, or do not care about auto traits like `Send` on the `Future`
   = note: `#[warn(async_fn_in_trait)]` on by default
help: you can alternatively desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
   |
42 -     async fn get<const S: usize>(&self, cid: &CidGeneric<S>) -> Result<Option<Vec<u8>>>;
42 +     fn get<const S: usize>(&self, cid: &CidGeneric<S>) -> impl std::future::Future<Output = Result<Option<Vec<u8>>>> + Send;
   |
```
Plus discussion from the async fn stabilisation https://github.com/rust-lang/rust/pull/115822#issuecomment-1731149475

fl0rek: @oblique

oblique: I think I like the workaround with `docsrs` flag

oblique:
I tried this and works:

**blockstore/src/lib.rs**
```rust
pub trait Blockstore {
    // ....

    async fn put_many<const S: usize, B, I>(&self, blocks: I) -> Result<()>
    where
        B: Block<S>,
        I: IntoIterator<Item = B> + Send,
        <I as IntoIterator>::IntoIter: Send,
    {
        for b in blocks {
            let cid = b.cid()?;
            self.put_keyed(&cid, b.data()).await?;
        }
        Ok(())
    }
    
    // ...
}

pub trait Block<const S: usize>: Sync + Send {
    fn cid(&self) -> Result<CidGeneric<S>, CidError>;
    fn data(&self) -> &[u8];
}
```

**types/src/share.rs**
```rust
impl Block<NMT_ID_SIZE> for Share {
    fn cid(&self) -> Result<CidGeneric<NMT_ID_SIZE>, CidError> {
        let hasher = NamespacedSha2Hasher::with_ignore_max_ns(true);
        let digest = hasher.hash_leaf(self.as_ref()).iter().collect::<Vec<_>>();

        // size is correct, so unwrap is safe
        let hash = Multihash::wrap(NMT_CODEC, &digest).unwrap();

        Ok(CidGeneric::new_v1(NMT_MULTIHASH_CODE, hash))
    }

    fn data(&self) -> &[u8] {
        &self.data
    }
}
```

blockstore/src/in_memory_blockstore.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    async fn put_keyed<const SS: usize>(&self, cid: &CidGeneric<SS>, data: &[u8]) -> Result<()> {
+        let hash = cid.hash();
+        let hash =
+            Multihash::wrap(hash.code(), hash.digest()).map_err(|_| BlockstoreError::CidTooLong)?;
+        let cid = CidGeneric::new_v1(cid.codec(), hash);
+
+        self.insert_cid(cid, data)
+    }
+}
```
oblique, line 56: Implement also `Blockstore::has` method and use `DashMap::contains_key`. It will avoid allocation of the `get` result.

fl0rek: added 👍

oblique (approved):
lgtm

zvolin (approved):
:guitar:

PR #169, opened 2023-12-19
feat(node): Add shwap data types

fl0rek:


Wondertan:
Hey @fl0rek. 

We had to make some changes to shwap, which landed in [this](https://github.com/celestiaorg/celestia-node/pull/2675/commits/fc082f4f759b7390b0eca94c136dc2fa902af541) commit. The TLDR is described in the commit-msg.

types/src/extended_data_square.rs
```diff
@@ -0,0 +1,150 @@
...
+        match value {
+            0 => Ok(AxisType::Row),
+            1 => Ok(AxisType::Col),
+            n => Err(Error::InvalidAxis(n.into())),
+        }
+    }
+}
+
+/// Represents ExtendedDataSquare with deserialized Shares
+pub struct ExtendedDataSquare {
```
zvolin, line 31: As discussed offline, please try to modify the existing eds

types/src/namespaced_data.rs
```diff
@@ -23,6 +26,57 @@ pub struct NamespacedDataId {
...
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(try_from = "RawNamespacedData", into = "RawNamespacedData")]
+pub struct NamespacedData {
+    pub namespaced_data_id: NamespacedDataId,
+
+    pub proof: NamespaceProof,
+    pub shares: Vec<Share>,
+}
```
zvolin, line 36: please add docs

types/src/sample.rs
```diff
@@ -2,45 +2,153 @@ use std::mem::size_of;
...
-    /// converted to row/col coordinates internally). Same location can be sampled row or
-    /// column-wise, axis_type is used to distinguish that. Axis root hash is calculated from the
-    /// DataAvailabilityHeader
-    #[allow(dead_code)] // unused for now
+/// Represents Sample, with proof of its inclusion and location on EDS
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(try_from = "RawSample", into = "RawSample")]
+pub struct Sample {
+    pub sample_id: SampleId,
+
```
zvolin, line 36: remove empty line

types/src/sample.rs
```diff
@@ -2,45 +2,153 @@ use std::mem::size_of;
...
+    pub sample_proof_type: AxisType,
+    pub share: Share,
+    pub proof: NamespaceProof,
+}
+
+impl Sample {
+    /// Create a new Sample. Index references sample number from the entire Data Square
     pub fn new(
         axis_type: AxisType,
         index: usize,
```
zvolin, line 46: I'm unsure about this `index` + `axis` api. Is there a reason for it to be more convenient than providing the `x` and `y`?

zvolin: if you decide to change that, please update also in SampleId

types/src/sample.rs
```diff
@@ -2,45 +2,153 @@ use std::mem::size_of;
...
             AxisType::Col => (index % square_len, index / square_len),
         };
 
+        let shares = eds.axis(axis_type, axis_index)?;
+
+        let mut tree = Nmt::with_hasher(NamespacedSha2Hasher::with_ignore_max_ns(true));
+
+        // XXX: are erasure coded shares correctly prefixed with parity namespace?
+        for s in &shares {
+            tree.push_leaf(s.data(), *s.namespace())
```
zvolin, line 63: the other half of shares needs to be inserted with `*Namespace::PARITY_SHARE`.
Please correct it together with a test that the proof can be verified with the root from DAH

zvolin: just saw it answers the comment above :sweat_smile: No. Parity shares are generated, so the namespace they have is just a random bytes from our point of view

types/src/sample.rs
```diff
@@ -2,45 +2,153 @@ use std::mem::size_of;
...
+        Ok(Sample {
+            sample_id,
+            sample_proof_type: axis_type,
+            share: shares[sample_index].clone(),
+            proof: proof.into(),
+        })
+    }
+
+    /// Validate sample with root hash from ExtendedHeader
+    pub fn validate(&self, hash: Hash) -> Result<()> {
```
zvolin, line 83: as this has all the required information to navigate the correct hash in dah, I think requiring a dah here would be more user-friendly

types/src/sample.rs
```diff
@@ -2,45 +2,153 @@ use std::mem::size_of;
...
+    fn from(sample: Sample) -> RawSample {
+        let mut sample_id_bytes = BytesMut::new();
+        sample.sample_id.encode(&mut sample_id_bytes);
+        let sample_proof = RawProof {
+            start: sample.proof.start_idx() as i64,
+            end: sample.proof.end_idx() as i64,
+            nodes: sample.proof.siblings().iter().map(|h| h.to_vec()).collect(),
+            leaf_hash: vec![], // this is an inclusion proof
+            is_max_namespace_ignored: true,
+        };
```
zvolin, line 126: don't we have this implemented somewhere already?

types/src/error.rs
```diff
@@ -73,6 +76,12 @@ pub enum Error {
     #[error("Range proof verification failed: {0:?}")]
     RangeProofError(nmt_rs::simple_merkle::error::RangeProofError),
 
+    #[error("Index {0} is out of range for NMT Range Proof")]
+    RangeProofIndexOutOfRange(i64),
```
zvolin, line 80: I couldn't find a usage

types/src/extended_data_square.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /// Get EDS square length
+    pub fn square_len(&self) -> usize {
+        self.square_len
+    }
+
+    /// Return all the shares that belong to the provided namespace in the EDS.
+    /// Results are returned as a list of rows of shares with the inclusion proof
+    pub fn get_namespaced_data(
```
zvolin, line 86: could we add a test for that?

types/src/extended_data_square.rs
fl0rek: Removed extra EDS and moved changes to existsing one 👍

types/src/row.rs
```diff
@@ -0,0 +1,307 @@
...
+        unimplemented!("unable to compute parity shares")
+    }
+}
+
+impl Protobuf<RawRow> for Row {}
+
+impl TryFrom<RawRow> for Row {
+    type Error = Error;
+
+    fn try_from(axis: RawRow) -> Result<Row, Self::Error> {
```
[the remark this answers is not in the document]
fl0rek: should be done now 👍

types/src/sample.rs
fl0rek: Switched to using entire DAH here and in Row 👍

types/src/sample.rs
fl0rek: Indeed we have it in the `namespace_proof`, replacing with `into()` - thanks!

types/src/sample.rs
fl0rek: Got it thanks for clearing that up!

types/src/sample.rs
fl0rek: `axis` isn't used to specify the location of the sample, but only to indicate whether proof is calculated by row or column. Notice how `SampleId` doesn't  use `AxisType` at all, but always works in row-major order.
As for the `index` vs pair `x` + `y` itself, I don't have strong opinions here as I've not yet seen our usage of it, so I copied the API from celestia :P

types/src/extended_data_square.rs
fl0rek: I've added a test (now in `rsmt2d.rs`). I don't think we have a code generating entire DAH from EDS, so I've mocked the DAH manually as this is getting scope creepy already.

types/src/sample.rs
zvolin: hmmm I think they use it only in the DASer and sampling index for a one dimensional array is simpler as you sample only one number

types/src/extended_data_square.rs
zvolin: ooh I was thinking about fetching extended header and then calling get eds for it and save it in test data, and here just use

zvolin: so you can eg. modify the celestia-rpc get_eds test and submit sample a bit bigger blob, like 1000b, then submit it and fetch the data for that block

types/src/sample.rs
zvolin: I'm concerned because I guess celestia guys have it as something pretty internal somewhere near shwap p2p protocol, whereas we expose it from a pub api of some crate. I'd either:
- switch to x, y always row-major in the `new` and rename this one as let's say `with_axis`
- move the shwap types entirely into lumina-node and hide them

zvolin: @oblique wdyt?

types/src/extended_data_square.rs
zvolin: or you can submit many blobs with different namespaces

types/src/sample.rs
zvolin: since oblique has different topics currently, I think it's fine to go with the current one and leave a todo whether we would want an x, y api in a future

types/src/share.rs
```diff
@@ -45,14 +48,54 @@ pub struct Share {
 }
 
 impl Share {
+    /// Create a new Share with given namespace, and data, appropriately setting share sequence
+    /// metadata.
+    /// If sequence length is provided, first share is created, with sequence length field included
+    /// and sequence start indicator set. Otherwise continuation share is created.
+    /// Provided data must fit into the share type and is zero padded, if it's shorter.
+    pub fn new_v0(namespace: Namespace, sequence_length: Option<u32>, data: &[u8]) -> Result<Self> {
```
zvolin, line 56: I'm not so sure about the new api in `Share`, why is it needed for those changes?

zvolin: there are sparse shares and compact shares, and they should be differentiated by the namespace. Also constructing the shares by hand doesn't seem like a usecase for celestia users. You can check the ShareBuilders in celestia-app, but in general I'd stick to the Blob::split_to_shares api for now

types/src/share/info_byte.rs
```diff
@@ -30,4 +31,17 @@ impl InfoByte {
...
+        if version > appconsts::MAX_SHARE_VERSION {
+            Err(Error::MaxShareVersionExceeded(version))
+        } else {
+            Ok(InfoByte(byte))
+        }
+    }
+
+    pub fn from_raw_unchecked(byte: u8) -> Self {
+        InfoByte(byte)
+    }
```
zvolin, line 46: can we keep those pub(crate)?

types/src/share.rs
```diff
@@ -18,6 +19,8 @@ mod info_byte;
 
 pub use info_byte::InfoByte;
 
+const FIRST_SPARSE_SHARE_SEQUENCE_LENGTH_OFFSET: usize = NS_SIZE + appconsts::SHARE_INFO_BYTES;
```
zvolin, line 22:
```suggestion
const SHARE_SEQUENCE_LENGTH_OFFSET: usize = NS_SIZE + appconsts::SHARE_INFO_BYTES;
```
it doesn't depend on being sparse or compact

types/src/share.rs
fl0rek: These were functions I wanted to use mostly for testing, but I thought others might find useful. Seems like for both cases `Blob` would probably suffice, so removing this as suggested

types/src/share/info_byte.rs
fl0rek: yes, updated 👍

types/src/extended_data_square.rs
fl0rek: Good idea, I've captured EDS and DAH using tests used that for testing instead 👍

types/src/row.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+impl Row {
+    /// Create Row with given index from the EDS
+    pub fn new(index: usize, eds: &ExtendedDataSquare, block_height: u64) -> Result<Self> {
+        let square_len = eds.square_len();
+
+        let row_id = RowId::new(index, block_height)?;
+        let mut shares = eds.row(index)?;
+        shares.truncate(square_len / 2);
```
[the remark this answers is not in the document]
fl0rek: I think I agree, moved parity share recomputation into the row creation

types/src/namespaced_data.rs
```diff
@@ -32,18 +32,109 @@ pub const NAMESPACED_DATA_ID_CODEC: u64 = 0x7820;
...
+    /// ```no_run
+    /// use celestia_types::nmt::Namespace;
+    /// # use celestia_types::{ExtendedDataSquare, ExtendedHeader};
+    /// # fn get_extended_data_square(height: usize) -> ExtendedDataSquare {
+    /// #    unimplemented!()
+    /// # }
+    /// # fn get_extended_header(height: usize) -> ExtendedHeader {
+    /// #    unimplemented!()
+    /// # }
+    ///
```
zvolin, line 73:
```suggestion
    /// #
```

types/src/namespaced_data.rs
```diff
@@ -32,18 +32,109 @@ pub const NAMESPACED_DATA_ID_CODEC: u64 = 0x7820;
...
     ///
     /// [`Share`]: crate::Share
     pub namespace: Namespace,
-    /// Index of the row in the [`ExtendedDataSquare`].
+}
+
+/// `NamespacedData` is constructed out of the ExtendedDataSquare, containing up to a row of
+/// shares belonging to a particular namespace and a proof of their inclusion. If, for
+/// particular EDS, shares from the namespace span multiple rows, one needs multiple
+/// NamespacedData instances to cover the whole range.
```
zvolin, line 48: For docs, please go with first short title what is what and then add an explanation in a body below. [rustdoc guide](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components)

```suggestion
/// `NamespacedData` contains up to a row of shares belonging to a particular namespace and a proof of their inclusion.
///
/// It is constructed out of the ExtendedDataSquare. If, for particular EDS, shares from the namespace span multiple rows,
/// one needs multiple NamespacedData instances to cover the whole range.
```

zvolin: I don't know if thats well formatted :facepalm:

types/src/row.rs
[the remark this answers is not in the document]
zvolin: then I think this call to truncate should also disappear :thinking:

types/src/share.rs
```diff
@@ -58,6 +60,8 @@ pub struct Share {
     pub data: [u8; appconsts::SHARE_SIZE],
 }
 
+impl Share {}
+
```
zvolin, line 64: looks like a leftover

zvolin (commented):
looks good, last findings I think

types/src/row.rs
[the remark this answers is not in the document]
fl0rek: true 😅 
fixed, lemmie write few unttests more in the mean time

zvolin (approved):
LGTM

PR #161, opened 2023-12-07
doc: document public api

zvolin:

PR #170, opened 2024-01-05
chore: add missing metadata to the toml files

zvolin:
This requires #161 for the readmes. See the last commit for diff as for now


cli/Cargo.toml
```diff
@@ -3,6 +3,19 @@ name = "lumina-cli"
 version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
+description = "Celestia data availability node implementation in Rust"
+repository = "https://github.com/eigerco/lumina"
+# readme = "README.md" -- It's using the project's root readme as for now, should it have a separate one?
```
zvolin, line 8: how should we handle this?

fl0rek: I guess root `README` could link to a separate `README` in `cli`, but I don't think it's worth it right now. So I'd vote for leaving it as is.

fl0rek (approved):
closer and closer to the release ✨

cli/Cargo.toml
zvolin: there is also a symlink way

blockstore/Cargo.toml
```diff
@@ -3,6 +3,13 @@ name = "blockstore"
 version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
+description = "An IPLD blockstore capable of holding arbitrary data indexed by CID"
+repository = "https://github.com/eigerco/lumina"
+readme = "README.md"
+# crates.io is limited to 5 keywords and 5 categories
+keywords = ["blockstore", "ipld", "cid"]
```
oblique, line 10: add `bitswap` too

oblique (approved):

PR #171, opened 2024-01-05
chore: Enable compatibility with sha2 v0.10.6

preston-evans98:
In the Sovereign SDK, we use sha2 `v0.10.6`. This change makes Lumina compatible with that version while still allowing consumers to rely on the newer version if they desire (see Cargo's [Semver policy](https://doc.rust-lang.org/cargo/reference/semver.html)).

### DIFF
--- a/types/Cargo.toml
+++ b/types/Cargo.toml
@@ -25,7 +25,7 @@ rand = { version = "0.8.5", optional = true }
 ruint = { version = "1.8.0", features = ["serde"] }
 serde = { version = "1.0.164", features = ["derive"] }
 serde_repr = { version = "0.1", optional = true }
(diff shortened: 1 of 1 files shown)


zvolin (approved):
looks good! thanks

fl0rek (approved):

oblique:
@preston-evans98 The crates that you are restricted to a specific versions are only [these](https://github.com/Sovereign-Labs/sovereign-sdk/blob/2cc0656df3f12fca2026c20554b5f78ccb210b89/examples/demo-rollup/provers/risc0/guest-celestia/Cargo.toml#L21-L24)? Is it because of optimizations that was done by risc0 project?

Issue #185, opened 2024-01-16
Publish the lumina-cli crate to crates.io


zvolin:
Currently the lumina-cli crate cannot be published because for browser node it requires artifacts from lumina-node-wasm to be built first

fl0rek:
it's available: https://crates.io/crates/lumina-cli
closing

PR #190, opened 2024-01-19
fix(ci): Fix release for lumina (cli)

fl0rek:
unblocks #185 

- rename leftover `celestia_node_wasm.js` to lumina

~~- release-plz copies only the crate it's releasing into a new directory in `target`, which breaks our inter-crate references. Switching README in `cli` to symlink causes release-plz to copy the linked file so that it can be used during building.~~
- split main README into main `README.md`, which is now more about using lumina, the crate, and  `cli/README.md` which is inteded more for someone interested in just running the node, either locally or in a browser, not necessarily development.

- Related to above release-plz publish behaviour, embedding node-wasm into the lumina-cli did not work for releases. Since we're planning to move to npm release soon(tm), removing embedded wasm lumina node by default.

- Additionally, since now it's again possible to have node-only lumina-cli, made it so that having executable named `lumina-node` causes lumina to act as if `lumina node` was called as a shorthand.


zvolin:
as far as I understand the behaviors you mentioned are `cargo publish` specific, not really release-plz specific. Release-plz should just resolve which crates to publish and in which order but then invoke the `cargo publish` to do the work

cli/Cargo.toml
```diff
@@ -36,10 +34,13 @@ axum = "0.6.20"
 clap = { version = "4.4.4", features = ["derive"] }
 dotenvy = "0.15.7"
 mime_guess = "2.0"
-rust-embed = "8.0.0"
+rust-embed = { version = "8.0.0", features = ["interpolate-folder-path"]}
```
zvolin, line 37:
```suggestion
rust-embed = { version = "8.0.0", features = ["interpolate-folder-path"] }
```

cli/src/common.rs
```diff
@@ -20,17 +24,30 @@ pub(crate) enum CliArgs {
...
 /// Run the Lumina node.
 pub async fn run() -> Result<()> {
     let _ = dotenvy::dotenv();
-    let args = CliArgs::parse();
+
+    let args = if current_exe()?
+        .file_name()
+        .unwrap_or_default()
+        .to_string_lossy()
+        == "lumina-node"
```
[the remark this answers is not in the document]
fl0rek: Many existing binaries do that. I wanted a way to avoid the awkward default of having to spell `lumina node` every time, despite the fact that `node` is the only choice.

README.md
```diff
@@ -11,65 +11,32 @@ Supported features:
...
-# install rust
-curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
-
-# open a new terminal or run
-source "$HOME/.cargo/env"
-
-# clone the repository
-git clone https://github.com/eigerco/lumina
-cd lumina
-```
```
zvolin, line 31: I would leave the build from sources details in the main repo's readme but moved to a new section below, like build from sources. Here and in `cli` I would just keep the `cargo install lumina-cli --locked` step.

cli/README.md
```diff
@@ -0,0 +1,52 @@
...
+# clone the repository
+git clone https://github.com/eigerco/lumina
+cd lumina
+```
+
+### Running the node natively
+
+```bash
+# install lumina
+cargo install --path cli
```
zvolin, line 32: This is what a user will see on crates.io/crates/lumina-cli, I think we should keep there only the instructions for a installation from `crates.io` directly

zvolin, line 32:
```suggestion
## Building and running

Install common dependencies

```bash
# install dependencies
sudo apt-get install -y build-essential curl git protobuf-compiler

# install rust
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

# open a new terminal or run
source "$HOME/.cargo/env"

# clone the repository
git clone https://github.com/eigerco/lumina
cd lumina
```

### Running the node natively

```bash
# install lumina
cargo install lumina-cli --locked
```

zvolin, line 32:
```suggestion
cargo install lumina-cli --locked
```

cli/README.md
```diff
@@ -0,0 +1,52 @@
...
+```bash
+# build wasm-node to be bundled with lumina
+wasm-pack build --target web node-wasm
+
+# install lumina with wasm node embedded
+cargo install --path cli --features embedded-lumina
+
+# serve lumina node on default localhost:9876
+lumina browser
+```
```
zvolin, line 52: as there is no way to install the browser node from crates.io, I would keep those instructions only in main readme

zvolin: + a mention that browser node is currently only supported with sources build

cli/src/common.rs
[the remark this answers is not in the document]
oblique: With this, the user needs to symlink `lumina` to `lumina-node` correct? In that case, we are talking about the user installs `lumina` from a package manager other than `cargo`, right?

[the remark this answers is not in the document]
fl0rek: symlink or rename. For the symlink I think this would work with cargo too, rename probably more applicable to manual build. 
Not particularly adamant about this change, thought I'd be a small convenience thing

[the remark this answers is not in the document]
oblique: ok keep it

README.md
fl0rek: I see, the distinction makes sense, moved all the development/source related notes to the main readme, leaving small running notes in `lumina-cli` 👍

cli/Cargo.toml
```diff
@@ -36,10 +34,18 @@ axum = "0.6.20"
...
 tracing-appender = "0.2.2"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
+
+[build-dependencies]
+envy = "0.4"
+anyhow = "1.0"
+serde = { version = "1.0", features = ["derive"] }
+
+[features]
+embedded-lumina = []
```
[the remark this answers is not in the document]
fl0rek: I've switched to `embedded-wasm-node` wdyt?

zvolin (approved):

cli/Cargo.toml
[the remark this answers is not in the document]
oblique: I like `browser-node`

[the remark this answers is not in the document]
fl0rek: I like it, renamed 🚀

oblique (approved):
lgtm

zvolin (approved):

PR #199, opened 2024-01-21
`types` crate `no_std` support

frisitano:
This PR proposes the introduction of `no_std` support for the `types` crate. This will allow it to be used in `no_std` targets. To achieve this change we have had to remove the usage of `thiserror` in the `types` crate as well as introduce the usage of alloc for `no_std` environment. I wanted to gauge support for this change. If we have support then I will clean it up and make it ready for merge.

### DIFF
--- a/node/src/syncer.rs
+++ b/node/src/syncer.rs
@@ -46,8 +46,8 @@ pub enum SyncerError {
     Store(#[from] StoreError),
 
     /// An error propagated from the [`celestia_types`].
-    #[error(transparent)]
-    Celestia(#[from] celestia_types::Error),
+    #[error(" Celestia error: {0}")]
+    Celestia(celestia_types::Error),
 
     /// The worker has died.
     #[error("Worker died")]
--- a/types/Cargo.toml
+++ b/types/Cargo.toml
@@ -35,7 +35,6 @@ ruint = { version = "1.8.0", features = ["serde"] }
 serde = { version = "1.0.164", features = ["derive"] }
 serde_repr = { version = "0.1", optional = true }
 sha2 = "0.10.6"
-thiserror = { version = "1.0.40", default-features = false }
 
 [dev-dependencies]
 ed25519-consensus = "2.1.0"
--- a/types/Cargo.toml
+++ b/types/Cargo.toml
@@ -35,7 +35,7 @@ ruint = { version = "1.8.0", features = ["serde"] }
 serde = { version = "1.0.164", features = ["derive"] }
 serde_repr = { version = "0.1", optional = true }
 sha2 = "0.10.6"
-thiserror = "1.0.40"
+thiserror = { version = "1.0.40", default-features = false }
 
 [dev-dependencies]
 ed25519-consensus = "2.1.0"
@@ -51,9 +51,10 @@ getrandom = { version = "0.2.10", features = ["js"] }
 wasm-bindgen-test = "0.3"
 
 [features]
-default = ["p2p"]
+default = ["p2p", "std"]
 p2p = ["dep:libp2p-identity", "dep:multiaddr", "dep:serde_repr"]
 test-utils = ["dep:ed25519-consensus", "dep:rand"]
+std = []
 wasm-bindgen = ["celestia-tendermint/wasm-bindgen"]
 
 [package.metadata.docs.rs]
--- a/types/src/blob.rs
+++ b/types/src/blob.rs
@@ -10,6 +10,7 @@ pub use self::commitment::Commitment;
 use crate::consts::appconsts;
 use crate::nmt::Namespace;
 use crate::serializers::none_as_negative_one;
+use crate::types::Vec;
 use crate::{bail_validation, Error, Result, Share};
 
 /// Options for configuring the blob submission to the network.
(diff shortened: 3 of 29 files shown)


zvolin:
Hey, may I ask what is a use case for the no_std support?

frisitano:
> Hey, may I ask what is a use case for the no_std support?

The use case is using the types in a (substrate)[https://github.com/paritytech/substrate] runtime.

types/src/error.rs
```diff
@@ -1,212 +1,279 @@
...
 ///
 /// [`celestia_types::Error`]: crate::Error
-pub type Result<T, E = Error> = std::result::Result<T, E>;
+pub type Result<T, E = Error> = core::result::Result<T, E>;
 
 /// Representation of all the errors that can occur when interacting with [`celestia_types`].
 ///
 /// [`celestia_types`]: crate
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug)]
```
oblique, line 13: For all the error types that you removed `thiserror::Error` implement manually the `Error` trait if `std` feature is set. Otherwise our users will not be able to use our `Error` types with `anyhow` and `thiserror`.

```rust
#[cfg(feature = "std")]
impl std::error::Error for Error {}
```

node/src/syncer.rs
```diff
@@ -46,8 +46,8 @@ pub enum SyncerError {
     Store(#[from] StoreError),
 
     /// An error propagated from the [`celestia_types`].
-    #[error(transparent)]
-    Celestia(#[from] celestia_types::Error),
+    #[error(" Celestia error: {0}")]
+    Celestia(celestia_types::Error),
```
oblique, line 50: This is not needed after you implement `std::error::Error` as explained in the other comment.

types/src/blob.rs
```diff
@@ -10,6 +10,7 @@ pub use self::commitment::Commitment;
 use crate::consts::appconsts;
 use crate::nmt::Namespace;
 use crate::serializers::none_as_negative_one;
+use crate::types::Vec;
```
oblique, line 13: why not `alloc::vec::Vec`?

types/src/block.rs
```diff
@@ -139,11 +140,13 @@ impl CommitExt for Commit {
             block_id: Some(self.block_id),
             timestamp: Some(timestamp),
             validator_address,
-            validator_index: signature_idx.try_into()?,
+            validator_index: signature_idx.try_into().map_err(Error::Tendermint)?,
```
oblique, line 143: `map_err` is not needed because you implemented the `From`. Remove all `map_err` that you added.

oblique (changes requested):

types/Cargo.toml
```diff
@@ -51,9 +50,10 @@ getrandom = { version = "0.2.10", features = ["js"] }
 wasm-bindgen-test = "0.3"
 
 [features]
-default = ["p2p"]
+default = ["p2p", "std"]
 p2p = ["dep:libp2p-identity", "dep:multiaddr", "dep:serde_repr"]
 test-utils = ["dep:ed25519-consensus", "dep:rand"]
+std = []
```
oblique, line 56: We need to disable `std` feature in dependencies too and enable them in our `std` feature. This list is incomplete list:

```
std = ["celestia-tendermint/std", "cid/std", "serde/std"]
```

oblique:
`no_std` it is done on compiler level and it doesn't enforce no-std on dependencies. It basically says "do not import `std` for this crate". To complete go no-std you need much more changes. You can use [cargo-no-std-check](https://github.com/mystor/cargo-no-std-check) to complete compile without `std`.

I will try and summarize what extra changes you need to do:

1. In `celestia-types`, the `blockstore` crate needs std, however it is not going to be needed in no-std context, so make it optional`.
2. You need to make dependencies no-std too. For example for `celestia-proto` crate, the following is needed.
```diff
diff --git a/proto/Cargo.toml b/proto/Cargo.toml
index a6365c5..830b579 100644
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -15,9 +15,9 @@ categories = ["encoding", "cryptography::cryptocurrencies"]

 [dependencies]
 celestia-tendermint-proto = { workspace = true }
-prost = "0.12.0"
-prost-types = "0.12.0"
-serde = { version = "1.0.188", features = ["derive"] }
+prost = { version = "0.12.0", default-features = false }
+prost-types = { version = "0.12.0", default-features = false }
+serde = { version = "1.0.188", default-features = false, features = ["derive"] }

 [build-dependencies]
 anyhow = "1.0.75"
@@ -28,3 +28,7 @@ wasm-bindgen-test = "0.3"

 [dev-dependencies]
 serde_json = "1.0.105"
+
+[features]
+default = ["std"]
+std = ["prost/std", "prost-types/std", "serde/std"]
```
3. You may need more stuff. The target is to make `celestia-type` to compile with `cargo +nightly no-std-check --no-default-features`. If you stuck somewhere that requires design decision, please let us know.

frisitano:
Yes, this change turned out to be a bit more involved than initially anticipated.  In the first instance we will extract the required logic (computation of commitments) to our codebase and look to introduce no_std support for lumina in the coming weeks. I propose that we close this PR and create an issue to capture this feature.  Thank you for taking the time to detail required changes to support this.

PR #202, opened 2024-01-24
feat(node): Integrate bitswap protocol for shwap

oblique:


node/src/p2p.rs
```diff
@@ -393,6 +468,7 @@ where
     S: Store + 'static,
 {
     autonat: autonat::Behaviour,
+    bitswap: BitswapBehaviour<64, InMemoryBlockstore<64>>,
```
fl0rek, line 471: Could you define const for 64 and swap there everywhere?

node/src/p2p.rs
```diff
@@ -106,6 +123,18 @@ pub enum P2pError {
...
     #[error("Bootnode multiaddrs without peer ID: {0:?}")]
     BootnodeAddrsWithoutPeerId(Vec<Multiaddr>),
+
+    /// An error propagated from [`beetswap::Behaviour`].
+    #[error("Bitswap: {0}")]
+    Bitswap(#[from] beetswap::Error),
+
+    /// ProtoBuf message failed to be decoded.
+    #[error("ProtoBuf decoding error: {0}")]
+    ProtoDecodeFailed(celestia_tendermint_proto::Error),
```
zvolin, line 133: any reason to not use `#[from]` there?

zvolin: and `Ok(foo::decode(...)?)` instead of map_err

zvolin (approved):
looks good, left some nit

node/src/p2p.rs
oblique: no reason. added it

fl0rek (approved):
👌

PR #209, opened 2024-01-31
feat(node): Extend header Store for use with DAS-er

### LINKED ISSUE #193: Extend header store with current_sampled_head and add rejected/accepted metadata to headers
- [ ] add the last sampled height to the header store
- [ ] add the accepted/rejected metadata to headers in store, Option<bool>?

fl0rek:
closes #193


node/src/store/in_memory_store.rs
```diff
@@ -118,6 +133,57 @@ impl InMemoryStore {
...
+        if self.sampling_data.insert(height, metadata).is_some() {
+            info!("Overriding existing sampling metadata for height {height}");
+            // modified header wasn't new, no need to update the height
+            Ok(self.get_next_unsampled_height())
+        } else {
+            self.update_cached_sampling_height()
+        }
+    }
+
+    fn update_cached_sampling_height(&self) -> Result<u64> {
```
zvolin, line 156:
```suggestion
    fn update_lowest_unsampled_height(&self) -> Result<u64> {
```
since it's in memory then everything is 'cached', I think this name reads better for me

node/src/store/indexed_db_store.rs
```diff
@@ -5,18 +5,33 @@ use async_trait::async_trait;
...
+// Additional indexes set on HEADER_STORE, for querying by height and hash
 const HASH_INDEX_NAME: &str = "hash";
 const HEIGHT_INDEX_NAME: &str = "height";
 
+#[derive(Serialize_repr, Deserialize_repr, Debug, PartialEq)]
+#[repr(u8)]
+enum HeaderAccepted {
+    Rejected = 1,
+    Accepted = 2,
+}
```
zvolin, line 33: I can't find usage of this

fl0rek: I ended up not using it, removed 👍

node/src/store/in_memory_store.rs
fl0rek: Agree, changed 👌

node/src/store/indexed_db_store.rs
```diff
@@ -280,6 +370,38 @@ async fn get_head_from_database(db: &Rexie) -> Result<ExtendedHeader> {
...
+
+    let pending_headers_keyrange = KeyRange::lower_bound(&to_value(&last_known_unsampled)?, false)?;
+    let pending_sampled_headers = store
+        .get_all(
+            Some(&pending_headers_keyrange),
+            None,
+            None,
+            Some(Direction::Next),
+        )
+        .await?;
```
zvolin, line 390: this is a bit dangerous. If I understand correctly, this will load all headers from the indexeddb after some index. Can this result in loading all the headers after we release the DASer? Since someone can be in sync with mainnet but have nothing dased yet.

I think we should have another loop there, and paginate through headers. Let's add some limit on this like 1000 headers at the time in each iteration

zvolin: or maybe even by 10 (50, 100?) headers at the time? Since from what I see this is gonna be called each time a header is marked sampled, and it's likely we only need to check very few next headers in a regular case

node/src/store/indexed_db_store.rs
```diff
@@ -209,6 +221,64 @@ impl IndexedDbStore {
...
+        let store = tx.store(SAMPLING_STORE_NAME)?;
+
+        let height_key = to_value(&height)?;
+        let sampling_entry = store.get(&height_key).await?;
+
+        if sampling_entry.is_falsy() {
+            return Ok(None);
+        }
+
+        Ok(Some(from_value::<SamplingMetadata>(sampling_entry)?))
```
zvolin, line 280:
```suggestion
        Ok(Some(from_value(sampling_entry)?))
```

node/src/store/indexed_db_store.rs
```diff
@@ -209,6 +221,64 @@ impl IndexedDbStore {
...
+            .put(&metadata_jsvalue, Some(&height_key))
+            .await?;
+
+        tx.commit().await?;
+
+        let previously_lowest_unsampled = *self.lowest_unsampled_height.borrow();
+        let new_highest_sampled_height =
+            get_next_unsampled_height_from_database(&self.db, previously_lowest_unsampled).await?;
+        self.lowest_unsampled_height
+            .replace(new_highest_sampled_height);
```
zvolin, line 258: are we sure that this is concurrent safe?

node/src/store/sled_store.rs
```diff
@@ -210,6 +236,60 @@ impl SledStore {
...
+        &self,
+        height: u64,
+        accepted: bool,
+        cids: Vec<Cid>,
+    ) -> Result<u64> {
+        let inner = self.inner.clone();
+
+        spawn_blocking(move || {
+            // Make sure we have the header being marked
+            let head_height = read_height_by_db_key(&inner.db, HEAD_HEIGHT_KEY).unwrap_or(0);
```
zvolin, line 249: I think there is a funny corner case when someone could insert sampling metadata for height 0 if store didn't have any header appended yet

```suggestion
            let head_height = read_height_by_db_key(&inner.db, HEAD_HEIGHT_KEY)?;
```

zvolin (commented):
nice, cool to see the migration tests too :rocket: left some questions

node/src/store/sled_store.rs
fl0rek: You're right, it also looks like `contains_height` doesn't always check for 0 correctly.
Fixed both above issues and added unittests for 0-height headers and sampling.

node/src/store/indexed_db_store.rs
fl0rek: As I understand rust-wasm, since there's an await point between the read and write, it's indeed a race condition. 
I've separated the logic into `update_unsampled_height`, which checks whether the unsampled height was updated before write. This should result in behaviour equivalent to atomic compare/exchange (since we're running single threaded).

node/src/store/in_memory_store.rs
```diff
@@ -115,7 +115,7 @@ impl InMemoryStore {
             return false;
         };
 
-        height <= head_height
+        height != 0 && height <= head_height
```
zvolin, line 118: get_head_height returns an error for 0 height already

node/src/store/indexed_db_store.rs
```diff
@@ -222,7 +222,7 @@ impl IndexedDbStore {
             return false;
         };
 
-        height <= head_height
+        height > 0 && height <= head_height
```
zvolin, line 225: here height also can't be zero. It would mean someone put a header with height zero but it would fail on validation

node/src/store/in_memory_store.rs
fl0rek: This covers the case of already filled store. Header with height zero was then incorrectly reported as present here, since 0 < head_height.

node/src/store/indexed_db_store.rs
zvolin: ignore me :sweat_smile:

node/src/store/sled_store.rs
```diff
@@ -246,8 +246,8 @@ impl SledStore {
 
         spawn_blocking(move || {
             // Make sure we have the header being marked
-            let head_height = read_height_by_db_key(&inner.db, HEAD_HEIGHT_KEY).unwrap_or(0);
-            if height > head_height {
+            let head_height = read_height_by_db_key(&inner.db, HEAD_HEIGHT_KEY)?;
+            if height == 0 || height > head_height {
```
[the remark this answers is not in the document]
zvolin: I confused things, sry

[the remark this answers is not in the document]
fl0rek: `height` is coming from the calller, so without this validation they'd be able to insert sampling metadata for header of height 0. Not particularly dangerous, but I don't think that's correct.

node/src/store.rs
```diff
@@ -93,6 +110,19 @@ pub trait Store: Send + Sync + Debug {
...
+
+    /// Sets sampling result for header, updating last sampled height so that it points
+    /// to the next that wasn't verified
+    /// Returns new highest sampled height or error, if occured
+    async fn mark_header_sampled(&self, height: u64, accepted: bool, cids: Vec<Cid>)
+        -> Result<u64>;
+
+    /// Gets the sampling data for height. `Ok(None)` indicates that the sampling data
+    /// wasn't set in store
+    async fn get_sampling_data_for_height(&self, height: u64) -> Result<Option<SamplingMetadata>>;
```
zvolin, line 124: what would you say for `set_sampling_metadata` and `get_sampling_metadata`? I think those are a more common and symmetric

zvolin (approved):
looks good

node/src/store.rs
```diff
@@ -199,6 +233,48 @@ pub enum StoreError {
     InvalidHeadersRange,
 }
 
+#[derive(Message)]
+struct RawSamplingMetadata {
+    #[prost(bool, tag = "1")]
+    pub accepted: bool,
```
oblique, line 239: I think `pub` is not needed

oblique (approved):
a nit

oblique (approved):

PR #210, opened 2024-01-31
feat(rpc): Implement WASM Client

distractedm1nd:
This PR adds a WASM-compatible Client implementation. I am fairly new to Rust, so apologies if there is an easier way to get around the trait implementations.

jsonrpsee-wasm-client only supports Websocket, and also does not have the ability to set headers. This means that celestia-node needs to run with `--rpc.skip-auth`, which is coming out in the next release.

### DIFF
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -12,10 +12,10 @@ readme = "README.md"
 keywords = ["blockchain", "celestia", "lumina"]
 # Must be one of <https://crates.io/category_slugs>
 categories = [
-    "api-bindings",
-    "asynchronous",
-    "encoding",
-    "cryptography::cryptocurrencies",
+  "api-bindings",
+  "asynchronous",
+  "encoding",
+  "cryptography::cryptocurrencies",
 ]
 
 [dependencies]
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -12,10 +12,10 @@ readme = "README.md"
 keywords = ["blockchain", "celestia", "lumina"]
 # Must be one of <https://crates.io/category_slugs>
 categories = [
-  "api-bindings",
-  "asynchronous",
-  "encoding",
-  "cryptography::cryptocurrencies",
+    "api-bindings",
+    "asynchronous",
+    "encoding",
+    "cryptography::cryptocurrencies",
 ]
 
 [dependencies]
@@ -32,13 +32,12 @@ jsonrpsee = { version = "0.20", features = ["http-client", "ws-client"] }
 
 [dev-dependencies]
 libp2p = { workspace = true, features = [
-  "tokio",
-  "macros",
-  "tcp",
-  "noise",
-  "yamux",
+    "tokio",
+    "macros",
+    "tcp",
+    "noise",
+    "yamux",
 ] }
-
 anyhow = "1.0.71"
 dotenvy = "0.15.7"
 futures = "0.3.28"
@@ -47,10 +46,14 @@ rand = "0.8.5"
 tokio = { version = "1.32.0", features = ["rt", "macros"] }
 tracing = "0.1.37"
 
+[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+getrandom = { version = "0.2", features = ["js"] }
+wasm-bindgen-test = "0.3.0"
+
 [features]
 default = ["p2p"]
 p2p = ["celestia-types/p2p"]
-wasm-bindgen = ["celestia-types/wasm-bindgen"]
+wasm-bindgen = ["celestia-types/wasm-bindgen", "jsonrpsee/wasm-client"]
 
 [package.metadata.docs.rs]
 features = ["p2p"]
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -32,11 +32,11 @@ jsonrpsee = { version = "0.20", features = ["http-client", "ws-client"] }
 
 [dev-dependencies]
 libp2p = { workspace = true, features = [
-    "tokio",
-    "macros",
-    "tcp",
-    "noise",
-    "yamux",
+  "tokio",
+  "macros",
+  "tcp",
+  "noise",
+  "yamux",
 ] }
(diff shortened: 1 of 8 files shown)


zvolin:
we could run 2 bridge nodes and only one with `--rpc.skip-auth`. I'm not sure if it's worth it tho and if no other tests asserting something that would break with 2 nodes, like peers they discover or sth

distractedm1nd:
Hey, I had to take an unexpected break for 2 months but I am back. I have implemented the review changes AFAIU, so I think this should be ready to go soon.

The one thing that I'm unsure about is that now run-bridge uses skip-auth, so the auth mechanism is not actually being tested. If there are tests for the negative case I'm assuming CI will fail on this PR

rpc/tests/wasm.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use celestia_rpc::client::Client;
+use celestia_rpc::prelude::*;
+use wasm_bindgen_test::*;
+
+const CELESTIA_RPC_URL: &str = "ws://localhost:26658";
+
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+#[wasm_bindgen_test]
```
fl0rek, line 11: We haven't enabled wasm tests for RPC as there wasn't anything to test yet, so these won't run in CI.
Please add a similar `wasm-pack` line with rpc crate to our [`workflows/ci.yaml`](https://github.com/eigerco/lumina/blob/main/.github/workflows/ci.yml#L98).

Also I see you're setting `run_in_browser`. Is this necessary? At glance these could run in node.js and that I think is slightly less heavy than entire browser.

fl0rek (commented):
Thanks for helping us out!

We still might want to test with auth enabled, but as you've noted we aren't covering negative auth cases now, so that's not a hard requirement and we can fix that later by ourselves.

I think once CI is running and green we can get this merged 🎸

rpc/tests/wasm.rs
distractedm1nd: run_in_browser is unfortunately necessary, because jsonrpsee uses gloo for the networking which uses browser WebSocket API

zvolin (approved):
loooks good

fl0rek (approved):
LGTM!

zvolin:
Hey @distractedm1nd, all looks good but the merge is blocked due to the unsigned commits. Would you mind signing them with your gpg key and force pushing?

rpc/Cargo.toml
```diff
@@ -30,6 +30,9 @@ tracing = "0.1.37"
 http = "0.2.9"
 jsonrpsee = { version = "0.20", features = ["http-client", "ws-client"] }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+jsonrpsee = { version = "0.20", features = ["wasm-client"] }
+
```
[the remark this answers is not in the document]
distractedm1nd: Thanks for the context! Will fix and sign.

distractedm1nd:
@zvolin I've squashed all the commits into a single, signed commit. Hope this is fine

rpc/src/lib.rs
```diff
@@ -11,7 +11,6 @@ mod share;
 mod state;
 
 pub use crate::blob::BlobClient;
-#[cfg(not(target_arch = "wasm32"))]
 pub use crate::client::Client;
```
oblique, line 14: Use this:

```rust
#[cfg(any(
    not(target_arch = "wasm32"),
    all(target_arch = "wasm32", feature = "wasm-bindgen")
))]
#[cfg_attr(
    docsrs,
    doc(cfg(any(
        not(target_arch = "wasm32"),
        all(target_arch = "wasm32", feature = "wasm-bindgen")
    )))
)]
pub use crate::client::Client;
```

rpc/Cargo.toml
```diff
@@ -12,10 +12,10 @@ readme = "README.md"
...
 # Must be one of <https://crates.io/category_slugs>
 categories = [
-  "api-bindings",
-  "asynchronous",
-  "encoding",
-  "cryptography::cryptocurrencies",
+    "api-bindings",
+    "asynchronous",
+    "encoding",
+    "cryptography::cryptocurrencies",
```
oblique, line 18: We use 2 spaces in toml

```suggestion
  "api-bindings",
  "asynchronous",
  "encoding",
  "cryptography::cryptocurrencies",
```

rpc/Cargo.toml
```diff
@@ -32,13 +32,12 @@ jsonrpsee = { version = "0.20", features = ["http-client", "ws-client"] }
...
-  "tokio",
-  "macros",
-  "tcp",
-  "noise",
-  "yamux",
+    "tokio",
+    "macros",
+    "tcp",
+    "noise",
+    "yamux",
```
oblique, line 39:
```suggestion
  "tokio",
  "macros",
  "tcp",
  "noise",
  "yamux",
```

rpc/Cargo.toml
```diff
@@ -32,13 +32,12 @@ jsonrpsee = { version = "0.20", features = ["http-client", "ws-client"] }
 
 [dev-dependencies]
```
oblique, line 33: Change this to `[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]`

rpc/Cargo.toml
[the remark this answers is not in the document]
zvolin: > why are we requiring it behind a (browser-bound) wasm-bindgen feature flag?

The thing is that this particular client will be only enabled when we run in browser. In other wasm environments that supports http/ws there is still a possibility to implement the client manually and still use traits provided by this crate, but if we won't guard it by a feature but rather just a `wasm32` it will fail to compile.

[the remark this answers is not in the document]
distractedm1nd: I see, that makes sense - thanks!

oblique (approved):

zvolin (approved):

fl0rek (approved):

oblique:
Thank @distractedm1nd!

PR #217, opened 2024-02-07
feat!(blockstore): add sled blockstore

zvolin:


blockstore/src/sled_blockstore.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub async fn new(db: Db) -> Result<Self> {
+        let blockstore = spawn_blocking(move || {
+            let blocks = db.open_tree(BLOCKS_TREE_ID)?;
+
+            Ok::<_, SledError>(Self {
+                inner: Arc::new(Inner { _db: db, blocks }),
+            })
+        })
+        .await??;
+        Ok(blockstore)
```
fl0rek, line 48:
```suggestion
        spawn_blocking(move || {
            let blocks = db.open_tree(BLOCKS_TREE_ID)?;

            Ok(Self {
                inner: Arc::new(Inner { _db: db, blocks }),
            })
        })
        .await?
```
gets rid of manually specifying Err type, wdyt?

blockstore/src/lib.rs
```diff
@@ -31,6 +36,21 @@ pub enum BlockstoreError {
     /// Error occured when trying to compute CID.
     #[error("Error generating CID: {0}")]
     CidError(#[from] CidError),
+
+    /// An error propagated from the IO operation.
+    #[cfg(feature = "sled")]
```
oblique, line 41: This will cause a compilation issue if a library that doesn't use `sled` feature, does a `match` case that handles all the errors except the `sled` ones and another library enables the `sled` feature. We have two options:

1. We make `BlockstoreError` non_exhaustive
2. Remove feature gating on the error variants

zvolin: well, `non_exhaustive` may be a future compatible way for adding any errors without breaking api. which way you suggest?

fl0rek: Newly added variants represent categories of errors that can happen regardless of specific persistent storage method being used. My instinct is towards the second option, that's what we have in header store (and I'm biased since I've put it there :P)

oblique (approved):

fl0rek (approved):
🎸 nice 🎸

PR #218, opened 2024-02-07
feat!(node): use generic blockstore in node

zvolin:


cli/src/native.rs
```diff
@@ -84,6 +88,37 @@ pub(crate) async fn run(args: Params) -> Result<()> {
...
+            fs::rename(old_cache_dir, &cache_dir).await?;
+        }
+    }
+
+    cache_dir.push(network_id);
+    // TODO: should we create there also a subdirectory for the 'db'
+    // in case we want to put there some other stuff too?
+    let db = spawn_blocking(|| sled::open(cache_dir)).await??;
+    Ok(db)
+}
```
oblique, line 120: Should we provide this as a helper in `lumina-node`? Before this PR a user could use `SledStore` in `Node` and continue from the state `lumina-cli` had. We can name it `open_default_sled_db`.

node/src/p2p.rs
```diff
@@ -146,21 +145,18 @@ impl From<oneshot::error::RecvError> for P2pError {
 
 /// Component responsible for the peer to peer networking handling.
 #[derive(Debug)]
-pub struct P2p<S>
-where
-    S: Store + 'static,
-{
+pub struct P2p {
```
oblique, line 148: In the end, I'm going to need `Store` when I remove `async` from `Multihasher`, so I think we should keep it.

cli/src/native.rs
zvolin: I actually liked the removal of this logic from the `lumina-node` but I can move it back there, wdyt @fl0rek

node/src/p2p.rs
zvolin: won't it be needed only in `Worker`?

cli/src/native.rs
```diff
@@ -84,6 +88,37 @@ pub(crate) async fn run(args: Params) -> Result<()> {
...
+    if let Some(path) = path {
+        let db = spawn_blocking(|| sled::open(path)).await??;
+        return Ok(db);
+    }
+
+    let cache_dir =
+        ProjectDirs::from("co", "eiger", "lumina").context("Couldn't find lumina's cache dir")?;
+    let mut cache_dir = cache_dir.cache_dir().to_owned();
+
+    // TODO: remove it after we publish a new version to crates.io?
```
fl0rek, line 101: I think we might need to leave it up for few versions, for everybody to get a chance to run the version with migration code. Not sure how important is it? We lose few hours of download, which is bad, but not terrible.

cli/src/native.rs
oblique: For me the main issue is that users need to inspect our code to figure out that we are using cache directory of `ProjectDirs` and then do the opening of the sled db. If we provide them a helper, we are making this part of our API instead.

fl0rek: Maybe we can add `pub fn lumina_default_db() -> PathBuf` for that case?

cli/src/native.rs
zvolin: changed the todo

oblique (approved):
lgtm

fl0rek (approved):
nice 🚀

PR #223, opened 2024-02-12
feat!(node): Implement DASer

oblique:


node/src/daser.rs
```diff
@@ -0,0 +1,189 @@
...
+    Store(#[from] StoreError),
+}
+
+pub struct Daser<S>
+where
+    S: Store + 'static,
+{
+    cmd_tx: mpsc::Sender<DaserCmd>,
+    cancellation_token: CancellationToken,
+    _store: PhantomData<S>,
```
zvolin, line 39: we could get rid of that

node/src/daser.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+pub struct DaserArgs<S>
+where
+    S: Store + 'static,
```
zvolin, line 44: I think `+ 'static` is only needed on the `start` fn

node/src/store/in_memory_store.rs
```diff
@@ -211,6 +216,26 @@ impl Store for InMemoryStore {
...
     }
 
+    async fn wait_height(&self, height: u64) -> Result<()> {
+        let mut notifier = pin!(self.check_height_notifier.notified());
+
+        loop {
+            // Make sure no notifications are lost by enable notifier
+            // before the check.
+            notifier.as_mut().enable();
+
```
zvolin, line 226: according to docs, the `Notified::enable` is only needed in case we don't want to lost the notifications coming from `notify_one`. As we are using the `notify_waiters` this has no effect

oblique: true, removed

node/Cargo.toml
```diff
@@ -34,8 +34,8 @@ libp2p = { workspace = true, features = [
 ] }
 
 async-trait = "0.1.73"
-beetswap = { git = "https://github.com/eigerco/beetswap", rev = "9316bd6" }
-cid = { version = "0.11.0", features = ["serde-codec"] }
+beetswap = { git = "https://github.com/eigerco/beetswap", rev = "f69ddd8" }
+cid = { version = "0.11.0", features = [ "serde-codec" ] }
```
zvolin, line 38:
```suggestion
cid = { version = "0.11.0", features = ["serde-codec"] }
```

types/src/sample.rs
```diff
@@ -111,20 +110,14 @@ impl Sample {
...
         };
 
-        let mut tree = Nmt::with_hasher(NamespacedSha2Hasher::with_ignore_max_ns(true));
-
-        let shares = eds.axis(axis_type, axis_index)?;
-        let (data_shares, parity_shares) = shares.split_at(square_len / 2);
+        let (row_index, column_index) = match axis_type {
+            AxisType::Row => (axis_index, sample_index),
+            AxisType::Col => (sample_index, axis_index),
+        };
```
zvolin, line 116: So this reverses the indexing from the previous match, making both of them pointless. If we want to have this, then we could remove both matches and just have
```
let (row_index, column_index) = (index / square_len, index % square_len);
```

Tho I think it was correct before and now it's not, because now axis doesn't matter at all

types/src/rsmt2d.rs
```diff
@@ -265,6 +326,13 @@ impl<'de> Deserialize<'de> for ExtendedDataSquare {
     }
 }
 
+/// Returns true if and only if the provided coordinates belongs to Original Data Square
+/// (i.e. first quadrant of Extended Data Square).
+pub fn is_ods_square(row: usize, column: usize, square_len: usize) -> bool {
```
zvolin, line 331: do we want to make this public api? I'm not sure

types/src/sample.rs
oblique: I will clean this up on my next PR

types/src/rsmt2d.rs
oblique: It would be useful if our users need to interact with ODS, but I'm also not sure.

types/src/sample.rs
zvolin: but shouldn't this be removed from this PR?

types/src/rsmt2d.rs
zvolin: let's make it `pub(crate)` for now

node/src/daser.rs
```diff
@@ -0,0 +1,294 @@
...
+    async fn sample_next_block(&mut self) -> Result<()> {
+        let height = self.store.next_unsampled_height().await?;
+        self.store.wait_height(height).await?;
+
+        let header = self.store.get_by_height(height).await?;
+        let (sampled_cids, accepted) = self.sample_block(&header).await?;
+
+        self.store
+            .update_sampling_metadata(height, accepted, sampled_cids)
+            .await?;
```
zvolin, line 136: could you add a log here that we sampled some height and tell the result?

types/src/rsmt2d.rs
```diff
@@ -135,17 +135,18 @@ impl TryFrom<u8> for AxisType {
 pub struct ExtendedDataSquare {
     /// The raw data of the EDS.
     #[serde(with = "celestia_tendermint_proto::serializers::bytes::vec_base64string")]
-    pub data_square: Vec<Vec<u8>>,
+    data_square: Vec<Vec<u8>>,
```
zvolin, line 138: breaking change detected :smile: please align the title

types/src/rsmt2d.rs
```diff
@@ -206,29 +263,33 @@ impl ExtendedDataSquare {
...
             let mut shares = Vec::with_capacity(self.square_len);
-            let mut tree = Nmt::with_hasher(NamespacedSha2Hasher::with_ignore_max_ns(true));
-            for (col, s) in self.row(i.into())?.iter().enumerate() {
-                let ns = if col < self.square_len / 2 {
-                    Namespace::from_raw(&s[..NS_SIZE])?
+
+            for col in 0..self.square_len {
+                let share = self.share(row, col)?;
+
+                let ns = if is_ods_square(row, col, self.square_len) {
```
zvolin, line 280: now that I think about it, since the row is ordered by the namespace, we could have something like this after this `let ns ...`:

```rust
if ns > namespace {
    break;
}
```

zvolin (commented):
nice, left some ideas

types/src/rsmt2d.rs
```diff
@@ -159,45 +160,101 @@ impl ExtendedDataSquare {
...
     pub fn axis(&self, axis: AxisType, index: usize) -> Result<Vec<Vec<u8>>> {
         match axis {
             AxisType::Col => self.column(index),
             AxisType::Row => self.row(index),
         }
     }
 
-    /// Get EDS square length
+    /// Returns the [`Nmt`] of column or row.
+    pub fn axis_nmt(&self, axis: AxisType, index: usize) -> Result<Nmt> {
```
fl0rek, line 244: `column_nmt` and `row_nmt` differ only when they fetch share from eds. I wonder whether I'd be more succinct to have shared code in `axis_nmt` and have `columnt_nmt` and `row_nmt` call it with appropriate `AxisType`?

node/src/daser.rs
```diff
@@ -0,0 +1,294 @@
+//! Component responsible for data availability sampling of the already synchronized block
+//! headers announced in the Celestia network.
+//!
+//! When a new header is insert in the [`Store`], [`Daser`] gets informed, then it fetches
+//! random [`Sample`]s of the block via Shwap protocol and verifies them. If all samples
+//! get verified successfuly, then block is marked as accepted.
```
fl0rek, line 6:
```suggestion
//! When a new header is inserted into the [`Store`], [`Daser`] gets notified. It then fetches
//! random [`Sample`]s of the block via Shwap protocol and verifies them. If all the samples
//! get verified successfuly, then block is marked as accepted.
```


Also maybe we should add something like "Otherwise, if `Daser` doesn't receive valid samples before `GET_SAMPLE_TIMEOUT`, block is marked as not accepted and dasing continues". 

I think this is implied by the first part, but you have to know how the code works to know what "get verified successfuly" means.

node/src/daser.rs
```diff
@@ -0,0 +1,294 @@
...
+use tokio::select;
+use tokio_util::sync::CancellationToken;
+use tracing::error;
+
+use crate::executor::spawn;
+use crate::p2p::shwap::convert_cid;
+use crate::p2p::{P2p, P2pError};
+use crate::store::{Store, StoreError};
+
+const DEFAULT_MAX_SAMPLES_NEEDED: usize = 16;
```
fl0rek, line 27:
```suggestion
const MAX_SAMPLES_NEEDED: usize = 16;
```

As is, I don't think it's configurable, so we could drop the prefix?

types/src/sample.rs
fl0rek: This reverses the indexing from the previous match, but on separate variables. `eds.axis_nmt` uses `axis_index`, `build_range_proof` uses `sample_index` to index along the axis; while `row/col_index` are used just for getting the share. 
So I think while mixing the two indexing methods is a bit misleading, it's still correct. Not sure how to eliminate that cleanly.

fl0rek: Oh and original bug was `shares[sample_index]` rather than intended `shares[index]`, it should have used "global" index, rather than offset along the axis. 🙈

oblique: > Not sure how to eliminate that cleanly.

I didn't open a PR yet, but this is how I did it: https://github.com/eigerco/lumina/commit/c43ca77b84c58eae7883ff4546939206e28c8281

oblique: Also I think the code of this PR is correct because `axis_nmt` was reversing indexes again in case of `AxisType::Col`.

types/src/rsmt2d.rs
oblique: I was thinking about this too, but couldn't beside because indexes cause us a lot of confusion and I wanted to avoid a `match` that was selecting indexes. But I can change it.

node/src/daser.rs
```diff
@@ -0,0 +1,294 @@
...
+            futs.push(fut);
+        }
+
+        let mut cids: Vec<Cid> = Vec::new();
+        let mut accepted = true;
+
+        while let Some(res) = futs.next().await {
+            match res {
+                Ok(sample) => cids.push(convert_cid(&sample.sample_id.into())?),
+                Err(P2pError::BitswapQueryTimeout) => accepted = false,
```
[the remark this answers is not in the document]
zvolin: Yeah just so far that's the only place which has kind of info what sample was rejected. Maybe we should add it into sampling metadata in a future. Anyway I wouldn't mind something on trace level here but up to you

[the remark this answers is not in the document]
zvolin: Or maybe we could instrument get shwap cid on trace level?

types/src/rsmt2d.rs
oblique: done

types/src/rsmt2d.rs
```diff
@@ -206,29 +263,33 @@ impl ExtendedDataSquare {
     ) -> Result<Vec<NamespacedData>> {
         let mut data = Vec::new();
 
-        for i in 0u16..self.square_len as u16 {
-            let row_root = dah.row_root(i.into()).unwrap();
+        for row in 0..self.square_len {
+            let Some(row_root) = dah.row_root(row) else {
+                continue;
```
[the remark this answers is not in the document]
oblique: Changed it to `break` for now

types/src/rsmt2d.rs
oblique: done

node/src/daser.rs
oblique: Done

node/src/daser.rs
oblique: Isn't this an implementation detail because of how Shwap works?

node/src/store/in_memory_store.rs
```diff
@@ -23,6 +25,8 @@ pub struct InMemoryStore {
     head_height: AtomicU64,
     /// Cached height of the lowest header that wasn't sampled yet
     lowest_unsampled_height: AtomicU64,
+    /// Notifier for height
+    check_height_notifier: Notify,
```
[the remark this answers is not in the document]
oblique: `new_header_notifier` reminds me a constructor. Another option is `height_changed_notifier`. Wdyt?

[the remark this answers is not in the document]
fl0rek: `header_added_notifier` sounds perfect to me 👍

types/src/rsmt2d.rs
fl0rek: Not feeling too strongly about it, thought I'd make code more succint. I've had something like this in mind. 
```rust 
     pub fn axis_nmt(&self, axis: AxisType, index: usize) -> Result<Nmt> {
        let mut tree = Nmt::with_hasher(NamespacedSha2Hasher::with_ignore_max_ns(true));

        let indices : Vec<_> = match axis {
            AxisType::Row => (0..self.square_len).zip(std::iter::repeat(index)).collect(),
            AxisType::Col => std::iter::repeat(index).zip(0..self.square_len).collect(),
        };

        for (x, y) in indices {
            let share = self.share(x, y)?;

            let ns = if is_ods_square(x, y, self.square_len) {
                Namespace::from_raw(&share[..NS_SIZE])?
            } else {
                Namespace::PARITY_SHARE
            };

            tree.push_leaf(share, *ns).map_err(Error::Nmt)?;
         }

        Ok(tree)
     }
```

Having to `collect` iterators, as they have different types, doesn't seem great though (also was this the match you were talking about? :P). So feel free to keep existing.

node/src/daser.rs
fl0rek: yeah, I'd definitely drop the reference to the exact constant name. 👍 

I'm torn on the timeout, because I think it might be considered a shwap implementation detail, but it's also pretty important detail to get wrong when trying to understand DASer. 

Maybe something vaguer? "If daser is unable to get samples before shwap timeout, sample is considered unavailable, block mared as not accepted and dasing continues". Since interpretation and consequences of bitswap timeout are within DASer purview.

node/src/store/in_memory_store.rs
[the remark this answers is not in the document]
oblique: done

types/src/rsmt2d.rs
oblique: No I was thinking for something like this:

```rust
    pub fn axis_nmt(&self, axis: AxisType, index: usize) -> Result<Nmt> {
        let mut tree = Nmt::with_hasher(NamespacedSha2Hasher::with_ignore_max_ns(true));

        for i in 0..self.square_len {
            let (row, col) = match axis {
                AxisType::Row => (index, i),
                AxisType::Col => (i, index),
            };

            let share = self.share(row, col)?;

            let ns = if is_ods_square(col, row, self.square_len) {
                Namespace::from_raw(&share[..NS_SIZE])?
            } else {
                Namespace::PARITY_SHARE
            };

            tree.push_leaf(share, *ns).map_err(Error::Nmt)?;
        }

        Ok(tree)
    }
 ```

Comment (oblique) [reply types/src/rsmt2d.rs]: done

File: node/src/daser.rs (hunk 7 above)
Comment (oblique) [reply node/src/daser.rs:137]: done

File: node/src/daser.rs [hunk 16]
```diff
@@ -0,0 +1,309 @@
...
+        let mut accepted = true;
+
+        while let Some(res) = futs.next().await {
+            match res {
+                Ok(sample) => cids.push(convert_cid(&sample.sample_id.into())?),
+                // Validation is done in Bitswap level thought `ShwapMultihasher`.
+                // When a sample is not valid, then is will never be delivered
+                // to us as the data of the CID. Because of that, the only way
+                // to know that data sampling verification failed is when a query
+                // times out.
```
fl0rek, line 166:
```suggestion
                // Validation is done at Bitswap level, though `ShwapMultihasher`.
                // If the sample is not valid, it will never be delivered to us
                // as the data of the CID. Because of that, the only signal 
                // that data sampling verification failed is query timing out.
```

node/src/daser.rs
```diff
@@ -0,0 +1,309 @@
...
+            now.elapsed()
+        );
+
+        Ok((cids, accepted))
+    }
+}
+
+fn random_indexes(block_len: usize, max_samples_needed: usize) -> HashSet<usize> {
+    // If block length is smaller than `max_samples_needed`, we are going
+    // to sample the whole block. Randomation is not needed for this.
```
fl0rek, line 185:
```suggestion
    // to sample the whole block. Randomness is not needed for this.
```

node/src/daser.rs
```diff
@@ -0,0 +1,309 @@
+//! Component responsible for data availability sampling of the already synchronized block
+//! headers announced in the Celestia network.
```
fl0rek, line 2:
```suggestion
//! headers announced on the Celestia network.
```

fl0rek (approved):
Nice work 🚀 
Code looks good to me, left my last grammar nits :D

node/src/daser.rs
zvolin: typo in suggestion, 'through'

zvolin (approved):

PR #224, opened 2024-02-19
fix!(blockstore): remove an error if cid already exists

zvolin:


fl0rek (approved):
🚀 

Would you mind adding information about new allowed beheaviour in commit header/message? Something like "Allow overwriting existing blockstore CIDs, instead of returning an error", I feel like the first part is silent now

blockstore/src/in_memory_blockstore.rs
```diff
@@ -22,13 +21,7 @@ impl<const MAX_MULTIHASH_SIZE: usize> InMemoryBlockstore<MAX_MULTIHASH_SIZE> {
...
 
     fn insert_cid(&self, cid: CidGeneric<MAX_MULTIHASH_SIZE>, data: &[u8]) -> Result<()> {
-        let cid_entry = self.map.entry(cid);
-        if matches!(cid_entry, Entry::Occupied(_)) {
-            return Err(BlockstoreError::CidExists);
-        }
-
-        cid_entry.insert(data.to_vec());
-
+        self.map.insert(cid, data.to_vec());
```
oblique, line 24: CID on the bitswap protocol can represent data that are up to 2MiB. So, as an optimization, I think we shouldn't reinsert if something already exists.

zvolin: done

fl0rek (approved):

blockstore/src/lru_blockstore.rs
```diff
@@ -30,12 +30,8 @@ impl<const MAX_MULTIHASH_SIZE: usize> Blockstore for LruBlockstore<MAX_MULTIHASH
     async fn put_keyed<const S: usize>(&self, cid: &CidGeneric<S>, data: &[u8]) -> Result<()> {
         let cid = convert_cid(cid)?;
         let mut cache = self.cache.lock().expect("lock failed");
-        if !cache.contains(&cid) {
-            cache.put(cid, data.to_vec());
-            Ok(())
-        } else {
-            Err(BlockstoreError::CidExists)
-        }
+        cache.put(cid, data.to_vec());
```
oblique, line 33:
```suggestion
        if cache.contains(&cid) {
            cache.promote(&cid);
        } else {
            cache.put(cid, data.to_vec());
        }
```

oblique (approved):

PR #225, opened 2024-02-19
chore: remove deny warnings from cargo udeps

zvolin:
this step is failing again... If we will want to have warning-less compilation of nightly, we should add nightly clippy run


oblique (approved):
lgtm

PR #231, opened 2024-02-27
feat(types): add encoding check when verifying befp

zvolin:


types/src/byzantine.rs
```diff
@@ -110,25 +110,74 @@ impl FraudProof for BadEncodingFraudProof {
...
+            .shares
+            .iter()
+            .map(|share| {
+                share
+                    .as_ref()
+                    .map(|sh| sh.leaf.share.to_vec())
+                    .unwrap_or_default()
+            })
+            .collect();
+        let data_shares = rebuilt_shares.len() / 2;
```
fl0rek, line 133: I find `*_shares` variables containing shares in one place and number of shares in another slightly confusing to read. Could we do `data_shares_number` here?

types/src/byzantine.rs
```diff
@@ -110,25 +110,74 @@ impl FraudProof for BadEncodingFraudProof {
...
+        if leopard_codec::reconstruct(&mut rebuilt_shares, data_shares).is_err() {
+            // we couldn't reconstruct the data even tho we had enough *proven* shares
+            // befp is legit
+            return Ok(());
+        }
+
+        // re-encode the parity data
+        if leopard_codec::encode(&mut rebuilt_shares, data_shares).is_err() {
+            // NOTE: this is unreachable for current implementation of encode, esp since
```
fl0rek, line 142: nit:
```suggestion
            // NOTE: this is unreachable for current implementation of encode, esp. since
```
or "especially", or "esp32" :P

types/src/byzantine.rs
zvolin: how about now?

types/src/rsmt2d.rs
```diff
@@ -601,7 +633,7 @@ mod tests {
                 parity_share.clone(),
                 // row 1
                 share(1), // ODS
-                share(1), // ODS
+                share(1), // ODS - error: smaller namespace in 2nd column
```
oblique, line 636: I would like 1 extra example that passes in this case. For example with:

```rust
                // row 0
                share(1), // ODS
                share(2), // ODS
                parity_share.clone(),
                parity_share.clone(),
                // row 1
                share(1), // ODS
                share(3), // ODS
                parity_share.clone(),
                parity_share.clone(),
```

types/src/data_availability_header.rs
```diff
@@ -53,16 +53,32 @@ use crate::{bail_validation, Error, Result, ValidateBasic, ValidationError};
...
+            row_roots: Vec::with_capacity(square_len),
+            column_roots: Vec::with_capacity(square_len),
+        };
+
+        for i in 0..square_len {
+            let row_root = eds.row_nmt(i).unwrap().root();
+            dah.row_roots.push(row_root);
+
+            let column_root = eds.column_nmt(i).unwrap().root();
+            dah.column_roots.push(column_root);
```
oblique, line 76:
```suggestion
            let row_root = eds.row_nmt(i).expect("EDS validated on construction").root();
            dah.row_roots.push(row_root);

            let column_root = eds.column_nmt(i).expect("EDS validated on construction").root();
            dah.column_roots.push(column_root);
```

types/src/byzantine.rs
```diff
@@ -109,26 +109,78 @@ impl FraudProof for BadEncodingFraudProof {
...
+                    .clone()
+                    .map(|share_with_proof| share_with_proof.leaf.share)
+                    .unwrap_or_default()
+            })
+            .collect();
+        let ods_width = rebuilt_shares.len() / 2;
+        if leopard_codec::reconstruct(&mut rebuilt_shares, ods_width).is_err() {
+            // we couldn't reconstruct the data even tho we had enough *proven* shares
+            // befp is legit
+            return Ok(());
```
oblique, line 140: So is this `Ok(())` because we confirm that there is a bad encoding?

zvolin: yes

types/src/rsmt2d.rs
zvolin: added

types/src/byzantine.rs
```diff
@@ -214,84 +274,229 @@ impl TryFrom<RawBadEncodingFraudProof> for BadEncodingFraudProof {
...
+        for share in proof.shares.iter_mut().step_by(2) {
+            *share = None
+        }
+
+        for ns in proof
+            .shares
+            .iter()
+            .map(|sh| sh.as_ref().map(|sh| sh.leaf.namespace.as_bytes()))
+        {
+            println!("{ns:?}");
```
fl0rek, line 424: making sure, is this print intentional?

zvolin: leftover, removed

fl0rek (approved):
🎸

types/src/rsmt2d.rs
```diff
@@ -151,54 +144,144 @@ pub struct ExtendedDataSquare {
...
+        const MAX_SHARES: usize = MAX_EXTENDED_SQUARE_WIDTH * MAX_EXTENDED_SQUARE_WIDTH;
+
         if shares.len() < MIN_SHARES {
             bail_validation!(
                 "shares len ({}) < MIN_SHARES ({})",
                 shares.len(),
                 MIN_SHARES
             );
         }
+        if shares.len() > MAX_EXTENDED_SQUARE_WIDTH.pow(2) {
```
zvolin, line 170:
```suggestion
        if shares.len() > MAX_SHARES {
```

types/src/rsmt2d.rs
```diff
@@ -151,54 +144,144 @@ pub struct ExtendedDataSquare {
...
+        if ods_width * ods_width != ods_shares.len() {
+            return Err(Error::EdsInvalidDimentions);
+        }
 
-                prev_ns = Some(ns);
+        let eds_width = ods_width * 2;
+        let mut eds_shares = Vec::with_capacity(eds_width * eds_width);
+        // take rows of ods and interleave them with parity shares
+        for _ in 0..ods_width {
+            eds_shares.extend(ods_shares.drain(..ods_width));
```
oblique, line 262: Do we need to check that shares from ODS actually have size of `SHARE_LEN`? Or `leopard_codec::encode` will take care of that?

zvolin: leopard will return an error then

zvolin: and even if not and encoding somehow worked, then `new` would

zvolin: https://github.com/eigerco/leopard-codec/blob/main/src/lib.rs#L42

oblique (approved):
Nice!

PR #232, opened 2024-02-28
feat!(types): Align with Shwap spec

oblique:


Wondertan (approved):

types/src/error.rs
```diff
@@ -101,6 +101,10 @@ pub enum Error {
     #[error("Invalid axis type: {0}")]
     InvalidAxis(i32),
 
+    /// Invalid proof type in Protobuf.
+    #[error("Invalid proof type: {0}")]
+    InvalidProofType(i32),
```
zvolin, line 106: this could easily be confused with `Error::WrongProofType`, which iirc is about Presence / Absence merkle proofs. Can we make them a bit more distinct?

zvolin (approved):

fl0rek (approved):
nice 👍

PR #233, opened 2024-02-28
feat(node): implement fraud-sub and services stopping on valid befp

### LINKED ISSUE #194: Listen on fraud-sub topic and implement a watcher similar to header-sub-watcher
- [ ] add the fraud-sub topic
- [ ] add the fraud_sub_watcher like header_sub_watcher
- [ ] optional: create some fraud proof in go node and have some tests where we announce it. We aren't going to be distributing those proofs from our side anytime soon, so we don't need constructing and announcing part in our node itself

zvolin:
closes #194

### DIFF
--- a/node/src/p2p.rs
+++ b/node/src/p2p.rs
@@ -3,10 +3,11 @@
 //! It is a high level integration of various p2p protocols used by Celestia nodes.
 //! Currently supporting:
 //! - libp2p-identitfy
-//! - header-sub topic on libp2p-gossipsub
 //! - libp2p-kad
 //! - libp2p-autonat
 //! - libp2p-ping
+//! - header-sub topic on libp2p-gossipsub
+//! - fraud-sub topic on libp2p-gossipsub
 //! - header-ex client
 //! - header-ex server
 
@@ -20,12 +21,12 @@ use std::time::Duration;
 use blockstore::Blockstore;
 use celestia_proto::p2p::pb::{header_request, HeaderRequest};
 use celestia_tendermint_proto::Protobuf;
-use celestia_types::hash::Hash;
 use celestia_types::namespaced_data::NamespacedData;
 use celestia_types::nmt::Namespace;
 use celestia_types::row::Row;
 use celestia_types::sample::Sample;
-use celestia_types::ExtendedHeader;
+use celestia_types::{fraud_proof::BadEncodingFraudProof, hash::Hash};
+use celestia_types::{ExtendedHeader, FraudProof, Height};
 use cid::Cid;
 use futures::StreamExt;
 use instant::Instant;
@@ -44,6 +45,7 @@ use libp2p::{
 use smallvec::SmallVec;
 use tokio::select;
 use tokio::sync::{mpsc, oneshot, watch};
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, instrument, trace, warn};
 
 mod header_ex;
@@ -51,7 +53,6 @@ mod header_session;
 pub(crate) mod shwap;
 mod swarm;
 
-use crate::executor::{self, spawn, Interval};
 use crate::p2p::header_ex::{HeaderExBehaviour, HeaderExConfig};
 use crate::p2p::header_session::HeaderSession;
 use crate::p2p::shwap::{namespaced_data_cid, row_cid, sample_cid, ShwapMultihasher};
@@ -62,6 +63,10 @@ use crate::store::Store;
 use crate::utils::{
     celestia_protocol_id, gossipsub_ident_topic, MultiaddrExt, OneshotResultSender,
     OneshotResultSenderExt, OneshotSenderExt,
+};
+use crate::{
+    executor::{self, spawn, Interval},
+    utils::fraudsub_ident_topic,
 };
 
 pub use crate::p2p::header_ex::HeaderExError;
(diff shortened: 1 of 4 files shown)


node/src/p2p.rs
```diff
@@ -899,6 +937,52 @@ where
...
+            // does this threshold make any sense if we're gonna ignore it anyway
+            // since we won't have the header
+            return gossipsub::MessageAcceptance::Ignore;
+        }
+
+        let hash = befp.header_hash();
+        let Ok(header) = self.store.get_by_hash(&hash).await else {
+            // we can't verify the proof without a header
+            // TODO: should we then store it and wait for the height? celestia doesn't
+            return gossipsub::MessageAcceptance::Ignore;
```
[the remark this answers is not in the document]
zvolin: ignore means that we don't propagate this message and don't change anything in peer scores and we drop the message. Whether we'll get it again in a future is a big `?`

node/src/p2p.rs
```diff
@@ -759,16 +794,21 @@ where
...
+                } else if message.topic == self.bad_encoding_fraud_sub_topic {
+                    self.on_bad_encoding_fraud_sub_message(&message.data[..], &peer)
+                        .await
                 } else {
                     trace!("Unhandled gossipsub message");
                     gossipsub::MessageAcceptance::Ignore
                 };
 
+                if !matches!(acceptance, gossipsub::MessageAcceptance::Reject) {
+                    // We may discovered a new peer
```
fl0rek, line 808:
```suggestion
                    // We may have discovered a new peer
```

node/src/p2p.rs
```diff
@@ -899,6 +939,55 @@ where
...
     }
+
+    #[instrument(skip_all)]
+    async fn on_bad_encoding_fraud_sub_message(
+        &mut self,
+        data: &[u8],
+        peer: &PeerId,
+    ) -> gossipsub::MessageAcceptance {
+        let Ok(befp) = BadEncodingFraudProof::decode(data) else {
+            trace!("Malformed or invalid bad encoding fraud proof from {peer}");
```
fl0rek, line 950: As I understand it, this branch is for malformed proof we couldn't parse, proof validity is checked later. (If we care to make that distinction, nitpicking here).

fl0rek (approved):
nice, quick work 👏

node/src/p2p.rs
```diff
@@ -899,6 +939,55 @@ where
...
+
+        let height = befp.height().value();
+        let current_height =
+            if let Some(network_height) = network_head_height(&self.header_sub_watcher) {
+                network_height.value()
+            } else if let Ok(local_head) = self.store.get_head().await {
+                local_head.height().value()
+            } else {
+                // we aren't tracking the network and have uninitialized store
+                return gossipsub::MessageAcceptance::Ignore;
```
oblique, line 963: We have a small edge case here: We can receive BEFP while we are fetching genesis. The window is small but still possible. However this can be solved if we request the BEFPs after we initialize genesis. We should keep this in mind for BEFP exchange.

node/src/p2p.rs
```diff
@@ -899,6 +939,55 @@ where
...
+            };
+
+        if height > current_height + FRAUD_PROOF_HEAD_HEIGHT_THRESHOLD {
+            // does this threshold make any sense if we're gonna ignore it anyway
+            // since we won't have the header
+            return gossipsub::MessageAcceptance::Ignore;
+        }
```
oblique, line 970: So how a new node will know that BFEP was published while it was syncing?

zvolin: when we have the fraud-sync, we'll constantly be asking nodes for the fraud proofs, that way we'll catch up with them after we sync enough to be able to validate

node/src/p2p.rs
zvolin: yeah, this will also be handled by fraud-sync

node/src/p2p.rs
```diff
@@ -899,6 +939,55 @@ where
...
+
+    #[instrument(skip_all)]
+    async fn on_bad_encoding_fraud_sub_message(
+        &mut self,
+        data: &[u8],
+        peer: &PeerId,
+    ) -> gossipsub::MessageAcceptance {
+        let Ok(befp) = BadEncodingFraudProof::decode(data) else {
+            trace!("Malformed or invalid bad encoding fraud proof from {peer}");
+            self.swarm.behaviour_mut().gossipsub.blacklist_peer(peer);
```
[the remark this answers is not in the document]
zvolin: I think it can be needed in the future

node/src/p2p.rs
zvolin:
```suggestion
            trace!("Malformed bad encoding fraud proof from {peer}");
```

node/src/p2p.rs
```diff
@@ -481,6 +492,19 @@ impl P2p {
         })
         .await
     }
+
+    /// Get the cancellation token which will be cancelled when the network gets compromised.
+    ///
+    /// After this token is cancelled, the network should be treated as insincere
+    /// and should not be trusted.
+    pub async fn get_network_compromised_token(&self) -> Result<CancellationToken> {
```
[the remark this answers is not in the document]
zvolin: switched to pub(crate) to not include this in public api yet

node/src/p2p.rs
fl0rek: If my reading is correct, go implementation [only requests proofs from 5 peers ](https://github.com/celestiaorg/go-fraud/blob/main/fraudserv/sync.go#L27) and calls it a day. 
If we want to mirror that it seems like we'd need to fraud-sync after being reasonably cought up with the network, otherwise we'd be dropping proofs on the floor again.

zvolin: hmm I thought it's 5 peers at once but it's a continuously running process :detective: we'll rethink that when we will be writing the sync tho, maybe we will differ in behavior there

zvolin: btw it seems like in some way, those nodes re-announce the proofs when they get them, so maybe we will get those proofs again on the fraud-sub anyway

oblique (approved):

PR #241, opened 2024-03-08
feat(types): Add constructor for creating EDS of an empty block

oblique:


types/src/rsmt2d.rs
```diff
@@ -237,6 +237,19 @@ impl ExtendedDataSquare {
         Ok(eds)
     }
 
+    /// Crate a new EDS that represents an empty block
+    pub fn empty() -> ExtendedDataSquare {
```
oblique, line 241: I wasn't sure which is better, `empty` or `empty_block`

zvolin: I like empty

zvolin (approved):

fl0rek (approved):
🚀

PR #254, opened 2024-03-09
blob index field

S1nus:
celestia-node 13.1 adds the index field to the Blob type

### DIFF
--- a/types/src/blob.rs
+++ b/types/src/blob.rs
@@ -28,6 +28,7 @@ pub struct Blob {
     pub data: Vec<u8>,
     pub share_version: u8,
     pub commitment: Commitment,
+    pub index: u64,
 }
 
 impl Blob {
@@ -40,6 +41,7 @@ impl Blob {
             data,
             share_version: appconsts::SHARE_VERSION_ZERO,
             commitment,
+            index: -1
         })
     }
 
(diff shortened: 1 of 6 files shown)


types/src/blob.rs
```diff
@@ -28,6 +28,12 @@ pub struct Blob {
     pub data: Vec<u8>,
     pub share_version: u8,
     pub commitment: Commitment,
+    #[serde(default = "default_blob_index")]
+    pub index: i64,
```
zvolin, line 32: I think this information can only be known if we receive Blob from a full node from already existing EDS. -1 looks like a C-ish default. I'd prefer Option<u64> here (or u16/usize?) which will be None for blobs created locally and set for blobs received.
Additionallly could you add a check for that, here that locally created blobs have None and in rpc that the field is set?
This likely require #239 to be merged first but it should be there quickly

S1nus: Good point. I will look into a idiomatic way to use `Option` here.

zvolin: > Additionallly could you add a check for that, here that locally created blobs have None and in rpc that the field is set?

actually checknig the None in `celestia-types` sounds like a pointless thing, it'd be nice tho to use the index in `celestia-rpc` tests

PR #261, opened 2024-04-04
Add FifoStore

fl0rek:


fl0rek:
probably won't be needed, we can resurrect it as necessary.

PR #265, opened 2024-04-09
feat(wasm): Run Lumina in a Shared Worker

fl0rek:
This PR introduces several features:
- either [SharedWorker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker) or [Worker](https://developer.mozilla.org/en-US/docs/Web/API/Worker) is spawned for the lumina instance, so that we no longer run in main "GUI" thread.
- `NodeDriver` replaces WasmNode and is responsible for steering actual lumina node running inside a Worker. Multiple tabs from the same context can access the same lumina instance. After spawning the shared worker, NodeDriver establishes a request-response communication channel over JS primitives.
- Node inside the SharedWorker has server-like infrastructure. `onconnect` event means that a new connection happened and we're supposed to use `event.port[0]` to communicate with a newly connected `NodeDriver` by responding with results of the command we're sent.
- Node inside Worker has simplified architecture, since it doesn't need to handle new tabs connecting, but otherwise works the same


node-wasm/src/worker/channel.rs
```diff
@@ -0,0 +1,191 @@
...
+                    };
+
+                    if let Err(e) = local_tx.send(WorkerMessage::NewConnection(port)).await {
+                        error!("command channel inside worker closed, should not happen: {e}");
+                    }
+                })
+            });
+
+        let worker_scope = SharedWorker::worker_self();
+        worker_scope.set_onconnect(Some(onconnect_callback.as_ref().unchecked_ref()));
```
oblique, line 123: So, worker clients are only increasing? How do we know if a worker client got disconnected?

node-wasm/src/worker/channel.rs
```diff
@@ -0,0 +1,191 @@
...
+        let onmessage_callback = move |ev: MessageEvent| {
+            let local_tx = near_tx.clone();
+            spawn_local(async move {
+                let message_data = ev.data();
+
+                let data = from_value(message_data)
+                    .map_err(|e| {
+                        error!("could not convert from JsValue: {e}");
+                    })
+                    .ok();
```
oblique, line 39: I believe instead of `None` here, we should generate something like `WorkerError::InvalidResponse` and `WireMessage` be `Result<WorkerResponse, WorkerError>`.

node-wasm/src/worker/channel.rs
```diff
@@ -0,0 +1,191 @@
...
+        let client_message_callback: Closure<dyn Fn(MessageEvent)> =
+            Closure::new(move |ev: MessageEvent| {
+                let local_tx = near_tx.clone();
+                spawn_local(async move {
+                    let client_id = ClientId(client_id);
+                    let message_data = ev.data();
+                    let Ok(command) = from_value::<NodeCommand>(message_data) else {
+                        warn!("could not deserialize message from client {client_id}");
+                        return;
+                    };
```
oblique, line 153: We should still respond because client is waiting. Maybe we can do something like `.ok_or(NodeCommand::Invalid)` and then worker respond with something like `WokerError::InvalidCommand`.

node-wasm/src/utils.rs
```diff
@@ -103,3 +106,79 @@ where
...
+    fn worker_self() -> Self::GlobalScope {
+        JsValue::from(js_sys::global()).into()
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub(crate) enum JsResult<T, E>
+where
+    T: JsCast,
+    //E: Serialize + DeserializeOwned,
```
oblique, line 128:
```suggestion
```

node-wasm/src/node.rs
```diff
@@ -38,141 +37,226 @@ pub struct WasmNodeConfig {
...
+        let response = self.channel.exec(command).await?;
+        let result = response.into_connected().check_variant()?;
+
+        Ok(result?)
     }
 
     /// Wait until the node is connected to at least 1 trusted peer.
     pub async fn wait_connected_trusted(&self) -> Result<()> {
-        Ok(self.0.wait_connected_trusted().await?)
+        let command = NodeCommand::WaitConnected { trusted: false };
```
oblique, line 124:
```suggestion
        let command = NodeCommand::WaitConnected { trusted: true };
```

node/src/p2p.rs
```diff
@@ -506,7 +506,6 @@ impl P2p {
         Ok(rx.await?)
     }
 }
-
```
oblique, line 509: Revert this

node-wasm/src/utils.rs
```diff
@@ -103,3 +106,79 @@ where
...
+
+impl<T, E> Debug for JsResult<T, E>
+where
+    T: JsCast,
+    E: Serialize + DeserializeOwned,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            JsResult::Ok(_) => f.debug_tuple("JsResult::Ok"),
+            JsResult::Err(_) => f.debug_tuple("JsResult::Err"),
```
zvolin, line 180: is there any reason not to require just `T & E: Debug` and include actual filelds?

node-wasm/src/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+    fn from(error: MultiaddrError) -> Self {
+        WorkerError::BootstrapNodeMultiaddrInvalid(error.to_string())
+    }
+}
+
+impl From<serde_wasm_bindgen::Error> for WorkerError {
+    fn from(error: serde_wasm_bindgen::Error) -> Self {
+        WorkerError::SerdeError(error.to_string())
+    }
+}
```
zvolin, line 100:
```suggestion
}

```

node-wasm/src/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+struct NodeWorker {
+    node: Node<IndexedDbStore>,
+}
+
+fn to_jsvalue_or_undefined<T: Serialize>(value: &T) -> JsValue {
+    to_value(value).unwrap()
+}
```
zvolin, line 107: I don't understand how `.unwrap()` achieves `undefined` there

node-wasm/src/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+    async fn get_listeners(&mut self) -> Result<Vec<Multiaddr>> {
+        Ok(self.node.listeners().await?)
+    }
+
+    async fn wait_connected(&mut self, trusted: bool) -> Result<()> {
+        if trusted {
+            self.node.wait_connected().await?;
+        } else {
+            self.node.wait_connected_trusted().await?;
+        }
```
zvolin, line 155:
```suggestion
        if trusted {
            self.node.wait_connected_trusted().await?;
        } else {
            self.node.wait_connected().await?;
        }
```

node-wasm/src/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    info!("js shim: {script}");
+
+    let array = Array::new();
+    array.push(&script.into());
+    let blob = Blob::new_with_str_sequence_and_options(
+        &array,
+        BlobPropertyBag::new().type_("application/javascript"),
+    )
+    .map_err(|e| JsError::new(&format!("could not create SharedWorker Blob: {e:?}")))?;
```
zvolin, line 359: we could replace those `.map_err` with `.js_context` unless we want to remove it completely

zvolin (commented):
nice one, did some first pass, will likely get back to it

cli/static/worker.js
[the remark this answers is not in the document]
fl0rek: yes, removed 👍

node-wasm/src/utils.rs
fl0rek: I wanted to be conservative with the `Debug` requirement on contained types, but ultimately it looks like it's not gonna be a problem.
Changed to full debug print 👍

node-wasm/src/worker/channel.rs
fl0rek: Good point. I've added new worker message and handling of it at the `run_worker` level.

node-wasm/src/worker.rs
```diff
@@ -0,0 +1,372 @@
...
+            .map(|h| to_value(&h))
+            .collect::<Result<Array, _>>()?)
+    }
+
+    async fn get_last_seen_network_head(&mut self) -> JsValue {
+        self.node
+            .get_network_head_header()
+            .as_ref()
+            .map(to_jsvalue_or_undefined)
+            .unwrap_or(JsValue::UNDEFINED)
```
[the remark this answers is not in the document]
fl0rek: Right, fixed `to_jsvalue_or_undefined` and it's usage here.

node-wasm/src/worker.rs
fl0rek: Right, missed it 👍

node-wasm/src/worker.rs
fl0rek: Existing JsContext didn't exactly fit the need here, but as it's not used anyway I've modeled `JsValueToJsError` on it

node-wasm/src/worker/channel.rs
fl0rek: Good idea, reworked `WireMessage` to carry a `Result` with a `WorkerError::CouldNotDeserialiseResponse`.

node-wasm/src/node.rs
```diff
@@ -38,141 +37,226 @@ pub struct WasmNodeConfig {
...
+#[wasm_bindgen]
+impl NodeDriver {
+    /// Create a new connection to a Lumina node in a Shared Worker.
+    /// Note that single Shared Worker can be accessed from multiple tabs, so Lumina may already
+    /// be running be running, before `NodeDriver::start` call.
     #[wasm_bindgen(constructor)]
-    pub async fn new(config: WasmNodeConfig) -> Result<WasmNode> {
-        let config = config.into_node_config().await?;
+    pub async fn new() -> Result<NodeDriver> {
+        let worker = spawn_worker(LUMINA_SHARED_WORKER_NAME, "/wasm/lumina_node_wasm.js")?;
```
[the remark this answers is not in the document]
fl0rek: I'm not sure I follow - two above linked properties would allow us to get a URL of the page that the user is on, but what we need here is wasm entrypoint location.
Even though we're essentially loading the same wasm module as is running again, I don't see a way to get it's location programatically at runtime.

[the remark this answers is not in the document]
zvolin: if we cannot get this programatically, then I think we should accept it as an argument to avoid hardcode in library code

node-wasm/Cargo.toml
```diff
@@ -24,18 +24,27 @@ crate-type = ["cdylib", "rlib"]
...
+serde_repr = "0.1"
+thiserror = "1.0"
 time = { version = "0.3", features = ["wasm-bindgen"] }
+tokio =  { version = "*", features = ["sync"]}
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.18", features = ["time"] }
 tracing-web = "0.1.2"
 wasm-bindgen = "0.2.88"
 wasm-bindgen-futures = "0.4.37"
+web-sys = { version = "0.3.69", features = ["BroadcastChannel", "MessageEvent", "Worker", "WorkerOptions", "WorkerType", "SharedWorker", "MessagePort", "SharedWorkerGlobalScope", "Blob", "BlobPropertyBag", "Url"]}
```
zvolin, line 49:
```suggestion
web-sys = { version = "0.3.69", features = [
  "BroadcastChannel",
  "MessageEvent",
  "Worker",
  "WorkerOptions",
  "WorkerType",
  "SharedWorker",
  "MessagePort",
  "SharedWorkerGlobalScope",
  "Blob",
  "BlobPropertyBag",
  "Url",
] }
```

node-wasm/Cargo.toml
```diff
@@ -24,18 +24,27 @@ crate-type = ["cdylib", "rlib"]
...
+enum-as-inner = "0.6"
```
zvolin, line 50: please keep this sorted :pray:

node-wasm/src/node.rs
```diff
@@ -38,141 +37,219 @@ pub struct WasmNodeConfig {
...
+
+#[wasm_bindgen(js_class = NodeClient)]
+impl NodeDriver {
+    /// Create a new connection to a Lumina node running in a Shared Worker.
+    /// Note that single Shared Worker can be accessed from multiple tabs, so Lumina may
+    /// already have been started. Otherwise it needs to be started with [`NodeDriver::start`].
     #[wasm_bindgen(constructor)]
-    pub async fn new(config: WasmNodeConfig) -> Result<WasmNode> {
-        let config = config.into_node_config().await?;
+    pub async fn new() -> Result<NodeDriver> {
```
zvolin, line 53: since `SharedWorkers` aren't supported on chrome android, can we have `new_shared()` and `new_dedicated()` and somehow switch between the two?
https://issues.chromium.org/issues/40290702

node-wasm/src/node.rs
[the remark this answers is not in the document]
zvolin: switched to running a worker from a static script, since with blob url the url was unique for each lumina instance and resulted in many shared workers instead of a single. This is completely hidden from user, we get the worker path programatically

node-wasm/src/worker.rs
```diff
@@ -0,0 +1,328 @@
...
+                    .await
+                    .into(),
+            ),
+            NodeCommand::LastSeenNetworkHead => {
+                WorkerResponse::LastSeenNetworkHead(self.get_last_seen_network_head().await)
+            }
+            NodeCommand::GetSamplingMetadata { height } => {
+                WorkerResponse::SamplingMetadata(self.get_sampling_metadata(height).await)
+            }
+            NodeCommand::CloseWorker => {
```
zvolin, line 245: not sure where to put this, but re-opening lumina will result in worker never connecting to the bootstrap peer again. It only works the first time. The only thing that worked for me is reopening chromium

zvolin: I can see it trying to connect but doesn't get any response
image

zvolin: or rather we never manage to send the request? :thinking:

node-wasm/src/worker/channel.rs
```diff
@@ -0,0 +1,211 @@
...
+        self.send_response(client, Err(error))
+    }
+
+    pub fn add(&mut self, port: MessagePort) {
+        let client_id = self.clients.len();
+
+        let near_tx = self.command_channel.clone();
+        let client_message_callback: Closure<dyn Fn(MessageEvent)> =
+            Closure::new(move |ev: MessageEvent| {
+                let local_tx = near_tx.clone();
```
oblique, line 163: Rename `near_tx` and `local_tx` to `command_channel`

node-wasm/src/worker/channel.rs
```diff
@@ -0,0 +1,211 @@
...
+use web_sys::{MessageEvent, MessagePort, SharedWorker};
+
+use crate::utils::WorkerSelf;
+use crate::worker::commands::{NodeCommand, WorkerResponse};
+use crate::worker::WorkerError;
+
+type WireMessage = Result<WorkerResponse, WorkerError>;
+type WorkerClientConnection = (MessagePort, Closure<dyn Fn(MessageEvent)>);
+
+const WORKER_CHANNEL_SIZE: usize = 1;
```
oblique, line 18: Add a comment here why this doesn't need to be more than one.

fl0rek: Added comment 👍

node-wasm/src/worker.rs
fl0rek: I've not been able to drill down to the cause of this issue, but using dedicated workers [just added](https://github.com/eigerco/lumina/pull/265#discussion_r1603211525), seems to be a workaround to this issue.
I've added code to default to dedicated worker for chrome and with it I can't seem to trigger the issue.

node-wasm/js/worker.js
```diff
@@ -0,0 +1,29 @@
...
+export function worker_script_url() {
+  return import.meta.url;
+}
+
+// if we are in a worker
+if (typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope) {
+  Error.stackTraceLimit = 99;
+
+  // for SharedWorker we queue incoming connections
+  // for dedicated Workerwe queue incoming messages (coming from the single client)
```
zvolin, line 14:
```suggestion
  // for dedicated Worker we queue incoming messages (coming from the single client)
```

node-wasm/src/node.rs
```diff
@@ -38,141 +37,246 @@ pub struct WasmNodeConfig {
...
+            }
+        };
+
+        Ok(Self { client })
+    }
+
+    /// Check whether Lumina is currently running
+    pub async fn is_running(&self) -> Result<bool> {
+        let command = NodeCommand::IsRunning;
+        let response = self.client.exec(command).await?;
```
zvolin, line 100: shouldn't errors in this function mean the node is not running?

node-wasm/src/worker.rs
oblique: Why it doesn't fit the need? The error is not printed correctly with `Display`?

fl0rek: Reworked Worker errors to use newly introduced js error types

node-wasm/src/node.rs
fl0rek: That's bordering on philosophical question :P 
I wanted to distinguish situation of "Worker may be running, but definitely isn't responding" and "We can communicate with worker, but it's not running", since they have different recovery procedures - `start`ing the node and trying to re-create a worker.

node-wasm/src/worker/channel.rs
fl0rek: https://github.com/eigerco/lumina/issues/307

node-wasm/src/utils.rs
```diff
@@ -73,6 +79,134 @@ pub(crate) fn js_value_from_display<D: fmt::Display>(value: D) -> JsValue {
...
+
+    if let Some(user_agent) = user_agent {
+        user_agent
+            .as_deref()
+            .unwrap_or("")
+            .contains(CHROME_USER_AGENT_DETECTION_STR)
+    } else {
+        false
+    }
+}
```
oblique, line 208: It would be nice if the logic runs once and then return a cached result

node-wasm/src/worker/channel.rs
```diff
@@ -0,0 +1,347 @@
...
+    /// Response enum variant can be converted into appropriate type at runtime with a provided
+    /// [`CheckableResponseExt`] helper.
+    ///
+    /// [`CheckableResponseExt`]: crate::utils::CheckableResponseExt
+    pub async fn exec(&self, command: NodeCommand) -> Result<WorkerResponse, WorkerError> {
+        let mut response_channel = self.response_channel.lock().await;
+        self.send(command)
+            .map_err(WorkerError::WorkerCommunicationError)?;
+
+        let message: WireMessage = response_channel.recv().await.ok_or_else(|| {
```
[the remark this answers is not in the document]
fl0rek: You're right, since `WorkerClient` is alive, channel must be alive at this point. 👍

node-wasm/src/utils.rs
```diff
@@ -188,23 +163,14 @@ const CHROME_USER_AGENT_DETECTION_STR: &str = "Chrome/";
...
-            .unwrap_or("")
-            .contains(CHROME_USER_AGENT_DETECTION_STR)
-    } else {
-        false
-    }
+pub(crate) fn is_chrome() -> Result<bool, Error> {
+    js_sys::Reflect::get(&js_sys::global(), &JsValue::from_str("navigator"))
+        .context("failed to get `navigator` from global object")?
+        .dyn_into::<web_sys::Navigator>()
+        .context("`navigator` is not instanceof `Navigator`")?
```
oblique, line 170: Move this to `get_navigator`

node-wasm/src/utils.rs
```diff
@@ -73,6 +79,134 @@ pub(crate) fn js_value_from_display<D: fmt::Display>(value: D) -> JsValue {
...
+        JsValue::from(js_sys::global()).into()
+    }
+
+    fn is_worker_type() -> bool {
+        js_sys::global().has_type::<Self::GlobalScope>()
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub(crate) enum JsResult<T, E>
```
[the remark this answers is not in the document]
fl0rek: Added comment with explanation why `JsResult` was added 👍

node-wasm/src/utils.rs
fl0rek: True, but do we actually need it? Currently we do detection once and fingers crossed this doesn't increase. 
We could bring in `memoize` crate or do caching manually when/if this is called multiple times.

oblique (approved):

zvolin (approved):
:rocket:

PR #269, opened 2024-04-19
feat(node/daser)!: Implement backward sampling and sampling window

oblique:


node/src/daser.rs
```diff
@@ -8,25 +8,27 @@
...
-use crate::p2p::shwap::convert_cid;
-use crate::p2p::{P2p, P2pError};
-use crate::store::{Store, StoreError};
+use crate::p2p::shwap::sample_cid;
+use crate::p2p::{P2p, P2pError, GET_SAMPLE_TIMEOUT};
+use crate::store::{SamplingStatus, Store, StoreError};
+use crate::utils::FusedReusableBoxFuture;
 
 const MAX_SAMPLES_NEEDED: usize = 16;
+const SAMPLING_WINDOW: Duration = Duration::from_secs(14678036);
```
fl0rek, line 31: If my calculation is correct, that's ~169.88 days. Where's that coming from? I thought sampling window was supposed to be 30 days.
Also could we do something like `Duration::from_secs($days * 24 * 60 * 60)`, so that it's more obvious what the value is?

node/src/daser.rs
```diff
@@ -110,74 +122,295 @@ where
...
             }
         }
+    }
+
+    async fn connected_event_loop(&mut self) -> Result<()> {
+        debug!("Entering connected_event_loop");
+
+        let mut peer_tracker_info_watcher = self.p2p.peer_tracker_info_watcher();
+
+        // Check if connection status changed before creating the watcher
```
fl0rek, line 179:
```suggestion
        // Check if connection status changed before the watcher was created
```

node/src/daser.rs
```diff
@@ -110,74 +122,295 @@ where
...
+                }
+                _ = peer_tracker_info_watcher.changed() => {
+                    if peer_tracker_info_watcher.borrow().num_connected_peers == 0 {
+                        warn!("All peers disconnected");
+                        break;
+                    }
+                }
+                res = &mut self.sampling_fut => {
+                    // Beetswap returns only fatal errors that are not related
+                    // to P2P nor networking.
```
fl0rek, line 205:
```suggestion
                    // Beetswap only returns fatal errors that are not related
                    // to P2P nor networking.
```

node/src/daser.rs
oblique: Good catch. It is a leftover from some manual tests I was doing.

node/src/daser.rs
```diff
@@ -110,74 +122,295 @@ where
...
+    async fn schedule_next_sample_block(&mut self) -> Result<()> {
+        assert!(self.sampling_fut.is_terminated());
+
+        let Some(args) = self.queue.pop_front() else {
+            return Ok(());
+        };
+
+        // Make sure that block is still in the sampling window.
+        if !in_sampling_window(args.time) {
+            // Queue is sorted, so as soon as we reach a block
```
fl0rek, line 240:
```suggestion
            // Queue is sorted by block height in descending order, so as soon as we reach a block
```

node/src/daser.rs
```diff
@@ -110,74 +122,295 @@ where
...
+            // rest wouldn't be either.
+            self.queue.clear();
+            return Ok(());
+        }
+
+        let mut cids = Vec::new();
 
-        let header = self.store.get_by_height(height).await?;
-        let (sampled_cids, accepted) = self.sample_block(&header).await?;
+        // Select random shares to be sampled and generate their CID.
```
fl0rek, line 249:
```suggestion
        // Select random shares to be sampled and generate their CIDs.
```

node/src/daser.rs
```diff
@@ -110,74 +122,295 @@ where
...
+        for (row, col) in random_indexes(args.square_width, self.max_samples_needed) {
+            let cid = sample_cid(row, col, args.height)?;
+            cids.push(cid);
+        }
 
+        // Update CID list before we start the sampling procedure, otherwise there are
+        // some edge cases that can leak the CIDs and never cleaned from blockstore.
```
fl0rek, line 258:
```suggestion
        // Update the CID list before we start sampling, otherwise it's possible for us to leak CIDs
        // causing associated blocks to never get cleaned from blockstore
```

node/src/daser.rs
```diff
@@ -110,74 +125,295 @@ where
...
+            }
+            None => {
+                // Reverse iterate heights
+                for height in (1..=new_head).rev() {
+                    if is_sampled(&*self.store, height).await {
+                        continue;
+                    }
+
+                    let Ok(header) = self.store.get_by_height(height).await else {
+                        // We reached the tail of the known heights
```
fl0rek, line 348: Mostly to confirm, 
> TODO: Adjust algorithm for backward syncing

above is to account for upcoming backwards header sync changes, that make it so that we're no longer guaranteed to store single continuous block of headers in store?

node/src/daser.rs
```diff
@@ -110,74 +125,295 @@ where
...
+                        break;
+                    };
+
+                    if !in_sampling_window(header.time()) {
+                        // We reached the tail of the sampling window
```
fl0rek, line 353:
```suggestion
                        // We've reached the tail of the sampling window
```

node/src/daser.rs
```diff
@@ -110,74 +125,295 @@ where
...
+    }
+
+    let Ok(age) = now.duration_since(time) else {
+        return false;
+    };
+
+    age <= SAMPLING_WINDOW
+}
+
+/// Returns true is block is sampled (i.e. it was accepted or rejected).
```
fl0rek, line 408:
```suggestion
/// Returns true if block has been sampled (i.e. it was accepted or rejected).
```

node/src/store/redb_store.rs
```diff
@@ -369,6 +347,25 @@ impl Store for RedbStore {
         self.get_by_height(height).await
     }
 
+    async fn wait_new_head(&self) -> u64 {
```
fl0rek, line 350: With this fn added, would it make sense to express `wait_height` with a common code in `Store` using it? Seems like an opportunity to reduce code duplication.

oblique: With forward syncing we could do that, however an edge case exists in backward syncing:

When the store is empty, we don't know the network head, so we don't know which height we should await for. Of course we can workaround this with a loop of `head_height` and a sleep, until head is initialized. I'm ok with both approaches.

node/src/daser.rs
oblique: Yes. Now the algorithm iterates from head to block 1. This is not possible with backward syncing.

node/src/daser.rs
```diff
@@ -110,74 +124,295 @@ where
...
-        let indexes = random_indexes(header.dah.square_width(), self.max_samples_needed);
-        let mut futs = FuturesUnordered::new();
+    /// Rollback to the highest accepted block.
+    ///
+    /// Rollback is needed because of how Shwap works: The only way to know
+    /// if sampling has failed is via a timeout on bitswap requests and this
+    /// can be generated from network issues too. We rollback just to recheck
+    /// the previous failed samplings.
+    async fn rollback(&mut self) {
+        // TODO
```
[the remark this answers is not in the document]
oblique: After some private discussion we took different approach than rollback: We resample all rejected blocks after disconnection.

node/src/daser.rs
```diff
@@ -69,7 +74,7 @@ impl Daser {
 
         spawn(async move {
             if let Err(e) = worker.run().await {
-                error!("Fatal DASer error: {e}");
+                error!("Daser stopped because of fatal error: {e}");
```
fl0rek, line 77:
```suggestion
                error!("Daser stopped because of a fatal error: {e}");
```

node/src/daser.rs
```diff
@@ -110,74 +125,290 @@ where
...
+    async fn schedule_next_sample_block(&mut self) -> Result<()> {
+        assert!(self.sampling_fut.is_terminated());
+
+        let Some(args) = self.queue.pop_front() else {
+            return Ok(());
+        };
 
-        let header = self.store.get_by_height(height).await?;
-        let (sampled_cids, accepted) = self.sample_block(&header).await?;
+        // Make sure that block is still in the sampling window.
```
fl0rek, line 240:
```suggestion
        // Make sure that the block is still in the sampling window.
```

node/src/daser.rs
```diff
@@ -110,74 +125,290 @@ where
...
-                // as the data of the CID. Because of that, the only signal
-                // that data sampling verification failed is query timing out.
-                Err(P2pError::BitswapQueryTimeout) => accepted = false,
-                Err(e) => return Err(e.into()),
+    /// Add to the queue the blocks that need to be sampled.
+    ///
+    /// NOTE: We resampled rejected blocks, because rejection can happen
+    /// because of edge-cases, such us network issues. This is a limitation
+    /// because of how Shwap works: The only way to know if sampling has failed
+    /// is via a timeout.
```
fl0rek, line 307:
```suggestion
    /// NOTE: We resample rejected blocks, because rejection can happen
    /// in some unrelated edge-cases, such us network issues. This is a Shwap 
    /// limitation that's coming from bitswap: only way for us to know if sampling
    /// failed is via timeout.
```

node/src/daser.rs
```diff
@@ -110,74 +125,290 @@ where
...
 
+/// Returns true if block has been sampled and accepted.
+async fn is_block_accepted(store: &impl Store, height: u64) -> bool {
+    match store.get_sampling_metadata(height).await {
+        Ok(Some(metadata)) => metadata.status == SamplingStatus::Accepted,
+        _ => false,
+    }
+}
+
+/// Returns unique and random indexes that will be used for the sampling.
```
fl0rek, line 411:
```suggestion
/// Returns unique and random indexes that will be used for sampling.
```

fl0rek (approved):
last few grammar nits, otherwise LGTM! 🚀

node/src/daser.rs
```diff
@@ -1,12 +1,33 @@
...
+//! Steps of how data sampling works:
 //!
-//! [`Sample`]: celestia_types::sample::Sample
+//! 1. Daser waits for at least one peer to connect.
+//! 2. Daser gets the local head height from the [`Store`] and iterates in reverse order
+//!    all the stored blocks.
+//!    - If a block is not sampled or if it was rejected, Daser queue it for sampling.
+//!      The rejected blocks are resampled because their rejection could be due unrelated
+//!      edge-cases than sampling, such us network issues.
+//!    - If a block is not within sampling window, it is not queued.
```
fl0rek, line 12:
```suggestion
//!    - If a block is not within the sampling window, it is not queued.
```

node/src/daser.rs
```diff
@@ -1,12 +1,33 @@
 //! Component responsible for data availability sampling of the already synchronized block
 //! headers announced on the Celestia network.
 //!
-//! When a new header is inserted into the [`Store`], [`Daser`] gets notified. It then fetches
-//! random [`Sample`]s of the block via Shwap protocol and verifies them. If all the samples
-//! get verified successfuly, then block is marked as accepted. Otherwise, if [`Daser`] doesn't
-//! receive valid samples, block is marked as not accepted and data sampling continues.
+//! Steps of how data sampling works:
```
fl0rek, line 4:
```suggestion
//! Sampling procedure comprises the following steps:
```

node/src/daser.rs
```diff
@@ -1,12 +1,33 @@
...
+//!      is needed for pruning them later on. This is done before retrival of CIDs is started because
+//!      a user can stop an ongoing sampling (e.g. with ctrl+c) but Bitswap already retrieved some of
+//!      the CIDs, and we don't want to lose this information.
+//!    - Initiates Bitswap retrival requests for the specified CIDs.
+//!    - If all CIDs are received, then block is considered sampled and accepted.
+//!    - If we reach a timeout of 10 seconds and at least one of the CIDs is not received, then
+//!      block is considered sampled and rejected.
+//!    - [`Store`] is updated with the sampling result.
+//! 5. Steps 3 and 4 run concurrently in a loop.
+//! 6. If all peers disconnect, Daser cleans the queue and moves to back to step 1.
```
fl0rek, line 30:
```suggestion
//! 3a. As new headers become available in the [`Store`], Daser adds them to the queue if
//!    they are within the sampling window.
//! 3b. If there isn't any ongoing sampling, Daser pops from the queue the next block to sample
//!     and initiates the following procedure:
//!    1. It makes sure that the block is still within the sampling window.
//!    2. It selects which random shares are going to be sampled and generates their Shwap CIDs.
//!    3. It updates [`Store`] with the CIDs that are going to be sampled. Tracking of the the CIDs
//!      is needed for pruning them later on. This is done before retrival of CIDs is started because otherwise
//!      user could stop the node after Bitswap stores the block in the blockstore, but before we can 
//!      record that in the [`Store`], causing a leak.
//!    4. Initiates Bitswap retrival requests for the specified CIDs.
//!    5. If all CIDs are received, then the block is considered sampled and accepted.
//!    6. If we reach a timeout of 10 seconds and at least one of the CIDs is not received, then
//!      block is considered sampled and rejected.
//!    7. [`Store`] is updated with the sampling result.
//! 4. Steps 3a and 3b are repeated concurently, unless we detect that all peers have disconnected. At that point daser cleans the queue and moves back to step 1.
```

PR #274, opened 2024-05-06
feat!(types): Add Blob::index field introduced in celestia 0.13

S1nus:
Another attempt at adding this field. Now, trying to use an `Option<u64>`.

### DIFF
--- a/types/src/blob.rs
+++ b/types/src/blob.rs
@@ -52,6 +52,10 @@ pub struct Blob {
     pub share_version: u8,
     /// A [`Commitment`] computed from the [`Blob`]s data.
     pub commitment: Commitment,
+    /// Index of the blob's first share in the EDS. Only set for blobs retrieved from chain.
+    // note: celestia supports deserializing blobs without index, so we should too
+    #[serde(default, with = "index_serde")]
+    pub index: Option<u64>,
 }
 
 impl Blob {
@@ -76,7 +80,8 @@ impl Blob {
     ///       "namespace": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQIDBAU=",
     ///       "data": "c29tZSBkYXRhIHRvIHN0b3JlIG9uIGJsb2NrY2hhaW4=",
     ///       "share_version": 0,
-    ///       "commitment": "m0A4feU6Fqd5Zy9td3M7lntG8A3PKqe6YdugmAsWz28="
+    ///       "commitment": "m0A4feU6Fqd5Zy9td3M7lntG8A3PKqe6YdugmAsWz28=",
+    ///       "index": -1
     ///     }"#},
     /// );
     /// ```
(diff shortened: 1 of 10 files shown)


types/src/blob.rs
```diff
@@ -52,6 +52,20 @@ pub struct Blob {
...
     pub commitment: Commitment,
+    #[serde(deserialize_with = "deserialize_optional_positive_integer")]
+    pub index: Option<u64>,
+}
+
+fn deserialize_optional_positive_integer<'de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value: i64 = Deserialize::deserialize(deserializer)?;
```
zvolin, line 63: nit

```suggestion
    let value = i64::deserialize(deserializer)?;
```

types/src/blob.rs
```diff
@@ -52,6 +52,20 @@ pub struct Blob {
     pub share_version: u8,
     /// A [`Commitment`] computed from the [`Blob`]s data.
     pub commitment: Commitment,
+    #[serde(deserialize_with = "deserialize_optional_positive_integer")]
+    pub index: Option<u64>,
+}
+
+fn deserialize_optional_positive_integer<'de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
```
zvolin, line 59: nit, could you move it to the very bottom just before `mod tests`?

types/src/blob.rs
```diff
@@ -52,6 +52,20 @@ pub struct Blob {
     pub share_version: u8,
     /// A [`Commitment`] computed from the [`Blob`]s data.
     pub commitment: Commitment,
+    #[serde(deserialize_with = "deserialize_optional_positive_integer")]
```
zvolin, line 55:
```suggestion
    /// Index of the blob's first share in the EDS. Only set for blobs retrieved from chain.
    #[serde(deserialize_with = "deserialize_optional_positive_integer")]
```

zvolin (approved):
Looks good :+1: left some suggestions

zvolin:
could you look into the CI failures too?

zvolin:
some of the failures seems to be coming from toolchain updates, I'll look into that tomorrow

types/src/blob.rs
zvolin: @S1nus this one seems to be overwritten by later commits. I can't spot this change in diff

fl0rek (approved):
looks good, thanks for the contribution! 🎸

fl0rek:
@S1nus I've just noticed that the commits aren't signed
Would you mind [setting up commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and force-pushing? Checks won't allow merge unless all commits are signed.

S1nus:
@fl0rek I set up signing, newer commits are signed. is it good? or need to do something else?

zvolin:
Previous commits needed to be signed too, but I just squashed them and it looks fine. I've added the fix for celestia-rpc tests, hopefully CI will be happy now

fl0rek (approved):

PR #276, opened 2024-05-14
feat(node): Generate events for data sampling that can be used by front-end

oblique:


zvolin:
> Events are not exposed to node-wasm yet. Should I do it now or after https://github.com/eigerco/lumina/pull/265?

there is high chance that the approach won't change after 265, so I'd say you can start already

node/src/daser.rs
```diff
@@ -548,32 +592,65 @@ mod tests {
...
 
         // Sample and reject block 19
-        handle_get_shwap_cid(&mut handle, &store, 19, &edses[18], true).await;
+        handle_get_shwap_cid(&mut handle, &store, &mut event_sub, 19, &edses[18], true).await;
 
         // Simulate disconnection
         handle.announce_all_peers_disconnected();
+
+        // Daser may scheduled Block 18 already, so we need to reply to that requests.
+        // For the sake of the test we reply with a bitswap timeout.
```
zvolin, line 607: actually I think the bitswap timeout is what would happen anyway if all peers disconnected :+1:

node/src/events.rs
```diff
@@ -0,0 +1,256 @@
...
+/// [`Node`]: crate::node::Node
+#[derive(Debug)]
+pub struct EventSubscriber {
+    rx: broadcast::Receiver<NodeEventInfo>,
+}
+
+impl EventChannel {
+    /// Create a new `EventSubscription`.
+    pub fn new() -> EventChannel {
+        let (tx, _) = broadcast::channel(32);
```
zvolin, line 55: I think 32 can quickly become too little, but no idea what would be the best value there, 1024 maybe? :sweat_smile:

node/src/events.rs
```diff
@@ -0,0 +1,256 @@
...
+                    continue;
+                }
+                Err(broadcast::error::RecvError::Closed) => return Err(RecvError::Closed),
+            }
+        }
+    }
+
+    /// Attempts to receive an already queued event from [`Node`] without awaiting.
+    ///
+    /// If no events queued, `Err(TryRecvError::Empty)` is returned.
```
fl0rek, line 119:
```suggestion
    /// If no events are queued, `Err(TryRecvError::Empty)` is returned.
```

node/src/events.rs
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+fn serialize_system_time<S>(value: &SystemTime, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::ser::Serializer,
+{
+    // Javascript express time as f64 and in milliseconds.
```
fl0rek, line 249:
```suggestion
    // Javascript expresses time as f64 and in milliseconds.
```

fl0rek (commented):
Nice one! Added few nits in comments

node/src/events.rs
oblique: changed to 1024

fl0rek (approved):
LGTM!
(docs and unused-deps are failing in `time` on latest nightly https://github.com/rust-lang/rust/issues/125319)

fl0rek:
Merged with bypassing two failing checks, @oblique checked manually that they're passing on previous nightly rust.

PR #279, opened 2024-05-15
feat(node/syncer)!: Implement backwards header sync

fl0rek:
## Store
- Relaxed rules around header insertion. Previously `Store` enforced storing a single contiguous range from genesis to store head, with appends allowed in a single slot right after the head. 
Now `Store` holds possibly multiple non-overlapping header ranges starting and ending at arbitrary heights. Append is allowed as long as it extends an existing range, or creates a new head range (creating a new range at the front is needed for syncing from the front).
- Fundamental unit for inserting headers is now a Vec of headers and single header insert is expressed in terms of that (previously it was other way around).

## Header sync
1. Syncer starts up, queries store for headers already present and swarm for the network head
2. Using above, it calculates header ranges which are already published, but absent from store
3. Taking 512 highest missing headers it starts new header session
4. Session starts 8 jobs, each taking up to 64 contiguous heights from the missing list and requesting them from the network 
Each time job finishes its work, next up to 64 headers are requested. If less than 64 headers are received, remaining range is re-requested again. 
5. After fetching all the requested ranges, they are sorted into contiguous ranges. Each range is verified internally.
6. Each range is inserted into the store, veryfing the range edges against headers in store, if they exist. (If they don't exist yet, syncer will fetch them eventually, at which point this they'll be verified).


node/src/store/utils.rs
```diff
@@ -0,0 +1,548 @@
...
+}
+
+impl Iterator for HeaderRangesIterator {
+    type Item = u64;
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(v) = self.inner_iter.as_mut()?.next() {
+            return Some(v);
+        }
+        self.inner_iter = self.outer_iter.next();
+        self.next()
```
oblique, line 164: I prefer loop instead of recursion in this case.

```suggestion
        loop {
            if let Some(v) = self.inner_iter.as_mut()?.next() {
                return Some(v);
            }
            self.inner_iter = self.outer_iter.next();
        }
```

node/src/store/utils.rs
```diff
@@ -0,0 +1,548 @@
...
+#[derive(Debug, Clone, PartialEq, Default, Serialize)]
+pub struct HeaderRanges(SmallVec<[RangeInclusive<u64>; 2]>);
+
+#[derive(Debug, Error)]
+pub enum HeaderRangeError {
+    #[error("Overlapping range: {0}, {1}")]
+    RangeOverlap(u64, u64),
+}
+
+impl HeaderRanges {
```
oblique, line 37: Move anything releated to `HeaderRanges`/`HeaderRangesIterator` in a `header_ranges.rs` file.

node/src/store/utils.rs
```diff
@@ -0,0 +1,548 @@
...
+    pub fn validate(&self) -> Result<(), HeaderRangeError> {
```
oblique, line 38: This assumes that `self.0` is sorted, but I don't see any sorting being done.

node/src/p2p/header_session.rs
```diff
@@ -137,10 +135,38 @@ impl HeaderSession {
     }
 }
 
+fn sort_header_spans_into_ranges(
```
oblique, line 138: Add a bit of documentation here. For examples that this input `[1, 2], [7, 8], [3, 4]` will give this result `[1, 2, 3, 4], [7, 8]`.

Also what do you think about this name? `sort_and_flatten_headers`

node/src/p2p/header_session.rs
```diff
@@ -1,52 +1,51 @@
...
     response_rx: mpsc::Receiver<(u64, u64, Result<Vec<ExtendedHeader>>)>,
     ongoing: usize,
 }
 
 impl HeaderSession {
-    pub(crate) fn new(from_height: u64, amount: u64, cmd_tx: mpsc::Sender<P2pCmd>) -> Result<Self> {
-        if from_height < 1 || amount < 1 {
-            return Err(P2pError::HeaderEx(HeaderExError::InvalidRequest));
-        }
+    pub(crate) fn new(ranges: HeaderRanges, cmd_tx: mpsc::Sender<P2pCmd>) -> Result<Self> {
```
oblique, line 26: Add a bit of documentation on how `HeaderRanges` are expected to be constructed. An example of an input and a possible output (example of output: `[1, 2, 3, 4], [7, 8]`)

node/src/store/utils.rs
```diff
@@ -0,0 +1,548 @@
...
+
+pub(crate) trait RangeLengthExt {
+    fn len(&self) -> u64;
+}
+
+impl RangeLengthExt for RangeInclusive<u64> {
+    fn len(&self) -> u64 {
+        self.end() - self.start() + 1
+    }
+}
```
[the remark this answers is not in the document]
fl0rek: In essence yes, but not exactly - `count` consumes the iterator, since it gets the count by repeatedly calling `next` on it. For `RangeInclusive`, the bounds are known, so we can get the number of elements with maths and avoid clone.
Initially I was hoping to use [ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html), but that's [not implemented for `u64`](https://github.com/rust-lang/rust/issues/36386).

node/src/store/utils.rs
fl0rek: split out the header ranges stuff into `store/header_ranges.rs`

node/src/store/utils.rs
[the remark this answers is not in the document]
zvolin: `count` can be a footgun because it can overflow... but here we can underflow, could you switch to `saturating_sub()`?

node/src/store.rs
```diff
@@ -913,19 +940,124 @@ mod tests {
...
+        store.append_single(fork.next()).await.unwrap_err();
+    }
+
     /// Fills an empty store
     async fn fill_store<S: Store>(store: &mut S, amount: u64) -> ExtendedHeaderGenerator {
         assert!(!store.has_at(1).await, "Store is not empty");
 
         let mut gen = ExtendedHeaderGenerator::new();
-        let headers = gen.next_many(amount);
+        //let headers = gen.next_many(amount);
```
oblique, line 1055:
```suggestion
```

node/src/store/in_memory_store.rs
```diff
@@ -34,65 +42,70 @@ impl InMemoryStore {
...
+            return Ok(());
+        };
 
-        // A light check before checking the whole map
-        if head_height > 0 && height <= head_height {
-            return Err(StoreError::HeightExists(height));
-        }
+        let headers_range = head.height().value()..=tail.height().value();
+        println!("I: {headers_range:?}");
+        let neighbours_exist = self.try_insert_to_range(headers_range).await?;
```
oblique, line 71: Because of this we have a race with `insert` and `contains_height`. To avoid the race, we need to keep the lock until the end, which makes `DashMap` unneeded.

Should we drop `DashMap` overall and have a single `RwLock` for the whole thing?

node/src/store/utils.rs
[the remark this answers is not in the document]
fl0rek: good call, switched to `saturating_sub`

node/src/p2p/header_session.rs
fl0rek: I like the suggestion more than the current version, changed :P 
Touch worried about the canonical interpretation of `flatten` as something that removes one level of array nesting, but I think we're close enough.

Added few words of documentation

node/src/p2p/header_session.rs
fl0rek: Added documentation with links to relevant functions and examples of parameters and responses.

node/src/store/in_memory_store.rs
fl0rek: You're right, the lock scope change during refactor.

Switched to having the header part (`headers`, `height_to_hash` and `stored_ranges`) in `InMemoryStoreInner` and behind RwLock. I've left sampling data as is - due to append-only nature of the store I think we're fine relying on just `DashMap` there.

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,455 @@
...
+        let mut found_range = loop {
+            let Some((idx, stored_range)) = stored_ranges_iter.next() else {
+                return Err(StoreError::InsertPlacementDisallowed(
+                    *to_insert.start(),
+                    *to_insert.end(),
+                ));
+            };
+
+            // TODO: break early
+            //if sth
```
oblique, line 115: maybe you wanted to do something here

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,455 @@
...
+            // Empty store case
+            return Ok(RangeScanResult {
+                range_index: 0,
+                range: to_insert.clone(),
+                range_to_remove: None,
+            });
+        };
+
+        // allow inserting a new header range in front of the current head range
+        if *to_insert.start() > head_range.end() + 1 {
```
oblique, line 97: This is a bit confusing. It took me some time to understand the purpose of `+ 1`. As I understand if the last stored range is `10..=11` you want to be able to add `13..=14` but not `12..=14`? Is it because if they are continues you want to merge them into `10..=14`?.

I think you need to add in the comment that a gap needs to be present.

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,455 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "[")?;
+        for (idx, range) in self.0.iter().enumerate() {
+            if idx == 0 {
+                write!(f, "{}..={}", range.start(), range.end())?;
+            } else {
+                write!(f, ", {}..={}", range.start(), range.end())?;
+            }
+        }
+        write!(f, "]")
```
oblique, line 182: I prefer something like `[10-11, 14-20]`.

fl0rek: Changed from rust-like notation, to more human recognisable `[1-2, 10-20]`

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,455 @@
...
+
+pub type HeaderRange = RangeInclusive<u64>;
+
+pub(crate) trait RangeLengthExt {
+    fn len(&self) -> u64;
+}
+
+impl RangeLengthExt for RangeInclusive<u64> {
+    fn len(&self) -> u64 {
+        self.end().saturating_sub(*self.start()) + 1
```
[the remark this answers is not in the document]
fl0rek: Thanks! missed that. Fixed and added unittests for `len`.

node/src/syncer.rs
```diff
@@ -450,82 +445,95 @@ where
...
-            warn!("Failed to store batch {start} until {end}: {e}");
+        for headers in headers_spans {
+            let first_header = headers.first().unwrap();
+            let last_header = headers.last().unwrap();
+
+            info!(
+                "on_fetch_next_batch_result received headers range: {}..={}",
+                first_header.height().value(),
+                last_header.height().value()
+            );
```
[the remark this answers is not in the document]
fl0rek: removed 👍

node/src/syncer.rs
```diff
@@ -305,41 +301,41 @@ where
...
 
         let ongoing_batch = self
             .ongoing_batch
             .as_ref()
-            .map(|ongoing| format!("[{}, {}]", ongoing.start, ongoing.end))
+            .map(|ongoing| format!("{}", ongoing.fetch_ranges))
             .unwrap_or_else(|| "None".to_string());
 
-        info!("syncing: {local_head}/{subjective_head}, ongoing batch: {ongoing_batch}",);
+        info!("syncing: {stored_headers}/{subjective_head}, ongoing batches: {ongoing_batch}",);
```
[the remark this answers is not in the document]
fl0rek: Good idea, `[range]/height` looked a bit weird in retrospect.  Updated 👍

node/src/store/in_memory_store.rs
```diff
@@ -1,199 +1,306 @@
...
+            return Ok(());
+        };
+
+        let headers_range = head.height().value()..=tail.height().value();
+        let range_scan_result = self.stored_ranges.check_range_insert(&headers_range)?;
+
+        if verify_neighbours {
+            let prev_exists = headers_range.start() != range_scan_result.range.start();
+            let next_exists = headers_range.end() != range_scan_result.range.end();
+            // header range is already internally verified against itself in `P2p::get_unverified_header_ranges`
```
oblique, line 202: That is true, however this is a public API and we shouldn't trust the input. So we have 2 options:

1. Add `verify_headers` flag.
2. Make `insert` a private API.

This issue exists in other stores too.

In case of option 1, I prefer something like:

```rust
enum Verify {
    No,
    Headers,
    Neightbours,
    HeadersAndNeightbours,
}
```

node/src/store/in_memory_store.rs
```diff
@@ -1,199 +1,306 @@
...
+            self.verify_against_neighbours(head, tail, (prev_exists, next_exists))?;
+        } else {
+            verify_range_contiguous(&headers)?;
```
oblique, line 205: Shouldn't we do this regardless? This issue exists in other stores too.

node/src/store/redb_store.rs
```diff
@@ -316,21 +321,28 @@ impl RedbStore {
     }
 
     async fn get_sampling_metadata(&self, height: u64) -> Result<Option<SamplingMetadata>> {
+        if !self.contains_height(height).await {
+            return Err(StoreError::NotFound);
+        }
+
```
oblique, line 327: This opens a read TX and later we open another one. Both should be done in the same read TX.

node/src/store/utils.rs
fl0rek: '

The way ranges are created, all created `HeaderRanges` should be sorted. Once `HeaderRanges` creation is make private, we should be good. I think I could add `debug_assert`s with checks at create time.'

node/src/node.rs
```diff
@@ -279,6 +279,11 @@ where
         self.p2p.header_sub_watcher().borrow().clone()
     }
 
+    /// Get ranges of headers currently stored.
```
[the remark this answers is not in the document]
fl0rek: No strong opinions here, I think we can hide it 👍

node/src/store/header_ranges.rs
fl0rek: This has no impact on which inserts are valid, but is mearly here to make sure that in case the ranges are contiguous, they merge into one and we don't create slightly pathological ranges like [1..=2, 3..=5], since they have simpler representation.
Added comment about it.

node/src/store/in_memory_store.rs
```diff
@@ -245,28 +352,50 @@ impl Store for InMemoryStore {
...
 
-impl Clone for InMemoryStore {
-    fn clone(&self) -> Self {
+impl From<Vec<ExtendedHeader>> for InMemoryStore {
+    fn from(hs: Vec<ExtendedHeader>) -> Self {
+        let range = match (hs.first(), hs.last()) {
+            (Some(head), Some(tail)) => [head.height().value()..=tail.height().value()].into(),
+            (None, None) => HeaderRanges::default(),
+            _ => unreachable!(),
+        };
```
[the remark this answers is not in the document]
fl0rek: I've put it as a workaround for disappeared `Clone` (since we have async lock inside now). After small refactor I'll remove it.

node/src/store/redb_store.rs
fl0rek: Thought I could get away with it, since we're add-only :P, but you're right doing it in one txn would be better. Fixed 👍

node/src/store/in_memory_store.rs
fl0rek: I've put it here mostly to have a similar API to what we had previously with `append_single`/`append_single_unchecked`. If we don't want to provide unchecked part, and enforce checks, I think we could remove it entirely.

node/src/store/in_memory_store.rs
fl0rek: Skipping this check here relied on it being run when validating the headers range. But similar to above, if we want to provide single "checked", interface this can go.

node/src/lib.rs
```diff
@@ -1,6 +1,18 @@
...
+        {
+            HeaderRanges::from_vec(smallvec::smallvec![
+            $(
+                $r,
+            )*
+            ])
+        }
+    };
+}
+
```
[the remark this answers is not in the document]
fl0rek: Nice, I didn't know about that one 👍

node/src/store/indexed_db_store.rs
```diff
@@ -67,11 +71,28 @@ impl IndexedDbStore {
...
+                    .transaction(&[RANGES_STORE_NAME], TransactionMode::ReadWrite)?;
+                let ranges_store = tx.store(RANGES_STORE_NAME)?;
+                let jsvalue_range = to_value(&(1, head.height().value()))?;
+                let jsvalue_key = to_value(&0)?;
+
+                ranges_store.put(&jsvalue_range, Some(&jsvalue_key)).await?;
+
+                tx.commit().await?;
+            }
+        }
```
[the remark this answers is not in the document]
fl0rek: Yes, as far as I understand it, rexie will do a schema upgrade in the background, preserving the data if possible. I've checked and this correctly converts head height to current range format on Chrome and Firefox.

node/src/store/indexed_db_store.rs
```diff
@@ -13,14 +13,17 @@ use serde::{Deserialize, Serialize};
 use serde_wasm_bindgen::{from_value, to_value};
 use tokio::sync::Notify;
 
+use crate::store::header_ranges::{HeaderRange, HeaderRanges, HeaderRangesExt};
+use crate::store::utils::{verify_range_contiguous, RangeScanResult};
 use crate::store::{Result, SamplingMetadata, SamplingStatus, Store, StoreError};
 
 /// indexeddb version, needs to be incremented on every schema schange
-const DB_VERSION: u32 = 2;
+const DB_VERSION: u32 = 3;
```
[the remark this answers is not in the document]
fl0rek: This will preserve existing headers - we're using the same db, so we might as well keep them I think. Pruning window is coming soon, should clean up the unneeded historical data.

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,530 @@
...
+            )*
+            ])
+        }
+    };
+}
+
+pub(crate) use header_ranges;
+
+/// Span of header that's been verified internally
+pub struct VerifiedHeaderSpan(Vec<ExtendedHeader>);
```
oblique, line 29: I prefer `VerifiedExtentedHeaders` as a name

node/src/store.rs
```diff
@@ -121,12 +126,25 @@ pub trait Store: Send + Sync + Debug {
...
+
+    /// Insert list of headers, validating that the inserted range is contiguous
+    async fn append_unchecked(&self, headers: Vec<ExtendedHeader>) -> Result<()> {
+        self.insert(headers.try_into()?).await
+    }
+
+    /// Insert list of headers validating them against neighbouring headers if they exist in store
+    async fn append(&self, headers: Vec<ExtendedHeader>) -> Result<()> {
+        self.insert(headers.try_into()?).await
+    }
```
oblique, line 147: Remove all of them and use `TryInto<VerifiedExtentedHeaders>` at `insert` parameters

node/src/store/redb_store.rs
```diff
@@ -410,6 +426,110 @@ impl Store for RedbStore {
...
+        range_to_remove,
+    } = stored_ranges.check_range_insert(&new_range)?;
+
+    if let Some(to_remove) = range_to_remove {
+        let (start, end) = ranges_table
+            .remove(u64::try_from(to_remove).expect("usize->u64"))?
+            .expect("missing range")
+            .value();
+
+        info!("consolidating range, new range: {range:?}, removed {start}..={end}");
```
oblique, line 461: `debug` or remove it

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,530 @@
...
+}
+
+impl AsRef<[ExtendedHeader]> for VerifiedHeaderSpan {
+    fn as_ref(&self) -> &[ExtendedHeader] {
+        &self.0
+    }
+}
+
+/// 1-length hedaer span is internally verified, this is valid
+impl From<[ExtendedHeader; 1]> for VerifiedHeaderSpan {
```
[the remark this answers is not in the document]
fl0rek: Added

node/src/store/header_ranges.rs
fl0rek: I like it better too, changed 👍

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,560 @@
...
+
+use crate::store::utils::{ranges_intersection, try_consolidate_ranges, RangeScanResult};
+use crate::store::StoreError;
+
+pub type HeaderRange = RangeInclusive<u64>;
+
+macro_rules! header_ranges {
+    ( $( $r:expr ),* ) => {
+        {
+            crate::store::header_ranges::HeaderRanges::from_vec(smallvec::smallvec![
```
oblique, line 19: I was a bit wrong before. This is the correct one:

```suggestion
            $crate::store::header_ranges::HeaderRanges::from_vec(smallvec::smallvec![
```

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,560 @@
...
+        head.verify_adjacent_range(&headers[1..])?;
+
+        Ok(Self(headers))
+    }
+}
+
+impl VerifiedExtendedHeaders {
+    /// Create a new instance out of pre-checked vec of headers
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn from_verified_vec(headers: Vec<ExtendedHeader>) -> Self {
```
oblique, line 79: I prefer this:

```suggestion
    pub unsafe fn new_unchecked(headers: Vec<ExtendedHeader>) -> Self {
```

node/src/store.rs
```diff
@@ -147,50 +145,15 @@ pub trait Store: Send + Sync + Debug {
...
     ///
-    /// This method does not validate or verify that `headers` are indeed correct.
-    async fn append_unchecked(&self, headers: Vec<ExtendedHeader>) -> Result<()> {
-        for header in headers.into_iter() {
-            self.append_single_unchecked(header).await?;
-        }
+    /// `verify_neighbours` determines whether entire range will be validated against headers next
+    /// to them (if present in store), or whether to just validate a contiguity of the inserted
+    /// header range
+    async fn insert(&self, headers: VerifiedExtendedHeaders) -> Result<()>;
```
oblique, line 153: I meant the following and remove all `try_into`/`into` from callers.


```suggestion
    async fn insert(&self, headers: impl TryInto<VerifiedExtendedHeaders>) -> Result<()>;
```

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,560 @@
...
+    /// Generate next amount verified headers
+    fn next_many_verified(&mut self, amount: u64) -> VerifiedExtendedHeaders;
+}
+
+#[cfg(any(test, feature = "test-utils"))]
+impl ExtendedHeaderGeneratorExt for ExtendedHeaderGenerator {
+    fn next_many_verified(&mut self, amount: u64) -> VerifiedExtendedHeaders {
+        VerifiedExtendedHeaders::from_verified_vec(self.next_many(amount))
+    }
+}
```
oblique, line 322: Maybe for later: We can actually move `VerifiedExtendedHeaders` in `celestia-types` and `ExtendedHeaderGenerator` should always return `VerifiedExtendedHeaders`.

node/src/store/header_ranges.rs
oblique: Yes. It does not really affect as because we do not expose the macro to users.
Check this: https://doc.rust-lang.org/reference/macros-by-example.html#hygiene

node/src/store/header_ranges.rs
zvolin: can there be a case we want to generate not verified?

node/src/store/header_ranges.rs
zvolin: oh nice, I guess I missed / forgot about the `$crate` syntax

node/src/store/header_ranges.rs
oblique: No. But we have this [helper](https://github.com/eigerco/lumina/blob/a20665fc42e73ef35f4b497130f3006e755e1ca1/types/src/test_utils.rs#L275)

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,560 @@
...
+impl IntoIterator for VerifiedExtendedHeaders {
+    type Item = ExtendedHeader;
+    type IntoIter = vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl AsRef<[ExtendedHeader]> for VerifiedExtendedHeaders {
```
oblique, line 43: Add `impl From<VerifiedExtendedHeaders> for Vec<ExtendedHeaders>`

node/src/store.rs
fl0rek: Nice! applied

node/src/store.rs
```diff
@@ -256,6 +244,13 @@ impl From<tokio::task::JoinError> for StoreError {
     }
 }
 
+impl From<Infallible> for StoreError {
```
oblique, line 247:
```suggestion
// Needed for `Into<VerifiedExtendedHeaders>`
impl From<Infallible> for StoreError {
```

node/src/store.rs
```diff
@@ -147,50 +146,26 @@ pub trait Store: Send + Sync + Debug {
...
-        }
-
-        self.append_single_unchecked(header).await
+    /// Inserts are allowed at the front of the store or at the ends of any existing ranges. Edges
+    /// of inserted header ranges are verified against headers present in the store, if they
+    /// exist.
+    async fn insert<R>(&self, headers: R) -> Result<()>
+    where
+        R: TryInto<VerifiedExtendedHeaders> + Send,
+        //<R as TryInto<VerifiedExtendedHeaders>>::Error: Into<StoreError>
```
oblique, line 157:
```suggestion
```

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,566 @@
...
+
+impl IntoIterator for VerifiedExtendedHeaders {
+    type Item = ExtendedHeader;
+    type IntoIter = vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
```
oblique, line 42: Add also this and remove the `to_vec()` from the callers

```suggestion

impl<'a> TryFrom<&'a [ExtendedHeader]> for VerifiedExtendedHeaders {
    type Error = celestia_types::Error;

    fn try_from(value: &'a [ExtendedHeader]) -> Result<Self, Self::Error> {
        value.to_vec().try_into()
    }
}

```

node/src/syncer.rs
```diff
@@ -404,7 +400,7 @@ where
                 // If our new header is adjacent to the HEAD of the store
                 if store_head_height + 1 == new_head_height {
                     // Header is already verified by HeaderSub
-                    if self.store.append_single_unchecked(new_head).await.is_ok() {
+                    if self.store.insert([new_head]).await.is_ok() {
```
oblique, line 403:
```suggestion
                    if self.store.insert(new_head).await.is_ok() {
```

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,568 @@
...
+        Self(value.into())
+    }
+}
+
+impl From<ExtendedHeader> for VerifiedExtendedHeaders {
+    fn from(value: ExtendedHeader) -> Self {
+        Self(vec![value])
+    }
+}
+
```
oblique, line 67: And remove `.clone()` on the callers

```suggestion

impl<'a> From<&'a ExtendedHeader> for VerifiedExtendedHeaders {
    fn from(value: &ExtendedHeader) -> Self {
        Self(vec![value.to_owned()])
    }
}

```

node/src/store/header_ranges.rs
```diff
@@ -0,0 +1,568 @@
...
+        head.verify_adjacent_range(&headers[1..])?;
+
+        Ok(Self(headers))
+    }
+}
+
+impl VerifiedExtendedHeaders {
+    /// Create a new instance out of pre-checked vec of headers
+    /// # Safety
```
oblique, line 84:
```suggestion
    ///
    /// # Safety
    ///
```

node/src/store.rs
```diff
@@ -256,6 +244,13 @@ impl From<tokio::task::JoinError> for StoreError {
     }
 }
 
+impl From<Infallible> for StoreError {
+    fn from(_: Infallible) -> Self {
+        // Infallable should not be possible to construct
+        unreachable!("")
```
oblique, line 250:
```suggestion
        unreachable!("Infallible failed")
```

node/src/store.rs
```diff
@@ -147,50 +146,26 @@ pub trait Store: Send + Sync + Debug {
...
-        match self.get_head().await {
-            Ok(head) => {
-                head.verify_adjacent_range(&headers)?;
-            }
-            // Empty store, we can not verify
-            Err(StoreError::NotFound) => {}
-            Err(e) => return Err(e),
-        }
+    /// Insert a range of headers into the store, see wrapper [`Store::insert`].
+    async fn insert_verified_headers(&self, header: VerifiedExtendedHeaders) -> Result<()>;
```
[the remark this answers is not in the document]
fl0rek: reworked to not include it 👍

node/src/store/header_ranges.rs
fl0rek: https://github.com/eigerco/lumina/issues/301

oblique (approved):
LGTM

node/src/syncer.rs
```diff
@@ -404,7 +400,7 @@ where
                 // If our new header is adjacent to the HEAD of the store
                 if store_head_height + 1 == new_head_height {
```
oblique, line 401: This needs to be relaxed and `self.ongoing_batch.is_none()` to be removed

oblique: Will be done in another PR

PR #284, opened 2024-05-22
feat(node-wasm): Implement easier way for handling JS errors.

oblique:


node-wasm/src/error.rs
```diff
@@ -0,0 +1,139 @@
...
+
+pub(crate) struct Error(JsValue);
+
+impl Error {
+    /// Create a new `Error` with the specified message.
+    pub(crate) fn new(msg: &str) -> Error {
+        Error(js_sys::Error::new(msg).into())
+    }
+
+    /// Create a new `Error` from `JsValue` without lossing any information.
```
fl0rek, line 17:
```suggestion
    /// Create a new `Error` from `JsValue` without losing any information.
```

node-wasm/src/utils.rs
```diff
@@ -71,27 +73,10 @@ pub(crate) fn js_value_from_display<D: fmt::Display>(value: D) -> JsValue {
...
-    where
-        C: fmt::Display + Send + Sync + 'static,
-    {
-        self.map_err(|e| JsError::new(&format!("{context}: {e}")))
-    }
-}
-
-pub(crate) fn get_crypto() -> Result<Crypto, JsError> {
+pub(crate) fn get_crypto() -> Result<Crypto, Error> {
     js_sys::Reflect::get(&js_sys::global(), &JsValue::from_str("crypto"))
```
fl0rek, line 77: by the looks of it, it should work both in normal js and in worker, nice one!

oblique: Yes. That is why I did it like this.

fl0rek (approved):
🎸 🎸

Issue #288, opened 2024-06-06
Add `#[track_caller]` in all assertion functions


oblique:

zvolin:
is this still relevant?

oblique:
`#[track_caller]` does not support async fn yet (https://github.com/rust-lang/rust/issues/110011) so some assertion functions in tests can not use it. Since what is left it is only related to test cases, we can close it.

PR #305, opened 2024-06-19
feat: Add `#[track_caller]` on non-async test utils and spawn utils

oblique:


oblique:
Closes #288

zvolin (approved):

fl0rek (approved):

oblique:
So `#[track_caller]` works only on non-async fns. There is a nightly feature for the async `track_caller`, but there is no progress on it (https://github.com/rust-lang/rust/issues/110011).

PR #306, opened 2024-06-19
feat(node): Always start data sampling of new HEAD immediately

oblique:


node/src/daser.rs
```diff
@@ -26,16 +26,17 @@
 //!    - If we reach a timeout of 10 seconds and at least one of the CIDs is not received, then
 //!      block is considered sampled and rejected.
 //!    - [`Store`] is updated with the sampling result.
-//! 5. Steps 3 and 4 are repeated concurently, unless we detect that all peers have disconnected.
+//! 5. Steps 3 and 4-5-6 are repeated concurently, unless we detect that all peers have disconnected.
```
zvolin, line 29:
```suggestion
//! 5. Steps 3, 4, 5 and 6 are repeated concurently, unless we detect that all peers have disconnected.
```

zvolin (approved):
:+1:

node/src/daser.rs
```diff
@@ -499,8 +506,14 @@ mod tests {
     use celestia_types::test_utils::{generate_eds, ExtendedHeaderGenerator};
     use celestia_types::{AxisType, DataAvailabilityHeader, ExtendedDataSquare};
     use cid::Cid;
+    use std::collections::HashMap;
     use std::time::Duration;
 
+    // In which request number the tests can simulate invalid sampling
```
fl0rek, line 512:
```suggestion
    // Request number for which tests will simulate invalid sampling
```

fl0rek: Also curious, why is this a `const` rather than plain variable passed in test?

node/src/daser.rs
```diff
@@ -682,118 +689,72 @@ mod tests {
...
-        handle: &mut MockP2pHandle,
-        store: &InMemoryStore,
-        event_sub: &mut EventSubscriber,
-        height: u64,
-        eds: &ExtendedDataSquare,
-        simulate_invalid_sampling: bool,
-    ) -> Vec<Cid> {
-        let square_width = eds.square_width() as usize;
-        let needed_samples = (square_width * square_width).min(MAX_SAMPLES_NEEDED);
+        // Check if CIDs we received succefully made it in the store
```
fl0rek, line 705:
```suggestion
        // Check if CIDs we received successfully made it in the store
```

node/src/daser.rs
```diff
@@ -682,118 +689,72 @@ mod tests {
...
+        assert_eq!(&sampling_metadata.cids, &cids);
 
-        // Check if we get the correct event
-        let mut remaining_shares = match expect_event(event_sub).await {
+        // Check if received `SamplingStarted` event
```
fl0rek, line 708:
```suggestion
        // Check if we received `SamplingStarted` event
```

node/src/daser.rs
oblique: The reason I did it like this is because 4-5-6 run as one thing. They are the same flow.

node/src/daser.rs
```diff
@@ -359,56 +380,42 @@ where
...
+                if first_check && is_block_accepted(&*self.store, height).await {
+                    // Skip already sampled blocks
+                    continue;
+                }
 
-                    if !in_sampling_window(header.time()) {
-                        // We've reached the tail of the sampling window
-                        break;
-                    }
+                let Ok(header) = self.store.get_by_height(height).await else {
```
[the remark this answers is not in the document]
oblique: If I remember correct pruning window is a bit bigger than sampling window so we never reach a "to-be-pruned" block.

node/src/daser.rs
oblique: Because 2nd share will always be requested and it is a bit simpler to have a const than an argument.

node/src/daser.rs
oblique: After some rewording @fl0rek did it made sense to merge the into one.

zvolin (approved):

PR #308, opened 2024-06-21
chore(node): Add comments about header validation

oblique:


fl0rek (approved):
🙌

node/src/p2p.rs
```diff
@@ -407,6 +409,11 @@ impl P2p {
         let mut session = HeaderSession::new(range, self.cmd_tx.clone());
         let headers = session.run().await?;
 
+        // Cryptographic validation is done on each header separatetly
+        // by `HeaderExClientHandler`.
```
zvolin, line 413: I think I'd prefer

```suggestion
        // `.validate()` is called on each header separately by `HeaderExClientHandler`.
```

zvolin (approved):

node/src/p2p.rs
```diff
@@ -407,6 +409,10 @@ impl P2p {
         let mut session = HeaderSession::new(range, self.cmd_tx.clone());
         let headers = session.run().await?;
 
+        // `.validate()` is called on each header separately by `HeaderExClientHandler`.
+        //
+        // The last step is to verify that all headers are from the same chain
+        // and indeed connected with the next one.
```
oblique, line 415:
```suggestion
        // `.validate()` was called on each header separately by `HeaderExClientHandler`.
        // The last step is to verify that all headers are from the same chain
        // and indeed connected with the next one.
```

PR #309, opened 2024-06-25
feat(node): Add syncing window for header sync

fl0rek:
When starting a new syncing session, check whether header rigth after that (if present in store) is still withing syncing window. This way we fetch at most one syncing batch (512) of old headers outside of syncing window without complicating headers sessions too much.


node/src/p2p/header_session.rs
```diff
@@ -9,7 +9,7 @@ use crate::p2p::{P2pCmd, P2pError};
 use crate::store::header_ranges::{HeaderRange, RangeLengthExt};
 
 const MAX_AMOUNT_PER_REQ: u64 = 64;
-const MAX_CONCURRENT_REQS: usize = 1;
+const MAX_CONCURRENT_REQS: usize = 8;
```
fl0rek, line 12: This was accidentaly de-paralelised when I was testing backwards header sync. Reverting to previous value.

zvolin (approved):

node/src/syncer.rs
```diff
@@ -443,6 +445,19 @@ where
...
+        });
+
+        // make sure we're inside the syncing window before we start
+        if let Ok(known_header) = self.store.get_by_height(next_batch.end() + 1).await {
+            if known_header.time().before(syncing_window_start) {
+                debug!("synced to the end of syncing window");
+                return;
+            }
+        }
+
```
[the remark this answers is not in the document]
oblique: I also prefer a helper function. In `daser.rs` we have something similar but called `in_sampling_window`

node/src/syncer.rs
```diff
@@ -443,6 +445,19 @@ where
...
 
+        let syncing_window_start = Time::now().checked_sub(SYNCING_WINDOW).unwrap_or_else(|| {
+            warn!("underflow when computing syncing window start, defaulting to unix epoch");
+            Time::unix_epoch()
+        });
+
+        // make sure we're inside the syncing window before we start
+        if let Ok(known_header) = self.store.get_by_height(next_batch.end() + 1).await {
+            if known_header.time().before(syncing_window_start) {
+                debug!("synced to the end of syncing window");
```
oblique, line 456: i think when we reach this line, we can safely say that this batch and anything before it, should not be rechecked when `fetch_next_batch` is called.

we should introduce a height limit that will be set to `next_batch.end() + 1` and `calculate_range_to_fetch` should exclude anything before that.

node/src/syncer.rs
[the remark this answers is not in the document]
fl0rek: Added separate `in_syncing_window` 👍

node/src/syncer.rs
fl0rek: Caching added a bit of code here, but added saing the known syncing window edge and accounting for it in ranges calculcation 👍

node/src/store/utils.rs
```diff
@@ -13,34 +13,48 @@ pub(crate) const VALIDATIONS_PER_YIELD: usize = 4;
...
     let mut store_headers_iter = store_headers.iter().rev();
 
     let Some(store_head_range) = store_headers_iter.next() else {
-        // empty store, just fetch from head
-        return head_height.saturating_sub(limit) + 1..=head_height;
+        // empty store, we're missing everything
+        return 1..=head_height;
     };
 
     if store_head_range.end() != &head_height {
```
oblique, line 49: This shouldn't be `store_head_range.end() < &head_height`?

fl0rek: Store head height > network head height, so probably something already went bit wrong, but I don't suppose it'd hurt here.

zvolin (approved):

PR #312, opened 2024-06-26
feat(node): Generate syncing related events

oblique:


node/src/events.rs
```diff
@@ -272,6 +312,29 @@ impl fmt::Display for NodeEvent {
             NodeEvent::FatalDaserError { error } => {
                 write!(f, "Daser stopped because of a fatal error: {error}")
             }
+            NodeEvent::NewSubjectiveHead { height } => {
+                write!(f, "New subjective head: {height}")
+            }
+            NodeEvent::AddedHeaderFromHeaderSub { height } => {
+                write!(f, "Added header {height} from HeaderSub")
```
zvolin, line 319:
```suggestion
                write!(f, "Added header {height} from header-sub")
```
I think protocol name uses this format, but up to u

node/src/events.rs
```diff
@@ -219,6 +228,37 @@ pub enum NodeEvent {
...
+    AddedHeaderFromHeaderSub {
+        /// The height of the header.
+        height: u64,
+    },
+
+    /// Fetching header of network head in progress.
+    FetchingHeadHeader,
+
+    /// Fetching headers for a specific block range.
+    FetchingHeaders {
```
zvolin, line 248: I think we should also have an event when we succeeded fetching and inserted them

cli/src/native.rs
```diff
@@ -80,14 +80,15 @@ pub(crate) async fn run(args: Params) -> Result<()> {
...
-    let mut events = node.event_subscriber();
-
-    // We have nothing else to do, but we want to keep main alive
     while let Ok(ev) = events.recv().await {
         match ev.event {
             // Skip noisy events
             NodeEvent::ShareSamplingResult { .. } => continue,
+            event @ (NodeEvent::FatalDaserError { .. }
+            | NodeEvent::FetchingHeadersFailed { .. }
+            | NodeEvent::NetworkCompromised) => {
```
oblique, line 89: I was thinking if I should add a `is_error` method in `NodeEvent`. What do you think?

node/src/events.rs
oblique: Done

cli/src/native.rs
zvolin: sure, why not

oblique: Done. Should we consider a rejected `ShareSamplingResult` an error and display it as a warning?

oblique: Also should I propagate the `is_error` to JS?

zvolin: I think yes to both

oblique: done

node/src/events.rs
```diff
@@ -219,6 +228,82 @@ pub enum NodeEvent {
...
+    FetchingHeadersFinished {
+        /// Start of the range.
+        from_height: u64,
+        /// End of the range (included).
+        to_height: u64,
+        /// How much time fetching took.
+        took: Duration,
+    },
+
+    /// Fetching headers of a specific block range just finished.
```
fl0rek, line 267:
```suggestion
    /// Fetching headers of a specific block range just failed.
```

node/src/syncer.rs
```diff
@@ -419,6 +444,12 @@ where
             return;
         }
 
+        if self.p2p.peer_tracker_info().num_connected_peers == 0 {
+            // No connected peers. We can't do the request.
+            // This will be recovered by `run`.
```
fl0rek, line 449:
```suggestion
            // We will recover from this in `run`.
```

fl0rek (commented):
nice work 🙌

zvolin (approved):

fl0rek (approved):

PR #313, opened 2024-06-26
fix(node-wasm): require serving and providing worker script

zvolin:


oblique (approved):
🙈

zvolin:
I believe we can do better in future but for now it's the only way we got it to work with bundlers like webpack :see_no_evil:

fl0rek (approved):

Issue #317, opened 2024-07-02
Graceful handling and recovery for QuotaExceededError


fl0rek:
## Context
Browsers use quotas to limit the amount of data page can store locally, see [mdn](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria) for details.

Available quota can be queried with [StorageManager](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager)

## Issue
When transaction which would put us above the quota is commited, `QuotaExceededError` is returned and transaction is rolled back. From this point on, Indexeddb is in read-only state until restart (note that this by itself doesn't change the quota, and db can lock up again, if quota is exceeded).

From the user perspective, node just stops syncing (as all the db inserts fail).

zvolin:
from sync discussion:
 - this is kinda similar to no space left on device error, we can't handle it really
 - but we can document it
 - and allow configuring the syncing window period

PR #318, opened 2024-07-02
feat: Add requesting storage persistence for more quota

fl0rek:
related #317 

[MDN page](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria) isn't exact on the details (as detail vary by browser), but requesting persistent storage on firefox raises the quota limit (and eliminates quota issues on my machine).
Browsers should also try more to keep our data from being removed.

Node that this will display prompt on Firefox, while chromium based browsers use heuristics to figure out whether to grant the storage.

image

### DIFF
--- a/node-wasm/src/error.rs
+++ b/node-wasm/src/error.rs
@@ -1,11 +1,11 @@
 //! Error type and utilities.
 
-use std::fmt::Display;
+use std::fmt::{self, Display};
 
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::convert::IntoWasmAbi;
 use wasm_bindgen::describe::WasmDescribe;
-use wasm_bindgen::JsValue;
+use wasm_bindgen::{JsCast, JsValue};
 
 /// Alias for a `Result` with the error type [`Error`].
 pub type Result<T, E = Error> = std::result::Result<T, E>;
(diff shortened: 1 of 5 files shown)


node-wasm/src/node.rs
```diff
@@ -92,6 +93,10 @@ impl NodeDriver {
         worker_script_url: &str,
         worker_type: Option<NodeWorkerKind>,
     ) -> Result<NodeDriver> {
+        if let Err(e) = request_storage_persistence().await {
```
zvolin, line 96: since this only affects quota in ff, can we guard this to be only called in ff?

fl0rek: I could have mentioned that beforehand, but

> In Firefox, when a site chooses to use persistent storage, the user is notified with a UI popup that their permission is requested.
> Safari and most Chromium-based browsers, such as Chrome or Edge, automatically approve or deny the request based on the user's history of interaction with the site and do not show any prompts to the user.

So the prompt would be displayed on firefox, other browser doing some heuristics. Maybe just leave it for all browsers then? We are intending on manually managing the data anyway and I want to do as little parsing of UserAgent as possible. Sniffing for chrome already felt bad enough :P

node-wasm/src/error.rs
```diff
@@ -50,6 +50,16 @@ impl Error {
     }
 }
 
+impl Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if let Some(error) = self.0.dyn_ref::<js_sys::Error>() {
+            write!(f, "{} ({})", error.name(), error.message())
```
zvolin, line 56: could you also add a new line with the `.cause()`?

oblique (approved):
LGTM 🙈

node-wasm/src/error.rs
fl0rek: Good idea, added printing `cause`s in a loop until we get to the bottom of the error 👍

oblique (approved):

zvolin (approved):

PR #328, opened 2024-07-10
Feat/store tail removal

fl0rek:


node/src/block_ranges.rs
```diff
@@ -443,6 +443,57 @@ impl Sub<&BlockRanges> for BlockRanges {
...
+        self.remove_relaxed(range).expect("shouldn't panic");
+        self
+    }
+}
+
+impl SubAssign<RangeInclusive<u64>> for BlockRanges {
+    fn sub_assign(&mut self, range: RangeInclusive<u64>) {
+        self.remove_relaxed(range).expect("shouldn't panic");
+    }
+}
```
oblique, line 459: These can panic if the range is invalid. For example `0..=10` or `11..=10`.

node/src/block_ranges.rs
```diff
@@ -443,6 +443,57 @@ impl Sub<&BlockRanges> for BlockRanges {
...
+        let next = self.current_range.as_mut()?.next();
+
+        if next.is_none() {
+            self.current_range = self.ranges.next();
+            return self.current_range.as_mut()?.next();
+        }
+
+        next
+    }
+}
```
oblique, line 495: Add a `pop_tail` (similar to `pop_head`) and then you can implement iterator without having `current_range`. Also why not implementing iterator on `BlockRanges` itself?

node/src/block_ranges.rs
fl0rek: Fixed to effectively ignore removal of empty ranges, added tests including this case.

node/src/block_ranges.rs
fl0rek: I think I started from `IntoIterator` and it sort of went that way :P
`impl Iterator for BlockRanges` is with `pop_tail` is indeed simpler, changed :+1:

node/src/block_ranges.rs
```diff
@@ -443,6 +457,33 @@ impl Sub<&BlockRanges> for BlockRanges {
...
+
+impl SubAssign<RangeInclusive<u64>> for BlockRanges {
+    fn sub_assign(&mut self, range: RangeInclusive<u64>) {
+        if !range.is_empty() {
+            self.remove_relaxed(range).expect("shouldn't panic");
+        }
+    }
+}
```
oblique, line 477: I think I don't like them. IMO, if we do something like this, we should make `remove_relaxed` infallible, which opens other questions in that case.

node/src/block_ranges.rs
```diff
@@ -443,6 +457,33 @@ impl Sub<&BlockRanges> for BlockRanges {
...
+
+impl Iterator for BlockRanges {
```
oblique, line 479: Implement [DoubleEndedIterator](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) too. With that we will be able to use `rev()`.

node/src/block_ranges.rs
oblique: We can just return in `insert_relaxed` and `remove_relaxed` if range does not pass `validate()`. However I'm not sure if this will enable us to do any mistakes.

@zvolin Wdyt?

fl0rek: My mental model for ranges is that of algebra of sets, so even though subtrahend here could be outside of `HeaderRange` strict domain, there's still a single, well-defined, and intuitive product we can return instead of erroring.
Since headers are going to be removed one by one, using `pop_tail`, I left these `impl`s mostly out of sentiment, so we can consider adding operators separately, if that's controversial.

node/src/block_ranges.rs
fl0rek: implemented :+1:

node/src/block_ranges.rs
oblique: In case of sets, that means `10..=5` should be equal to `5..=10`. Which again it may allow us to do mistakes.

node/src/block_ranges.rs
```diff
@@ -735,6 +741,16 @@ mod tests {
         assert_eq!(empty_heights, Vec::<u64>::new())
     }
 
+    #[test]
+    fn block_ranges_double_ended_iterator() {
+        let ranges = new_block_ranges([1..=5, 10..=15]);
+        let heights: Vec<_> = ranges.into_iter().rev().collect();
```
oblique, line 747: I think `into_iter()` is not needed

node/src/store/in_memory_store.rs
```diff
@@ -301,6 +306,35 @@ impl InMemoryStoreInner {
 
         Ok(Some(metadata.clone()))
     }
+
+    fn remove_last(&mut self) -> Result<u64> {
```
oblique, line 310: I think instead of returning `u64`, which many never be used, it should be nice to return `(ExtendedHeader, SamplingMetadata)`. That way you can use CIDs of sampling metadata without having a possibility of get/set/remove race.

node/src/store/indexed_db_store.rs
```diff
@@ -378,6 +299,30 @@ impl IndexedDbStore {
...
+
+    async fn remove_last(&self) -> Result<u64> {
+        self.write_tx(
+            &[HEADER_STORE_NAME, RANGES_STORE_NAME],
+            remove_last_tx_op,
+            (),
+        )
+        .await
+    }
+}
```
oblique, line 311:
```suggestion
}

```

node/src/store/redb_store.rs
```diff
@@ -413,6 +410,40 @@ impl RedbStore {
...
+
+            let hash = ExtendedHeader::decode(header.value())
+                .map_err(|e| StoreError::StoredDataError(e.to_string()))?
+                .hash();
+
+            if heights_table.remove(hash.as_bytes())?.is_none() {
+                return Err(StoreError::StoredDataError(format!(
+                    "inconsistency between header and height_to_hash tables, hash {hash}"
+                )));
+            };
```
oblique, line 441:
```suggestion
            }
```

node/src/store/redb_store.rs
```diff
@@ -504,6 +535,9 @@ impl Store for RedbStore {
     async fn get_accepted_sampling_ranges(&self) -> Result<BlockRanges> {
         self.get_sampling_ranges().await
     }
+    async fn remove_last(&self) -> Result<u64> {
```
oblique, line 538:
```suggestion

    async fn remove_last(&self) -> Result<u64> {
```

node/src/store/indexed_db_store.rs
```diff
@@ -378,6 +378,34 @@ impl IndexedDbStore {
...
+        set_ranges(&ranges_store, HEADER_RANGES_KEY, &header_ranges).await?;
+
+        let jsvalue_height = to_value(&height).expect("to create jsvalue");
+        let header = height_index.get(&jsvalue_height).await?;
+
+        let id = js_sys::Reflect::get(&header, &to_value("id")?)
+            .map_err(|_| StoreError::StoredDataError("could not get header's DB id".into()))?;
+
+        header_store.delete(&id).await?;
+
```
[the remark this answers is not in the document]
fl0rek: Added `write_tx` to deal with txn commitment more generaly.

oblique (approved):
LGTM

zvolin (approved):

PR #333, opened 2024-07-15
fix(node)!: Do not skip header-sub reports when store writes are slow

oblique:


oblique:
This contains #332 commits. Will be rebased after #332 is merged.

node/src/syncer.rs
```diff
@@ -618,12 +613,28 @@ where
...
-    // Insert HEAD to the store and initialize header-sub
+    // Insert HEAD to the store and initialize header-sub.
+    // This will apply insertion restrictions.
     store.insert(network_head.clone()).await?;
 
     Ok(network_head)
 }
 
+async fn header_sub_recv(
+    rx: &mut Option<mpsc::Receiver<ExtendedHeader>>,
```
zvolin, line 624: I think it'd be more idiomatic to accept `Option<&mut ...>`

node/src/syncer.rs
```diff
@@ -618,12 +613,28 @@ where
...
+    poll_fn(|cx| {
+        let rx = rx.as_mut().expect("header-sub not initialized");
+
+        match rx.poll_recv(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Some(header)) => Poll::Ready(Ok(header)),
+            Poll::Ready(None) => Poll::Ready(Err(SyncerError::P2p(P2pError::WorkerDied))),
+        }
+    })
+    .await
```
zvolin, line 635: I don't understand, why this cannot be
```rs
rx.recv().await.ok_or(...)
```

oblique: When I was writing this, its first implementation was returning `Poll::Pending` on closed channel instead of an error. At the time I didn't thought to change it to `ok_or` when I changed it to return an error.

node/src/syncer.rs
oblique: done

node/src/syncer.rs
oblique: switched to `ok_or`

oblique:
CI is failing because of this: https://github.com/Sovereign-Labs/nmt-rs/pull/30

fl0rek (approved):
LGTM! 💯

PR #336, opened 2024-07-15
fix(rpc): Increase max response size

oblique:


rpc/src/client.rs
```diff
@@ -27,6 +29,13 @@ mod native {
 
     use crate::Error;
 
+    const MAX_SHARES_SIZE_IN_EDS: usize =
```
zvolin, line 32: this feels a bit weird, how about sth like `MAX_EDS_SIZE_BYTES`

oblique: sure

zvolin (approved):
thanks, should we wait for confirmation from @S1nus?

PR #338, opened 2024-07-16
fix(types)!: Align byzantine fraud proofs with Go's implementation

### LINKED ISSUE #337: Difference in handling bad encoding fraud proofs to ref implementation
In the go implementation, `ShareWithProof` has an axis field: https://github.com/celestiaorg/celestia-node/blob/f98d632818d566c7b4fd995b0f4bdc6443a7ed06/share/eds/byzantine/share_proof.go#L29-L30 and befp itself as well: 
https://github.com/celestiaorg/celestia-node/blob/f98d632818d566c7b4fd995b0f4bdc6443a7ed06/share/eds/byzantine/bad_encoding.go#L33
In go, it is used here: https://github.com/celestiaorg/celestia-node/blob/f98d632818d566c7b4fd995b0f4bdc6443a7ed06/share/eds/byzantine/share_proof.go#L36C9-L41

In Rust/lumina, the axis in the shareWithProof is not used as it does not exist: https://github.com/eigerco/lumina/blob/4f2bbc602d9c8921b0d2054af751ed9793456ad0/types/src/byzantine.rs#L185-L188

After chatting with @walldiss, it seems that this was introduced later in node to handle all cases. Would be good if this was on-par with the reference implementation even if this code path will likely never be used 🤞🏼

oblique:
Fixes #337

### DIFF
--- a/types/src/byzantine.rs
+++ b/types/src/byzantine.rs
@@ -2,21 +2,14 @@ use celestia_proto::share::eds::byzantine::pb::BadEncoding as RawBadEncodingFrau
 use celestia_proto::share::eds::byzantine::pb::Share as RawShareWithProof;
 use celestia_tendermint::{block::Height, Hash};
 use celestia_tendermint_proto::Protobuf;
-use cid::CidGeneric;
 use serde::{Deserialize, Serialize};
 
 use crate::bail_validation;
 use crate::consts::appconsts;
 use crate::fraud_proof::FraudProof;
-use crate::nmt::{
-    Namespace, NamespaceProof, NamespacedHash, NamespacedHashExt, Nmt, NmtExt, NMT_CODEC,
-    NMT_ID_SIZE, NMT_MULTIHASH_CODE, NS_SIZE,
-};
+use crate::nmt::{Namespace, NamespaceProof, Nmt, NmtExt, NS_SIZE};
 use crate::rsmt2d::AxisType;
 use crate::{Error, ExtendedHeader, Result};
-
-type Cid = CidGeneric<NMT_ID_SIZE>;
-type Multihash = multihash::Multihash<NMT_ID_SIZE>;
 
 /// A proof that the block producer incorrectly encoded [`ExtendedDataSquare`].
 ///
@@ -235,9 +228,7 @@ impl TryFrom<RawShareWithProof> for ShareWithProof {
             return Err(Error::WrongProofType);
         }
 
-        let proof_axis = u8::try_from(value.proof_axis)
-            .map_err(|_| Error::InvalidAxis(value.proof_axis))?
-            .try_into()?;
+        let proof_axis = value.proof_axis.try_into()?;
 
         Ok(Self {
             leaf,
@@ -263,9 +254,7 @@ impl TryFrom<RawBadEncodingFraudProof> for BadEncodingFraudProof {
     type Error = Error;
 
     fn try_from(value: RawBadEncodingFraudProof) -> Result<Self, Self::Error> {
-        let axis = u8::try_from(value.axis)
-            .map_err(|_| Error::InvalidAxis(value.axis))?
-            .try_into()?;
+        let axis = value.axis.try_into()?;
 
         let index = u16::try_from(value.index).map_err(|_| Error::EdsInvalidDimentions)?;
 
--- a/types/src/byzantine.rs
+++ b/types/src/byzantine.rs
@@ -359,10 +359,10 @@ pub(crate) mod test_utils {
 
         // collect the shares for fraud proof
         for share_idx in 0..square_width {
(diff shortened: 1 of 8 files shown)


oblique:
@walldiss @wondertan, can you confirm if my changes in `byzantine.rs` are correct? Especially [this part](https://github.com/eigerco/lumina/pull/338/files#diff-7a8c9d84d7cea790a49dd6c231adc274636aa57261274bc7cf87e36690838f80R121-R126).

oblique:
@Wondertan We believe there is a bug in your [celestia-node implementation](https://github.com/celestiaorg/celestia-node/blob/f98d632818d566c7b4fd995b0f4bdc6443a7ed06/share/eds/byzantine/share_proof.go#L36-L41) (called from [here](https://github.com/celestiaorg/celestia-node/blob/f98d632818d566c7b4fd995b0f4bdc6443a7ed06/share/eds/byzantine/bad_encoding.go#L183)) or we are really missing something.

Your code simplifies to this:

```go
	switch s.Axis {
	case rsmt2d.Row:
		rootHash = dah.RowRoots[axisIdx]
	case rsmt2d.Col:
		rootHash = dah.ColumnRoots[axisIdx]
	}
```

However our code translates to this:

```go
	switch axisType {
	case rsmt2d.Row:
		switch s.Axis {
		case rsmt2d.Row:
			rootHash = dah.RowRoots[axisIdx]
		case rsmt2d.Col:
			rootHash = dah.ColumnRoots[shrIdx]
		}
	case rsmt2d.Col:
		switch s.Axis {
		case rsmt2d.Row:
			rootHash = dah.RowRoots[shrIdx]
		case rsmt2d.Col:
			rootHash = dah.ColumnRoots[axisIdx]
		}
	}
```

walldiss:
We will update our BEFP ADR soon, that will cover all changes of BEFP construction and validation logic. To unblock you I'll clarify algorithm here shortly. 
nfortunatly, prevous implementation did not cover all byzantine error cases. In order to fix it we had to make each share proof (inclusion proof) independednt of global befp axis orientation. That why we have Axis inside each ShareWithProof now. The the provided code below from node does only the check for inclusion proof of each individual share. Some might have row proof, some might have col proof. The provided code is correct
```
	switch s.Axis {
	case rsmt2d.Row:
		rootHash = dah.RowRoots[axisIdx]
	case rsmt2d.Col:
		rootHash = dah.ColumnRoots[axisIdx]
	}
``` 

Proof construction is more complicated part and relies on rsmt2d implementation being able to find all byzantine errors. It looks like it is not part of lumina node, so if true, there is nothing to worry about. Just in case you are interesteed, one of the reason for celestia-node changes waas added more byzantine cases to rsmt2d: https://github.com/celestiaorg/rsmt2d/commit/d2da72509b6527acde213977b34346da643d2f7a

zvolin (approved):
I believe the implementation is correct :+1:

oblique:
Ok I was missing something. The implementation is indeed correct.

types/src/byzantine.rs
```diff
@@ -70,53 +63,61 @@ impl FraudProof for BadEncodingFraudProof {
...
-        };
-
-        // verify if the root can be converted to a cid and back
-        let mh = Multihash::wrap(NMT_MULTIHASH_CODE, &root.to_array())?;
-        let cid = Cid::new_v1(NMT_CODEC, mh);
-        let root = NamespacedHash::try_from(cid.hash().digest())?;
+        let shares_count = self.shares.iter().filter(|s| s.is_some()).count();
+        if shares_count < ods_width {
+            bail_validation!(
+                "fraud proof non-nil shares count ({shares_count}) < dah ods width ({ods_width})"
```
zvolin, line 97:
```suggestion
                "fraud proof present shares count ({shares_count}) < dah ods width ({ods_width})"
```
I think this error should be translated to rust, imagine seeing some error in Rust saying like 'value is non-nil' :p

types/src/byzantine.rs
```diff
@@ -346,13 +352,35 @@ pub(crate) mod test_utils {
...
         let mut shares_with_proof: Vec<_> = Vec::with_capacity(square_width.into());
 
         // collect the shares for fraud proof
         for share_idx in 0..square_width {
-            let (share, proof) = nmt.get_index_with_proof(share_idx.into());
+            let proof_axis = match rng.gen_range(0..=1) {
+                0 => AxisType::Row,
+                1 => AxisType::Col,
+                _ => unreachable!(),
+            };
```
zvolin, line 366: we could do this, nit but doesn't need unreachable branch
```suggestion
            let proof_axis = if rand::random() {
                AxisType::Row
            } else {
                AxisType::Col
            };
```

zvolin (approved):

PR #340, opened 2024-07-18
feat: updating API for parity with celestia-node v0.15.0

distractedm1nd:
Not ready for merge until node v0.15.0 is released, CI will also fail until then.

- Adds `TxConfig` for blob and state modules
- Removes GasPrice type and gas_prize_serializer
- Removes SubmitTx endpoint from state module


rpc/src/blob.rs
joroshiba, line 1: I believe still need to update `blob_get_all` to return an `Option<Vec<Blob>>` instead of `Vec<Blob>`. Not 100% positive but I know that with changes null responses currently throw serde parse errors, I believe this will fix.

zvolin: That means we're missing test for that

zvolin: added an option, tho it may be worth raising to celestia guys if they wouldn't want to switch to `[]`

distractedm1nd:
We're going to still need blob.Subscribe but that could also be a secondary PR

zvolin:
should be easy enough to add it right now

ci/run-validator.sh
```diff
@@ -135,22 +136,8 @@ setup_private_validator() {
...
-      0x966e6f22781EF6a6A82BBB4DB3df8E225DfD9488 \
-      --from "$NODE_NAME" \
-      --fees 30000utia \
-      -b block \
-      -y
-
-    echo "Registered validator's EVM address"
-  } &
+  # Set app version to 1
+  sed -i'.bak' 's|"app_version": "2"|"app_version": "1"|g' "$CONFIG_DIR/config/genesis.json"
```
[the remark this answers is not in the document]
zvolin: mainnet is still on app_version 1 and we haven't yet resolved network updates on header `.verify()`
https://github.com/eigerco/lumina/issues/368

fl0rek (approved):
LGTM 🚢

PR #341, opened 2024-07-18
feat(node-wasm)!: Add websocket support

zvolin:


node/src/p2p/swarm.rs
```diff
@@ -73,6 +74,14 @@ mod imp {
...
             })
             .expect("webtransport_websys::Transport is infallible")
+            .with_other_transport(|local_keypair| {
+                let noise_config = noise::Config::new(local_keypair)?;
+                Ok(websocket_websys::Transport::default()
+                    .upgrade(Version::V1)
+                    .authenticate(noise_config)
+                    .multiplex(yamux::Config::default()))
+            })
+            .map_err(|_| P2pError::InitNoise("websocket-websys".to_string()))?
```
oblique, line 84:
```suggestion
            .map_err(|e| P2pError::InitNoise(format!("websocket-websys: {e}")))?
```

node-wasm/src/worker.rs
```diff
@@ -239,7 +239,14 @@ impl NodeWorker {
...
     let (tx, mut rx) = mpsc::channel(WORKER_MESSAGE_SERVER_INCOMING_QUEUE_LENGTH);
-    let events_channel_name = format!("NodeEventChannel-{}", get_crypto()?.random_uuid());
+    let events_channel_name = if !is_safari() {
+        format!("NodeEventChannel-{}", get_crypto()?.random_uuid())
+    } else {
+        // TODO: handle uniqueness in Safari
+        // it doesn't have `crypto` module in global scope
+        // should not be an issue as long as we run shared worker there
+        "NodeEventChannel-1".to_string()
+    };
```
oblique, line 249: How `get_crypto()` behaved here? Did it returned an error?

node-wasm/src/utils.rs
```diff
@@ -186,15 +186,24 @@ pub(crate) async fn request_storage_persistence() -> Result<(), Error> {
...
 
 // Currently, there's an issue with SharedWorkers on Chrome where restarting Lumina's worker
 // causes all network connections to fail. Until that's resolved detect chrome and apply
 // a workaround.
 pub(crate) fn is_chrome() -> Result<bool, Error> {
-    get_navigator()?
-        .user_agent()
-        .context("could not get UserAgent from Navigator")
-        .map(|user_agent| user_agent.contains(CHROME_USER_AGENT_DETECTION_STR))
+    user_agent_contains(CHROME_USER_AGENT_DETECTION_STR)
```
oblique, line 202: I think we should cache the results with `OnceCell` or `OnceLock`

node-wasm/src/worker.rs
zvolin: Reflection fails because there is no crypto object in global scope

node/src/p2p/swarm.rs
zvolin: I'm not sure it's worth it as we call it just once, and cache won't cross workers boundary

node-wasm/src/utils.rs
zvolin: I'm not sure it's worth it as we call it just once, and cache won't cross workers boundary

node/src/p2p/kademlia.rs
```diff
@@ -0,0 +1,189 @@
...
+    type ToSwarm = kad::Event;
+
+    fn handle_pending_outbound_connection(
+        &mut self,
+        connection_id: ConnectionId,
+        maybe_peer: Option<PeerId>,
+        addresses: &[Multiaddr],
+        effective_role: Endpoint,
+    ) -> Result<Vec<Multiaddr>, ConnectionDenied> {
+        // The that `libp2p::kad::Behaviour` uses to flow of dialing to a peer is as follows:
```
fl0rek, line 36:
```suggestion
        // `libp2p::kad::Behaviour` uses the following flow to dial a peer:
```

node/src/p2p/kademlia.rs
```diff
@@ -0,0 +1,189 @@
...
+        //
+        // 1. Kad's Behaviour discovers a new peer and its addresses.
+        // 2. Kad's Behaviour dials with `DialOpts::peer_id()` without setting the addresses.
+        // 3. `Swarm` calls `Behaviour::handle_pending_outbound_connection` to resolve the addresses.
+        // 4. Kad's Behaviour returns a vector of the discovered addresses.
```
fl0rek, line 41:
```suggestion
        // 4. Kad's Behaviour returns the vector of discovered addresses.
```

node/src/p2p/swarm.rs
```diff
@@ -1,73 +1,190 @@
...
+        // that, the following edge cases can happen:
+        //
+        // 1. Machine connects to WiFi A. Our node starts and uses DNS that
+        //    WiFi A announced. Then machine moves to WiFi B, the old DNS servers
+        //    become unreachable, but libp2p still uses the old ones.
+        // 2. Machine is not connected to the Internet. Our node starts and does not
+        //    find any DNS servers defined. Then machine connects to the Internet, but
+        //    libp2p still does not have any DNS nameservers defined.
+        //
+        // By having a pre-defined public servers, these edge cases solved.
```
fl0rek, line 50:
```suggestion
        // We do not use system's DNS because libp2p caches system DNS servers when
        // `Swarm` get constructed, and doesn't update them later. This can be a problem,
        // if device roams between networks (and old DNS addresses may not be reachable from 
        // the new network). Similarly, if node is started when there's no Internet connection,
        // it won't use the DNS servers offered when Internet connectivity is restored.
        // Instead we per-define globally-accessible public DNS servers.
```
Reworded it to be slightly more concise, wdyt?

node-wasm/src/utils.rs
```diff
@@ -186,29 +194,50 @@ pub(crate) async fn request_storage_persistence() -> Result<(), Error> {
 }
 
 const CHROME_USER_AGENT_DETECTION_STR: &str = "Chrome/";
+const FIREFOX_USER_AGENT_DETECTION_STR: &str = "Firefox/";
+const SAFARI_USER_AGENT_DETECTION_STR: &str = "Safari/";
```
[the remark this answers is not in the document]
oblique: Are you talking about the version string that follows? Because I want to avoid regex.

node/src/p2p/swarm.rs
oblique: looks good

node-wasm/src/utils.rs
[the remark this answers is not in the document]
fl0rek: I was taking mostly about the tables with rules for browser detection (which strings it shoud/shouldn't contain). 

Also for websocket, we might want to do detection per-engine and match on webkit? Especially since once iOS browsers move off of webkit, they'll probably update user-agent to match and we'll pick that up automatically:
Screenshot 2024-07-30 at 10 03 58 AM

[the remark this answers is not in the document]
fl0rek: Not insisting on changing it here, but it'd be nice to have browser detection rules coming from some authoritative source, rather than some ad-hoc set of rules we've come up with.

fl0rek (approved):
🍏

PR #343, opened 2024-07-19
feat(node)!: Implement graceful shutdown

oblique:


node/src/p2p/header_ex/client.rs
```diff
@@ -143,45 +187,68 @@ where
...
+                    }
                 }
-            }
 
-            // Otherwise return the header with the maximum height
-            let resp = resps.into_iter().next().expect("no reposnes");
-            respond_to.maybe_send_ok(vec![resp]);
-        });
+                // Otherwise return the header with the maximum height
+                let resp = resps.into_iter().next().expect("no reposnes");
```
fl0rek, line 247:
```suggestion
                let resp = resps.into_iter().next().expect("no responses");
```

node/src/utils/token.rs
```diff
@@ -0,0 +1,138 @@
...
+            token: CancellationToken::new(),
+        }
+    }
+
+    /// Trigger the event
+    pub(crate) fn trigger(&self) {
+        self.token.cancel();
+    }
+
+    /// Returns a guard that will trigger the event on drop.
```
fl0rek, line 22:
```suggestion
    /// Returns a guard that will trigger the token on drop.
```

node/src/utils/token.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+
+    /// Returns if event is triggered or not.
+    #[allow(dead_code)]
+    pub(crate) fn is_triggered(&self) -> bool {
+        self.token.is_cancelled()
+    }
+
+    /// Returns when the event is triggered. If it was triggered before,
```
fl0rek, line 35:
```suggestion
    /// Returns when the token is triggered. If it was triggered before,
```
I think it's touch clearer when talking about just the token, without involving potential events user might stack on top of it.

zvolin (approved):
nice, looks solid :mechanical_arm:

fl0rek (approved):
LGTM!

PR #347, opened 2024-07-25
fix(node)!: Relax initialization if HEAD is the same as the stored one

oblique:


node/src/syncer.rs
```diff
@@ -611,9 +611,23 @@ where
...
 
-    // Insert HEAD to the store and initialize header-sub.
-    // Normal insertion checks still apply here.
-    store.insert(network_head.clone()).await?;
+    // If network head and store head have the same height, then
+    // `insert` will error because of insertion contraints.
+    // However, if both headers are the exactly the same, we
+    // relax this by avoiding insertion.
+    //
+    // This can happen of fast node restart.
```
fl0rek, line 619:
```suggestion
    // This can happen in case of fast node restart.
```

node/src/syncer.rs
```diff
@@ -611,9 +611,23 @@ where
 
     let network_head = p2p.get_head_header().await?;
 
-    // Insert HEAD to the store and initialize header-sub.
-    // Normal insertion checks still apply here.
-    store.insert(network_head.clone()).await?;
+    // If network head and store head have the same height, then
+    // `insert` will error because of insertion contraints.
+    // However, if both headers are the exactly the same, we
+    // relax this by avoiding insertion.
```
fl0rek, line 617:
```suggestion
    // If the network head and the store head have the same height,
    // then `insert` will error because of insertion contraints.
    // However, if both headers are the exactly the same, we
    // can skip inserting, as the header is already there.
```
I think it's more of a "skipping" than "relaxing"

fl0rek (approved):
nice catch! Small wording comment

fl0rek (approved):

PR #349, opened 2024-07-25
feat(node): Trigger dial on bootnodes when all peers disconnect

oblique:


fl0rek (approved):
short and sweet! :shipit:

Issue #242, opened 2024-03-08
Implement data and header pruning


zvolin:
![image](https://github.com/eigerco/lumina/assets/34972409/f1e06fbb-b8f6-4bbb-b3d7-c0ebe31adc7c)
Currently with around 1M headers (and no daser samples) we hit around 30Gb. It would be great to keep it lower

PR #351, opened 2024-07-25
feat: Header pruning

fl0rek:
related #242


fl0rek:
Touch messy with commits, contains changes in https://github.com/eigerco/lumina/pull/328, should get cleared once it's merged.

node/src/pruner.rs
```diff
@@ -0,0 +1,272 @@
...
+    async fn run(&mut self) -> Result<()> {
+        let mut last_reported = None;
+        let mut last_removed = None;
+        loop {
+            let pruning_window_end = Time::now().checked_sub(PRUNING_WINDOW).unwrap_or_else(|| {
+                warn!("underflow when computing pruning window start, defaulting to unix epoch");
+                Time::unix_epoch()
+            });
+
+            while let Some(header) = self.get_tail_header_to_prune(&pruning_window_end).await? {
```
oblique, line 124: After `retain` is implemented. It would make sense to do a binary search to find up to which height we should remove.

fl0rek: Per conversation during daily, we'll go with one-by-one removal.

node/src/pruner.rs
```diff
@@ -0,0 +1,403 @@
...
+
+            if last_reported != last_removed {
+                last_reported = last_removed;
+                self.event_pub.send(NodeEvent::PrunedHeaders {
+                    to_height: last_removed.expect("last removed height to be set"),
+                });
+            }
+
+            select! {
+                _ = self.cancellation_token.cancelled() => break,
```
zvolin, line 151: this can potentially block graceful shutdown for very long time, maybe we should also check this token periodically in the inner loop?

fl0rek: Added checking for cancellation in the inner loop between each header removal :+1:

node/src/pruner.rs
```diff
@@ -0,0 +1,407 @@
...
+                    .store
+                    .get_sampling_metadata(height)
+                    .await?
+                    .map(|m| m.cids)
+                    .unwrap_or_default();
+                for cid in cids {
+                    self.blockstore.remove(&cid).await?;
+                }
+
+                let removed = self.store.remove_last().await?;
```
zvolin, line 137: I know that those invariants are held by the 1h gap between syncing and pruning window but I don't like much that we aren't explicitely removing height for which we wiped cid's from blockstore

node/src/pruner.rs
```diff
@@ -0,0 +1,407 @@
...
+                debug_assert_eq!(header.height().value(), removed);
+
+                last_removed = Some(height);
+
+                if self.cancellation_token.is_cancelled() {
```
zvolin, line 142: nit, imo this check should be at the beginning of the loop. the only difference is that in some strange start + stop case, we may not prune any headers whereas now at least one will be pruned

node/src/pruner.rs
```diff
@@ -0,0 +1,407 @@
...
+            pruner_event,
+            NodeEvent::PrunedHeaders { to_height: 50 }
+        ));
+        assert_eq!(
+            store.get_stored_header_ranges().await.unwrap(),
+            new_block_ranges([51..=70])
+        );
+
+        // sleep needs to be longer than PRUNING_INTERVAL for pruner to wake up
+        sleep(Duration::from_secs(12)).await;
```
zvolin, line 365: can we add sth like `pruning_interval: Option<Duration>` to `PrunerArgs`, that will only default to the const when `None` is set, to get rid of this sleep?

zvolin (commented):
nice, left some ideas

node/src/pruner.rs
zvolin: I don't know if there is better way tho, we shouldn't make ranges discontinuous. Can we maybe convert the debug_assert below to hard error?

node/src/pruner.rs
fl0rek: I wasn't sure about parametrising it just for the sake of tests, but ultimately it might be better this way, added :+1:

node/src/pruner.rs
fl0rek: moved to the top, should work mostly the same.

node/src/pruner.rs
fl0rek: Yeah, we're silently assuming nobody will be messing with the header's tail other than pruner. Since I think we can rely on the syncing/pruning window difference,(so it shouldn't happen), added an error to detect above and halt pruner when we encounter it.

zvolin (approved):
nice

PR #355, opened 2024-07-31
fix(node): Anchor syncing on already existing header ranges

fl0rek:
Changes `calculate_range_to_fetch` to prefer extending existing range ether to the left or right instead of offering latest `limit` headers starting from the network head. This way, any inserts happening to the store during syncing are immediately verified against already existing, trusted range.


node/src/store/utils.rs
```diff
@@ -13,52 +14,42 @@ use crate::store::{SamplingMetadata, StoreError};
...
+        // if we haven't caught up with the network head, start from there
+        return Either::Right(store_head_range.end() + 1..=head_height);
     }
 
     // there exists a range contiguous with network head. inspect previous range end
     let penultimate_range_end = store_headers_iter.next().map(|r| *r.end()).unwrap_or(0);
 
-    penultimate_range_end + 1..=store_head_range.start().saturating_sub(1)
+    Either::Left(penultimate_range_end + 1..=store_head_range.start().saturating_sub(1))
 }
```
oblique, line 53: Instead of using `Either`, it makes sense to merge this into one function. For example `Either::Right(1..=head_height)` can become `(1..=head_height).truncate_right(limit)`

Something like this:

```rust
pub(crate) fn calculate_range_to_fetch(
    subjective_head_height: u64,
    store_headers: &[RangeInclusive<u64>],
    limit: u64,
) -> BlockRange {
    let mut store_headers_iter = store_headers.iter().rev();

    let Some(store_head_range) = store_headers_iter.next() else {
        // empty store, we're missing everything
        let range = 1..=subjective_head_height;
        return range.truncate_right(limit);
    };

    if *store_head_range.end() < subjective_head_height {
        // if we haven't caught up with the network head, start from there
        let range = store_head_range.end() + 1..=subjective_head_height;
        return range.truncate_right(limit);
    }

    // there exists a range contiguous with network head. inspect previous range end
    let penultimate_range_end = store_headers_iter.next().map(|r| *r.end()).unwrap_or(0);

    let range = penultimate_range_end + 1..=store_head_range.start().saturating_sub(1);
    range.truncate_left(limit)
}
```

Also I prefer this helper function to be in `syncer.rs`

node/src/syncer.rs
```diff
@@ -493,7 +494,6 @@ where
         let next_batch = calculate_range_to_fetch(
             subjective_head_height,
             store_ranges.as_ref(),
-            self.estimated_syncing_window_end,
```
oblique, line 496: Remove this field from the struct

node/src/syncer.rs
```diff
@@ -273,6 +273,7 @@ where
                     info!("Setting initial subjective head to {network_head_height}");
                     self.set_subjective_head_height(network_head_height);
 
+
```
oblique, line 276:
```suggestion
```

node/src/syncer.rs
fl0rek: Removed :+1:

node/src/store/utils.rs
fl0rek: I think you're right, now it's simple enough that it doesn't make sense to split this into two functions. Merged and moved to syncer :+1:

node/src/syncer.rs
```diff
@@ -594,6 +589,35 @@ where
...
 }
 
+/// based on the stored headers and current network head height, calculate range of headers that
+/// should be fetched from the network, anchored on already existing header range in store
+pub(crate) fn calculate_range_to_fetch(
+    subjective_head_height: u64,
+    store_headers: &[BlockRange],
+    //syncing_window_edge: Option<u64>,
+    limit: u64,
+) -> BlockRange {
```
oblique, line 599:
```suggestion
fn calculate_range_to_fetch(
    subjective_head_height: u64,
    store_headers: &[BlockRange],
    limit: u64,
) -> BlockRange {
```

node/src/syncer.rs
```diff
@@ -741,26 +835,26 @@ mod tests {
         handle_session_batch(&mut p2p_mock, &header_28_30, 28..=30, true).await;
```
fl0rek, line 835: TODO: add more comprehensive backwards syncing test

node/src/syncer.rs
```diff
@@ -712,63 +806,76 @@ mod tests {
...
         p2p_mock.expect_no_cmd().await;
 
-        // New HEAD was received by HeaderSub (height 30), it should NOT be appended
-        let header_28_30 = gen.next_many(3);
-        p2p_mock.announce_new_head(header_28_30[2].clone());
-        assert_syncing(&syncer, &store, &[1..=27], 30).await;
+        // New HEAD was received by HeaderSub (height 1506), it should NOT be appended
+        let headers_1503_1506 = gen.next_many(3);
+        p2p_mock.announce_new_head(headers_1503_1506[2].clone());
+        assert_syncing(&syncer, &store, &[1..=1502], 1505).await;
```
oblique, line 845:
```suggestion
        // New HEAD was received by HeaderSub (height 1505), it should NOT be appended
        let headers_1503_1506 = gen.next_many(3);
        p2p_mock.announce_new_head(headers_1503_1505[2].clone());
        assert_syncing(&syncer, &store, &[1..=1502], 1505).await;
```

node/src/syncer.rs
```diff
@@ -712,63 +806,76 @@ mod tests {
...
 
         // New HEAD is not adjacent to store, so Syncer requests a range
-        handle_session_batch(&mut p2p_mock, &header_28_30, 28..=30, true).await;
-        assert_syncing(&syncer, &store, &[1..=30], 30).await;
+        handle_session_batch(&mut p2p_mock, &headers_1503_1506, 1503..=1505, true).await;
```
oblique, line 848: Because of previous suggestion

```suggestion
        handle_session_batch(&mut p2p_mock, &headers_1503_1505, 1503..=1505, true).await;
```

node/src/syncer.rs
```diff
@@ -712,63 +806,76 @@ mod tests {
...
+        p2p_mock.announce_new_head(header1502.clone());
+        assert_syncing(&syncer, &store, &[1..=1502], 1502).await;
         p2p_mock.expect_no_cmd().await;
 
-        // New HEAD was received by HeaderSub (height 30), it should NOT be appended
-        let header_28_30 = gen.next_many(3);
-        p2p_mock.announce_new_head(header_28_30[2].clone());
-        assert_syncing(&syncer, &store, &[1..=27], 30).await;
+        // New HEAD was received by HeaderSub (height 1505), it should NOT be appended
+        let headers_1503_1506 = gen.next_many(3);
```
oblique, line 843:
```suggestion
        let headers_1503_1505 = gen.next_many(3);
```

oblique (approved):

PR #358, opened 2024-08-08
chore(node): Update rexie and related wasm dependencies

fl0rek:
Upgrade rexie and related web dependencies and add required code changes.


tools/upgrade-deps.sh
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
fl0rek, line 1: this should make this script more portable, `bash` isn't always in `/bin`

node/src/store/indexed_db_store.rs
oblique, line 1: There are many changes for rexie. I prefer them in a separate PR.

fl0rek: Reverted to just rexie and friends, will open separate one for generic deps upgrade

zvolin (approved):
nice

zvolin:
@fl0rek I just noticed it's out of date and likely conflicts with update of libp2p

zvolin (approved):

PR #377, opened 2024-09-04
feat(node-wasm)!: Webpack compatibility

### LINKED ISSUE #365: Try removing requirement for providing custom worker script

fl0rek:
closes #365


lumina-node-shim/worker.js
```diff
@@ -0,0 +1,24 @@
+Error.stackTraceLimit = 99;
```
zvolin, line 1: as we anyway do a glueing npm package, can we split this into `index.js` and `worker.js`?

lumina-node-shim/worker.js
```diff
@@ -0,0 +1,24 @@
...
+
+import init, { NodeWorker, NodeClient } from "lumina-node-wasm"
+
+export default async function start_worker() {
```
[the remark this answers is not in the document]
fl0rek: Wanted to maintain the structure similar to what wasm-pack was serving, but not too attached to it :+1:

lumina-node-shim/worker.js
fl0rek: What exactly would be the split? 
The worker code (inside the if) would be in worker, but I think the function starting worker should land there too, to ensure paths not going out of sync on any file moves.

lumina-node-shim/README.md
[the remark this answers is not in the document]
fl0rek: Good call, lets not pollute the root

lumina-node-shim/worker.js
zvolin: in `index.js` the `spawnNode` and `export * from "lumina-node-wasm";`.

Worker script can then drop whole `if` because it's just a hack to re-use the module:
```js
import init, { NodeWorker } from "lumina-node-wasm"

Error.stackTraceLimit = 99;

init().then(async () => {
  let worker = new NodeWorker();
  console.log("starting worker: ", worker);
  worker.connect(self)

  while (true) {
    await worker.poll();
  }
});
```

zvolin: I think it's unlikely the path will get invalidated as this shim would then just have only `index.js` and `worker.js`

fl0rek: Nice, two birds with once stone :rocket: split the script

node-wasm/js/index.js
```diff
@@ -7,3 +7,4 @@ export async function spawnNode() {
     return client;
 }
 
+export * from "lumina-node-wasm";
```
zvolin, line 10: this apparently doesn't export the default export from lumina-node-wasm. If we want users to be able to just initialize wasm without using `spawnNode`, we must add
```suggestion
export * from "lumina-node-wasm";
export default init;
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,250 @@
...
+            .context("failed to setup port for ClientConnection")?;
+
+        Ok(ClientConnection {
+            port,
+            _onmessage: onmessage,
+        })
+    }
+
+    fn send(&self, message: &WorkerResponse) -> Result<()> {
+        let message_value = to_value(message).context("could not serialise message")?;
```
zvolin, line 69: So in extension template we currently need to sanitize messages.
```
// https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Chrome_incompatibilities#data_cloning_algorithm
// Chrom* uses JSON serialisation algoritm for messges sent over runtime.Port, which strips keys with a value of `undefined`.
// This mangles some of lumina responses, e.g. successful start node response is `{ NodeStarted: { Ok : undefined } }`, which
// then becomes `{ NodeStarted: {} }`.
// Thankfully, replacing `undefined` with `null` works as a workaround, since we don't care about key value anyway
```

We could migrate this comment and solve it here by using
```rs
use serde_wasm_bindgen::Serializer;
use serde::Serialize;
 ...
let serializer = Serializer::default().serialize_missing_as_null(true);
let message_value = message.serialize(&serializer).context("could not serialise message")?;
```

fl0rek: Nice, didn't know about it! Switched to using `Serializer::json_compatible`

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,254 @@
...
+                    let client_id = ClientId(self.ports.len());
+                    info!("Connecting client {client_id:?}");
+
+                        match ClientConnection::new(client_id, port, self._request_channel_tx.clone()) {
+                            Ok(port) =>
+                    self.ports.push(port),
+                    Err(e) => {
+                        error!("Failed to setup ClientConnection: {e}");
+                    }
+                        }
```
oblique, line 116:
```suggestion
                    match ClientConnection::new(client_id, port, self._request_channel_tx.clone()) {
                        Ok(port) => self.ports.push(port),
                        Err(e) => error!("Failed to setup ClientConnection: {e}"),
                    }
```

node-wasm/src/worker.rs
```diff
@@ -48,12 +41,88 @@ pub enum WorkerError {
...
+
+    pub async fn connect(&self, port: JsValue) {
+        self.control_channel
+            .send(port)
+            .await
+            .expect("RequestServer command channel should never close")
+    }
+
+    #[wasm_bindgen(js_name = runWorker)]
+    pub async fn run_worker(&self) -> Result<(), Error> {
```
oblique, line 81: You can use `&mut self` here and remove mutexes.

node-wasm/js/README.md
```diff
@@ -0,0 +1,19 @@
...
+A compatibility layer for the [`Lumina`](https://github.com/eigerco/lumina) node to
+work within a browser environment and be operable with javascript.
+
+# Example
+Starting lumina inside a dedicated worker
+
+```javascript
+import init, { Node, NodeConfig, Network } from "lumina-node";
+
+const node = await init();
```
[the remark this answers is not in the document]
zvolin: It's outdated. Before my previous comments spawnNode was a default export, used as init here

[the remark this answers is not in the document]
zvolin: https://github.com/eigerco/lumina/pull/377#discussion_r1751878092

node-wasm/js/index.js
fl0rek: Didn't know about it, changed!

node-wasm/js/README.md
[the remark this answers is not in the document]
fl0rek: Yes, it's outdated. Updated the example here to use new names and imports

node-wasm/js/README.md
```diff
@@ -0,0 +1,19 @@
...
+const config = NodeConfig.default(Network.Mainnet);
+
+await node.start(config);
+
+await node.wait_connected();
+await node.request_head_header();
+```
```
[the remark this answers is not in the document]
fl0rek: fixed, see https://github.com/eigerco/lumina/pull/377#discussion_r1758219910

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,254 @@
...
+            .post_message(&message_value)
+            .context("could not send command to worker")?;
+        Ok(())
+    }
+}
+
+pub struct RequestServer {
+    ports: Vec<ClientConnection>,
+    control_channel: mpsc::Receiver<JsValue>,
+    _request_channel_tx: mpsc::Sender<(ClientId, Result<NodeCommand, TypedMessagePortError>)>,
```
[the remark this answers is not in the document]
fl0rek: fixed

node-wasm/src/ports.rs
fl0rek: cargo fmt got a bit confused here, fixed

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,254 @@
...
+
+impl ClientConnection {
+    fn new(
+        id: ClientId,
+        object: JsValue,
+        forward_to: mpsc::Sender<(ClientId, Result<NodeCommand, TypedMessagePortError>)>,
+    ) -> Result<Self> {
+        let onmessage = Closure::new(move |ev: MessageEvent| {
+            let response_tx = forward_to.clone();
+            spawn_local(async move {
```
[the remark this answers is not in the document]
fl0rek: True, we're effectively had bounded queue and unbounded queue of tasks. Switched to just unbounded queue.

node-wasm/src/worker.rs
fl0rek: Originally I've added this to avoid runtime errors about js violating rust borrow rules (since `NodeWorker` would be owned by js). I expected `run_workers` and `connect` being called from callback to clash, but I can't seem to trigger it with current implementation based on channels. 
So switching to mut as suggested :+1:

node-wasm/src/ports.rs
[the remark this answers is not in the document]
oblique: Remove also `spawn_local`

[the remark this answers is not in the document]
oblique: and change other places with the same pattern

node-wasm/src/worker.rs
```diff
@@ -48,12 +39,88 @@ pub enum WorkerError {
     NodeError(Error),
 }
 
+#[wasm_bindgen]
 struct NodeWorker {
+    event_channel_name: String,
+    worker: Option<NodeWorkerInstance>,
+    request_server: Mutex<RequestServer>,
```
oblique, line 46: `Mutex` can be removed from here as well

node-wasm/src/client.rs
```diff
@@ -35,91 +33,42 @@ pub struct WasmNodeConfig {
...
-/// `NodeDriver` represents lumina node running in a dedicated Worker/SharedWorker.
-/// It's responsible for sending commands and receiving responses from the node.
-#[wasm_bindgen(js_name = NodeClient)]
-struct NodeDriver {
-    client: WorkerClient,
-}
-
-/// Type of worker to run lumina in. Allows overriding automatically detected worker kind
-/// (which should usually be appropriate).
+/// `NodeDriver` is responsible for steering [`NodeWorker`] by sending it commands and receiving
```
zvolin, line 36:
```suggestion
/// `NodeClient` is responsible for steering [`NodeWorker`] by sending it commands and receiving
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+use wasm_bindgen::prelude::*;
+use wasm_bindgen::{JsCast, JsValue};
+use wasm_bindgen_futures::spawn_local;
+use web_sys::{MessageEvent, MessagePort};
+
+use crate::commands::{NodeCommand, WorkerResponse};
+use crate::error::{Context, Error, Result};
+
+// Instead of just supporting communicaton with just `MessagePort`, allow using any object which
+// provides compatible interface
```
zvolin, line 17:
```suggestion
// Instead of supporting communication with just `MessagePort`, allow using any object which
// provides compatible interface, eg. `Worker`
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+#[wasm_bindgen]
+extern "C" {
+    pub type MessagePortLike;
+
+    #[wasm_bindgen (catch , method , structural , js_name = postMessage)]
```
zvolin, line 22:
```suggestion
    #[wasm_bindgen(catch, method, structural, js_name = postMessage)]
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+    pub fn post_message(this: &MessagePortLike, message: &JsValue) -> Result<(), JsValue>;
+
+    #[wasm_bindgen (catch , method , structural , js_name = postMessage)]
```
zvolin, line 25:
```suggestion
    #[wasm_bindgen(catch, method, structural, js_name = postMessage)]
```

node-wasm/src/ports.rs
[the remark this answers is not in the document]
zvolin: should spawn and clones be removed then?

node-wasm/js/README.md
```diff
@@ -0,0 +1,19 @@
+# Lumina node wasm
+
+A compatibility layer for the [`Lumina`](https://github.com/eigerco/lumina) node to
+work within a browser environment and be operable with javascript.
+
+# Example
+Starting lumina inside a dedicated worker
+
+```javascript
+import { spawnNode, Node, NodeConfig, Network } from "lumina-node";
```
zvolin, line 10:
```suggestion
import { spawnNode, NodeConfig, Network } from "lumina-node";
```

node-wasm/js/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+const node = await spawnNode();
+const mainnetConfig = NodeConfig.default(Network.Mainnet);
+
+await node.start(mainnetConfig);
+
+await node.wait_connected();
+await node.request_head_header();
+```
```
zvolin, line 19: I think we should mention there that if someone doesn't use the `spawnNode` then he must call init to initialize wasm module, which is a default export.

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+
+struct ClientConnection {
+    port: MessagePortLike,
+    _onmessage: Closure<dyn Fn(MessageEvent)>,
+}
+
+impl ClientConnection {
+    fn new(
+        id: ClientId,
+        object: JsValue,
```
zvolin, line 50: is object a port? can you add some meaningful name here?

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+                if let Err(e) = response_tx.send(message).await {
+                    error!("message forwarding channel closed, should not happen: {e}");
+                }
+            })
+        });
+
+        let port = prepare_message_port(object, &onmessage)
+            .context("failed to setup port for RequestResponse")?;
+
+        web_sys::console::log_1(&port);
```
zvolin, line 186: in case this is needed and not for debugging, then please attach some message and use tracing

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+            let channel = MessageChannel::new().unwrap();
+
+            tx.send(channel.port2().into()).unwrap();
+
+            let client0 = RequestResponse::new(channel.port1().into()).unwrap();
+
+
+            //let (tx, rx) = mpsc::channel(10);
+            //tx.send(channel0.port2().into()).await.unwrap();
+
```
zvolin, line 303:
```suggestion
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+    Ok(MessagePortLike::from(object))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use wasm_bindgen_test::{wasm_bindgen_test, wasm_bindgen_test_configure};
+    use web_sys::MessageChannel;
+
+    wasm_bindgen_test_configure!(run_in_browser);
```
zvolin, line 283: can we move it to `lib.rs` to not call it twice accidentally in future?

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    #[wasm_bindgen_test]
+    async fn client_server() {
+        let mut server = RequestServer::new();
+
+        let tx = server.get_connect_channel();
+
+
```
zvolin, line 291:
```suggestion
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+            let client0 = RequestResponse::new(channel.port1().into()).unwrap();
+
+
+            //let (tx, rx) = mpsc::channel(10);
+            //tx.send(channel0.port2().into()).await.unwrap();
+
+            let response = client0.exec(NodeCommand::IsRunning).await.unwrap();
+            assert!(matches!(response, WorkerResponse::IsRunning(true)));
+
```
zvolin, line 306:
```suggestion
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,314 @@
...
+        });
+
+        let (client, command) = server.recv().await;
+        assert!(matches!(command.unwrap(), NodeCommand::IsRunning));
+        server.respond_to(client, WorkerResponse::IsRunning(true));
+
```
zvolin, line 312:
```suggestion
```

zvolin (commented):
nice, some pass from me

node-wasm/src/ports.rs
fl0rek: My idea was to accept just an object and figure out whether the interface is ok with `Reflect`, to be most compatible. `prepare_message_port` later sanitizes the input and gives you `port` back.

zvolin: so maybe `maybe_port`? just to help me understand :p

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,306 @@
...
+                },
+                connection = self.connect_rx.recv() => {
+                    let port = connection.expect("command channel should not close ?");
+                    let client_id = ClientId(self.ports.len());
+                    info!("Connecting client {client_id:?}");
+
+                        match ClientConnection::new(client_id, port, self.request_tx.clone(), self.connect_tx.clone()) {
+                            Ok(port) => self.ports.push(port),
+                            Err(e) => error!("Failed to setup ClientConnection: {e}"),
+                        }
```
zvolin, line 137:
```suggestion
                    match ClientConnection::new(client_id, port, self.request_tx.clone(), self.connect_tx.clone()) {
                        Ok(port) => self.ports.push(port),
                        Err(e) => error!("Failed to setup ClientConnection: {e}"),
                    }
```

node-wasm/src/ports.rs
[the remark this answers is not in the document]
oblique: yes

node-wasm/src/ports.rs
oblique: `maybe_port` is confusing. what about `raw_port`?

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,306 @@
...
+            .serialize(&serializer)
+            .context("could not serialise message")?;
+        self.port
+            .post_message(&message_value)
+            .context("could not send command to worker")?;
+        Ok(())
+    }
+}
+
+pub struct RequestServer {
```
zvolin, line 101: I'm wondering about renaming `RequestServer` and `RequestResponse` to `WorkerServer` and `WorkerClient` respectively. initially I had a bit troubles to understand which part is for what. wdyt?

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    pub async fn recv(&mut self) -> (ClientId, Result<NodeCommand, TypedMessagePortError>) {
+        loop {
+            select! {
+                message = self.request_rx.recv() => {
+                    return message.expect("request channel should never close");
+                },
+                connection = self.connect_rx.recv() => {
+                    let port = connection.expect("command channel should not close ?");
```
zvolin, line 130: just aligning with previous one
```suggestion
                    let port = connection.expect("command channel should never close");
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,306 @@
...
+        RequestServer {
+            ports: vec![],
+            connect_tx,
+            connect_rx,
+            request_tx,
+            request_rx,
+        }
+    }
+
+    pub async fn recv(&mut self) -> (ClientId, Result<NodeCommand, TypedMessagePortError>) {
```
zvolin, line 123: I think we should just return `(ClientId, NodeCommand)` here. At the call site we just do:
```rs
        let (client_id, result) = server.recv().await;

        let command = result
            .with_context(|| format!("could not parse command received from {client_id:?}"))?;
```
so we could move the error reporting here and simplify types around

node-wasm/src/lib.rs
```diff
@@ -1,8 +1,13 @@
...
+pub mod client;
+mod commands;
 pub mod error;
-pub mod node;
+mod ports;
 pub mod utils;
 mod worker;
 mod wrapper;
+
+#[cfg(all(target_arch = "wasm32", test))]
```
zvolin, line 12: this crate is completely empty in non-wasm anyway :p
```suggestion
#[cfg(test)]
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn new(
+        id: ClientId,
+        object: JsValue,
+        forward_messages_to: mpsc::UnboundedSender<(ClientId, Result<NodeCommand, MessageError>)>,
+        forward_connects_to: mpsc::UnboundedSender<JsValue>,
+    ) -> Result<Self> {
+        let onmessage = Closure::new(move |ev: MessageEvent| {
+            let message_tx = forward_messages_to.clone();
+            let port_tx = forward_connects_to.clone();
+
```
zvolin, line 57: this is a leftover

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,287 @@
...
+    response_channel: Mutex<mpsc::UnboundedReceiver<Result<WorkerResponse, MessageError>>>,
+    _onmessage: Closure<dyn Fn(MessageEvent)>,
+}
+
+impl WorkerClient {
+    pub fn new(object: JsValue) -> Result<Self> {
+        let (tx, rx) = mpsc::unbounded_channel();
+
+        let onmessage = Closure::new(move |ev: MessageEvent| {
+            let response_tx = tx.clone();
```
zvolin, line 160: this also isn't needed

node-wasm/src/worker.rs
```diff
@@ -48,12 +39,77 @@ pub enum WorkerError {
     NodeError(Error),
 }
 
+#[wasm_bindgen]
 struct NodeWorker {
+    event_channel_name: String,
+    worker: Option<NodeWorkerInstance>,
```
zvolin, line 45: could we actually rename it to `node`? worker in worker is a bit confusing

node-wasm/src/ports.rs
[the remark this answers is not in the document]
fl0rek: Removed spawns and clones :+1:

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,306 @@
...
+    ) -> Result<Self> {
+        let onmessage = Closure::new(move |ev: MessageEvent| {
+            let message_tx = forward_messages_to.clone();
+            let port_tx = forward_connects_to.clone();
+            spawn_local(async move {
+                let message: Result<NodeCommand, _> =
+                    from_value(ev.data()).map_err(TypedMessagePortError::FailedToConvertValue);
+
+                let ports = ev.ports();
+                if Array::is_array(&ports) {
```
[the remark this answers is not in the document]
fl0rek: Moved to utils under `MessageEventExt`

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,306 @@
...
+            error!("Failed to send response to client: {e}");
+        }
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum TypedMessagePortError {
+    #[error("Could not convert JsValue: {0}")]
+    FailedToConvertValue(serde_wasm_bindgen::Error),
+}
```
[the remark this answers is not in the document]
fl0rek: Moved to helper in utils under `MessageEventExt`

node-wasm/src/worker.rs
fl0rek: sure, renamed :+1:

node-wasm/src/ports.rs
fl0rek: removed

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,283 @@
...
+            .context("could not send command to worker")?;
+        Ok(())
+    }
+}
+
+pub struct WorkerServer {
+    ports: Vec<ClientConnection>,
+    connect_tx: mpsc::UnboundedSender<JsValue>,
+    connect_rx: mpsc::UnboundedReceiver<JsValue>,
+    request_tx: mpsc::UnboundedSender<(ClientId, Result<NodeCommand, MessageError>)>,
```
zvolin, line 93: I thought that after the signature change of the stuff, we could simplify this to `mpsc::UnboundedSender<(ClientId, NodeCommand)>`. would that be possible? Could we just log the error at onmessage level and not pass it further?

fl0rek: I think we can just ignore the un-deserialisable messages in the callback for the server side :+1: 
I've left client side as is, with how `exec` is implemented, it waits for some kind of response error or not.

node-wasm/src/ports.rs
fl0rek: Moved error logging into callback and skipped the error passing via channels as suggested :+1:

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,286 @@
...
+    port: MessagePortLike,
+    _onmessage: Closure<dyn Fn(MessageEvent)>,
+}
+
+impl ClientConnection {
+    fn new(
+        id: ClientId,
+        object: JsValue,
+        forward_messages_to: mpsc::UnboundedSender<(ClientId, NodeCommand)>,
+        forward_connects_to: mpsc::UnboundedSender<JsValue>,
```
oblique, line 52: Since both sender are constructed from `WorkerServer` and their receivers are handled in the same function. Isn't it better to have a single enum for them?

fl0rek: Good idea, switched to single channel :+1:

node-wasm/src/worker.rs
```diff
@@ -45,7 +45,7 @@ struct NodeWorker {
     event_channel_name: String,
     node: Option<NodeWorkerInstance>,
     request_server: WorkerServer,
-    connect_channel: mpsc::UnboundedSender<JsValue>,
+    control_channel: mpsc::UnboundedSender<ClientMessage>,
```
zvolin, line 48: maybe we could just have `.send()` on `WorkerServer` :thinking:

zvolin: tho it could be confusing, like one could expect `.send()` to send to client

node-wasm/src/ports.rs
fl0rek: I'm trying to capture the spirit of "something that is a port or looks enough like a port for us to use it". I'm not sure about `maybe_port` or `raw_port` as they miss the ducktyping aspect of the situation. 
How about `maybe_port` (which coincidentally matches the `MessagePortLike` type).

node-wasm/js/README.md
fl0rek: Added note and example below 👍

node-wasm/js/README.md
```diff
@@ -0,0 +1,31 @@
...
+Note that `spawnNode` implicitly calls wasm initialisation code. If you want to set things up manually, make sure to call the default export before using any of the wasm functionality.
+
+```javascript
+import init, { NodeConfig, Network } from "lumina-node";
+
+await init();
+const config = NodeConfig.default(Network.Mainnet);
+
+console.log(config.bootnodes);
+
```
oblique, line 30: can you add also how to create the Node etc?

node-wasm/js/index.js
```diff
@@ -0,0 +1,11 @@
+import init, { NodeClient } from "lumina-node-wasm"
+
+export async function spawnNode() {
```
zvolin, line 3:
```suggestion
/**
* Spawn a worker running lumina node and get the `NodeClient` connected to it.
*/
export async function spawnNode() {
```

node-wasm/js/index.js
```diff
@@ -0,0 +1,11 @@
+import init, { NodeClient } from "lumina-node-wasm"
```
zvolin, line 1: please also add `index.d.ts` file with such content (I hope it's correct way to do it, but not 100% sure :p) and add it to list of files in `package.json`
```ts
/**
* Spawn a worker running lumina node and get the `NodeClient` connected to it.
*/
export function spawnNode(): Promise<NodeClient>;
export * from "lumina-node-wasm";
export default init;
```

zvolin (approved):
nice :+1:

node-wasm/js/README.md
zvolin: do you mean the worker?

node-wasm/js/README.md
```diff
@@ -0,0 +1,41 @@
...
+await init();
+const config = NodeConfig.default(Network.Mainnet);
+const worker = new NodeWorker();
+
+const channel = new MessageChannel();
+worker.connect(channel.port1);
+const client = await new NodeClient(channel.port2);
+
+// note that this runs lumina in the current context (and not in a worker, like `spawnNode`). runWorker doesn't return.
```
zvolin, line 36:
```suggestion
// note that this runs lumina in the current context (and not in a web-worker, like `spawnNode`). runWorker doesn't return.
```
to specify what worker exactly we mean :p

oblique (approved):
LGTM

node-wasm/js/package.json
```diff
@@ -0,0 +1,30 @@
...
+    ],
+    "description": "Lumina node for Celestia, running in browser",
+    "version": "0.2.0",
+    "license": "Apache-2.0",
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/eigerco/lumina.git"
+    },
+    "files": [
+        "index.js",
```
zvolin, line 15:
```suggestion
        "index.js",
        "index.d.ts",
```

node-wasm/js/README.md
fl0rek: added :+1:

node-wasm/README.md
```diff
@@ -3,14 +3,40 @@
...
+import init, { NodeConfig, Network } from "lumina-node";
+
+await init();
+const config = NodeConfig.default(Network.Mainnet);
+
+// client and worker accept any object with MessagePort like interface e.g. Worker
+const channel = new MessageChannel();
+const worker = new NodeWorker(channel.port1);
+
+// note that this runs lumina in the current context (and doesn't create a new web-worker). runWorker doesn't return.
```
zvolin, line 35:
```suggestion
// note that this runs lumina in the current context (and doesn't create a new web-worker). Promise created with `.run()` never completes.
```

zvolin (approved):
:rocket:

oblique (approved):
🦇

PR #379, opened 2024-09-12
feat(node): adding agent version

guillaumemichel:
At [ProbeLab](https://probelab.io/) we are measuring the Celestia network and producing [weekly reports](https://probelab.io/celestia/dht/2024-36/). We would like to include numbers on light nodes, however we noticed that Lumina nodes don't advertise an agent version through the libp2p identify protocol. So we cannot distinguish Lumina nodes, and we would like to include them in our statistics.

I added just the basic user agent, feel free to complete the PR and add more information, such as version number, build etc. For inspiration [`celestia-node`](https://github.com/celestiaorg/celestia-node) uses the following user agent format `celestia-node/celestia/light/v0.16.0/6744f64`.

### DIFF
--- a/node/src/p2p.rs
+++ b/node/src/p2p.rs
@@ -614,8 +614,7 @@ where
 
         let identify = identify::Behaviour::new(
             identify::Config::new(String::new(), args.local_keypair.public())
-                // TODO: add version number
-                .with_agent_version(String::from("lumina")),
+                .with_agent_version(format!("lumina/{}", env!("CARGO_PKG_VERSION"))),
         );
 
         let header_sub_topic = gossipsub_ident_topic(&args.network_id, "/header-sub/v0.0.1");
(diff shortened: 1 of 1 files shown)


node/src/p2p.rs
```diff
@@ -612,10 +612,11 @@ where
...
         let ping = ping::Behaviour::new(ping::Config::default());
 
-        let identify = identify::Behaviour::new(identify::Config::new(
-            String::new(),
-            args.local_keypair.public(),
-        ));
+        let identify = identify::Behaviour::new(
+            identify::Config::new(String::new(), args.local_keypair.public())
+                // TODO: add version number
+                .with_agent_version(String::from("lumina")),
```
oblique, line 618:
```suggestion
                .with_agent_version(format!("lumina/{}", env!("CARGO_PKG_VERSION"))),
```

oblique (approved):
LGTM. I added the crate version.

fl0rek (approved):
nice, thanks @guillaumemichel!

zvolin (approved):
Added also network name so the pattern now is `lumina/{network}/{version}`

oblique (approved):

PR #383, opened 2024-09-19
feat(node-wasm)!: Align JS api to use camelCase, allow using plain numbers as arguments

fl0rek:


node-wasm/src/wrapper/libp2p.rs
```diff
@@ -4,25 +4,33 @@ use libp2p::swarm::{
...
     pub num_peers: usize,
-    connection_counters: ConnectionCountersSnapshot,
+    #[wasm_bindgen(js_name = "connectionCounters")]
+    pub connection_counters: ConnectionCountersSnapshot,
 }
 
-#[wasm_bindgen]
-#[derive(Debug, Serialize, Deserialize)]
+#[wasm_bindgen(inspectable)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
```
oblique, line 17: How `inspectable` works? Can we remove `serde-wasm-bindgen` and use only that?

fl0rek: > By default, structs exported from Rust become JavaScript classes with a single ptr property. All other properties are implemented as getters, which are not displayed when calling toJSON.

So I guess this primarily affects serialisation to json and debugging.

this is with `inspect`, you don't get the `toJSON` fn at all otherwise
image

zvolin (approved):

oblique (approved):

PR #400, opened 2024-09-27
remove protoc dep

S1nus:
This is a hideous hack to remove `protoc` build dependency. Do not merge, but we need to solve this


.DS_Store
fl0rek, line 1: please remove

oblique:
I don't really like this, but if we gonna do it I really prefer move the `build.rs` code in a crate called `proto-gen` that will have [publishing disabled](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field). And that crate must set the [`prost_build::Config::out_dir`](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.out_dir) to `proto/src/out`.

dimazhornyk:
+1 to the importance of this, dependency on `protoc` means that anyone who imports your crates has to have protoc installed, which is quite unnecessary in most cases
e.g. our CI when importing your crate: https://github.com/matter-labs/zksync-era/actions/runs/11112479333/job/30874671258?pr=2983#step:8:739

the option that @oblique suggested is the easiest one, but you might also consider [protox](https://github.com/andrewhickman/protox), it's a great alternative that lets you avoid the `protoc` dependency

S1nus:
Protox looks like an easy compromise, https://github.com/eigerco/lumina/pull/402, let me know thoughts on 402

PR #402, opened 2024-09-30
use protox

S1nus:
Easy way to remove the `protoc` build dependency is to use [protox](https://github.com/andrewhickman/protox)

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -22,6 +22,7 @@ serde = { version = "1.0.203", features = ["derive"] }
 [build-dependencies]
 anyhow = "1.0.86"
 prost-build = "0.12.6"
+protox = "0.6.1"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen-test = "0.3.42"
--- a/proto/build.rs
+++ b/proto/build.rs
@@ -64,6 +64,24 @@ static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
 ];
 
 fn main() -> Result<()> {
+
+    println!("protox::compile_fds...");
+    let fds = protox::compile(&[
+        "vendor/celestia/da/data_availability_header.proto",
+        "vendor/celestia/blob/v1/tx.proto",
+        "vendor/header/pb/extended_header.proto",
+        "vendor/share/eds/byzantine/pb/share.proto",
+        "vendor/share/shwap/pb/shwap.proto",
+        "vendor/share/shwap/p2p/bitswap/pb/bitswap.proto",
+        "vendor/cosmos/base/v1beta1/coin.proto",
+        "vendor/cosmos/base/abci/v1beta1/abci.proto",
+        "vendor/cosmos/crypto/multisig/v1beta1/multisig.proto",
+        "vendor/cosmos/staking/v1beta1/query.proto",
+        "vendor/cosmos/tx/v1beta1/tx.proto",
+        "vendor/go-header/p2p/pb/header_request.proto",
+    ], &["vendor", "vendor/nmt"]).expect("protox faild to build");
+    println!("protox done");
+
     let mut config = prost_build::Config::new();
 
     for (type_path, attr) in CUSTOM_TYPE_ATTRIBUTES {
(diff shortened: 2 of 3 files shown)


fl0rek:
@S1nus could you also update the build instructions in README, since protoc isn't needed anymore? I think it'll just be "# install dependencies" line

fl0rek:
@S1nus could you also update the build instructions in README, since protoc isn't needed anymore? I think it'll just be "# install dependencies" line

zvolin (commented):
nice, could you also remove protoc installation in `.github/workflows`?

proto/build.rs
```diff
@@ -64,6 +64,27 @@ static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
...
+            "vendor/cosmos/base/abci/v1beta1/abci.proto",
+            "vendor/cosmos/crypto/multisig/v1beta1/multisig.proto",
+            "vendor/cosmos/staking/v1beta1/query.proto",
+            "vendor/cosmos/tx/v1beta1/tx.proto",
+            "vendor/go-header/p2p/pb/header_request.proto",
+        ],
+        ["vendor", "vendor/nmt"],
+    )
+    .expect("protox faild to build");
+    println!("protox done");
```
zvolin, line 86:
```suggestion
```

proto/build.rs
```diff
@@ -64,6 +64,27 @@ static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
 ];
 
 fn main() -> Result<()> {
+    println!("protox::compile_fds...");
```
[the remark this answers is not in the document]
fl0rek: I think it'll be visible with enough verbose flags passed, or in case of failed build. 
I initially thought it was some kind of custom build instruction, [similar to `cargo::`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) ones, but I don't believe it's that? 
Either way I think output from protox/proto build should be enough, +1 for removing it.

proto/Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde = { version = "1.0.203", features = ["derive"] }
 [build-dependencies]
 anyhow = "1.0.86"
 prost-build = "0.12.6"
+protox = "0.6.1"
```
fl0rek, line 25: I've noticed that latest protox version is 0.7.1. Could you update, or is there a particular reason to not use it?

S1nus: That would require upgrading Prost, which breaks the build. I assumed you guys intentionally locked prost to this version

fl0rek (approved):
Nice! 
Thanks @S1nus, I can finally uninstall protobuf from my machine :D

proto/build.rs
```diff
@@ -64,6 +64,25 @@ static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
...
+            "vendor/cosmos/base/v1beta1/coin.proto",
+            "vendor/cosmos/base/abci/v1beta1/abci.proto",
+            "vendor/cosmos/crypto/multisig/v1beta1/multisig.proto",
+            "vendor/cosmos/staking/v1beta1/query.proto",
+            "vendor/cosmos/tx/v1beta1/tx.proto",
+            "vendor/go-header/p2p/pb/header_request.proto",
+        ],
+        ["vendor", "vendor/nmt"],
+    )
+    .expect("protox faild to build");
```
oblique, line 84:
```suggestion
    .expect("protox failed to build");
```

proto/build.rs
```diff
@@ -87,23 +106,8 @@ fn main() -> Result<()> {
...
-                "vendor/cosmos/base/abci/v1beta1/abci.proto",
-                "vendor/cosmos/crypto/multisig/v1beta1/multisig.proto",
-                "vendor/cosmos/staking/v1beta1/query.proto",
-                "vendor/cosmos/tx/v1beta1/tx.proto",
-                "vendor/go-header/p2p/pb/header_request.proto",
-            ],
-            &["vendor", "vendor/nmt"],
-        )?;
+        .compile_fds(fds)
+        .expect("prost failed");
```
oblique, line 110: Since we only `expect`. Can you remove the `Result<()>` from `main` signature?

zvolin (approved):
:+1:

fl0rek (approved):
nice, thanks!

PR #408, opened 2024-10-02
chore(node-wasm): add automatic generation of the types file

zvolin:


.github/workflows/release-plz.yml
zvolin, line 1: should we run this manually every time we change something in node-wasm? I'm a bit worried about automating it after last CI hiccups :smiling_face_with_tear:

fl0rek: I'd say go for it. We won't get it right if we don't make some mistakes (sometimes :P)

fl0rek (approved):

oblique (approved):

fl0rek (approved):

PR #420, opened 2024-10-09
ci(node-wasm): Add integration tests for node-wasm

fl0rek:


ci/run-bridge.sh
```diff
@@ -72,7 +82,15 @@ main() {
...
-    --p2p.network "$P2P_NETWORK"
+    --p2p.network "$P2P_NETWORK" &
+
+  # each node without SKIP_AUTH connects to the one with, so that bridges can discover eachother
+  if [ ! "$SKIP_AUTH" == "true" ] ; then
+    connect_to_common_bridge
+  fi
+
+  # don't exit until celestia bridge running in the background does
+  wait
```
zvolin, line 93: can we put `connect_to_common_bridge` to background before `bridge start` and remove the `wait`?

fl0rek: Sure, swapped :+1:

node-wasm/src/client.rs
```diff
@@ -409,3 +409,145 @@ impl WasmNodeConfig {
...
+    }
+
+    fn setup_logs() {
+        let fmt_layer = tracing_subscriber::fmt::layer()
+            .with_ansi(false)
+            .with_timer(UtcTime::rfc_3339()) // std::time is not available in browsers
+            .with_writer(MakeConsoleWriter) // write events to the console
+            .with_filter(LevelFilter::DEBUG); // TODO: allow customizing the log level
+        tracing_subscriber::registry().with(fmt_layer).init();
+    }
```
zvolin, line 472: can we re-use the setup logging from utils?

node-wasm/src/worker.rs
```diff
@@ -41,7 +41,7 @@ pub enum WorkerError {
 }
 
 #[wasm_bindgen]
-struct NodeWorker {
+pub(crate) struct NodeWorker {
```
zvolin, line 44: actually this should just be `pub`, as anything we have `#[wasm_bindgen]` on, so our rust wasm api matches js one. Tho we can do it separately if you want as it's 'breaking' :stuck_out_tongue:

node-wasm/src/client.rs
fl0rek: This was leftover from debugging, we don't setup logging on native tests either. Removed :+1:

node-wasm/src/client.rs
```diff
@@ -409,3 +409,116 @@ impl WasmNodeConfig {
...
+            .addrs
+            .into_iter()
+            .find(|ma| {
+                let not_localhost = !ma
+                    .iter()
+                    .any(|prot| prot == Protocol::Ip4("127.0.0.1".parse().unwrap()));
+                let webtransport = ma
+                    .protocol_stack()
+                    .any(|protocol| protocol == "webtransport");
+                not_localhost && webtransport
```
zvolin, line 508:
```suggestion
                ma.iter().all(|prot| {
                    prot != Protocol::WebTransport && prot != Protocol::Ip4([127, 0, 0, 1].into())
                })
```

node/tests/node.rs
```diff
@@ -28,7 +28,8 @@ async fn connects_to_the_go_bridge_node() {
     let node = new_connected_node().await;
 
     let info = node.network_info().await.unwrap();
-    assert_eq!(info.num_peers(), 1);
+    // our docker setup is two bridges connected, so there's already one another node connected
+    assert_eq!(info.num_peers(), 2);
```
zvolin, line 32: may be more future proof
```suggestion
    assert!(info.num_peers() >= 1);
```

node-wasm/src/client.rs
```diff
@@ -409,3 +409,116 @@ impl WasmNodeConfig {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use super::*;
+    use crate::worker::NodeWorker;
```
zvolin, line 418: let's move those 2 below other imports

node-wasm/src/client.rs
```diff
@@ -409,3 +409,116 @@ impl WasmNodeConfig {
...
+        remove_database().await.expect("failed to clear db");
+        let rpc_client = Client::new(WS_URL).await.unwrap();
+        let bridge_ma = fetch_bridge_webtransport_multiaddr(&rpc_client).await;
+
+        let client = spawn_connected_node(vec![bridge_ma.to_string()]).await;
+
+        let info = client.network_info().await.unwrap();
+        assert_eq!(info.num_peers, 1);
+
+        gloo_timers::future::sleep(Duration::from_secs(2)).await;
```
zvolin, line 459:
```suggestion
        gloo_timers::future::sleep(Duration::from_millis(300)).await;
```
I think we can reduce this now that we solved other issues

ci/docker-compose.yml
```diff
@@ -22,8 +22,10 @@ services:
       # provide an id for the bridge node (default: 0)
       # each node should have a next natural number starting from 0
       - NODE_ID=0
+      # setting SKIP_AUTH to true disables the use of JWT for authentication
+      - SKIP_AUTH=true
     ports:
-      - 26658:26658
+      - 36658:26658
```
zvolin, line 28: could we revert the port change and align url in celestia-rpc tests instead?
+ bonus points for setting 46658 and NODE_ID=2 in the commented out bridge below :p
Forgive me, I swear I tried my best not commenting on this but my inner pendant won

node-wasm/src/client.rs
zvolin: + nit, maybe import the sleep

node-wasm/src/client.rs
zvolin: curious, why cannot we call webtransport?

zvolin: I'm blind, I thought we are avoiding webtransport, sorry :p

node-wasm/src/client.rs
fl0rek: both below? In other tests we seem to start with `super`, so I've put that one on top and moved crate import down.

ci/docker-compose.yml
fl0rek: Reordered dockers by port and fixed the one in comment 👍

oblique (approved):

PR #422, opened 2024-10-10
chore(rpc): Remove a workaround for go-jsonrpc not returning null ok status correctly

### LINKED ISSUE #53: Remove workaround for incorrect RPC response from go-jsonrpc
During RPC calls go celestia node returns invalid (JSON-RPC-wise) responses for following calls:
- `p2p.BlockPeer`
- `p2p.ClosePeer`
- `p2p.Connect`
- `p2p.Protect` (but _not_ `p2p.Unprotect`)
- `p2p.UnblockPeer`
and probably others.
 
In general it seems like `jsonrpsee` is unhappy with the calls that have `Returns: Null` in [docs](https://docs.celestia.org/api/v0.11.0-rc8). On success these send a response without a required `result` field:
```
Custom error: Unparseable message: {\"id\":3,\"jsonrpc\":\"2.0\"}")'
```

As a workaround, we can send these as notifications. Node still executes remote call, but response isn't expected, so `jsonrpsee` doesn't enter error state. This has an obvious downside of ignoring any errors that might actually come back. Once we no longer need to accommodate this behaviour, workaround should be removed.

Relevant issue in `celestia-node`: https://github.com/celestiaorg/celestia-node/issues/2460
and in `go-jsonrpc`: https://github.com/celestiaorg/celestia-node/issues/2460

fl0rek:
closes #53 

Celestia seems to return correct status now

### DIFF
--- a/rpc/src/p2p.rs
+++ b/rpc/src/p2p.rs
@@ -17,20 +17,17 @@ pub trait P2P {
     #[method(name = "p2p.BandwidthStats")]
     async fn p2p_bandwidth_stats(&self) -> Result<BandwidthStats, Error>;
 
-    // This method does not report errors due to a workaround to a go-jsonrpc bug, see https://github.com/eigerco/celestia-node-rs/issues/53
     /// BlockPeer adds a peer to the set of blocked peers.
     #[method(name = "p2p.BlockPeer")]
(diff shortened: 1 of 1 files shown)


zvolin (approved):
nice :+1:

PR #442, opened 2024-10-15
feat(node,node-wasm)!: Allow customising syncing window size

fl0rek:


node-wasm/src/client.rs
```diff
@@ -32,6 +34,8 @@ pub struct WasmNodeConfig {
     /// A list of bootstrap peers to connect to.
     #[wasm_bindgen(getter_with_clone)]
     pub bootnodes: Vec<String>,
+    /// Custom syncing window size. Pruning starts one hour after syncing window end.
```
zvolin, line 37: could you explain there shortly what syncing window is?

node/src/node.rs
```diff
@@ -74,6 +75,8 @@ where
     pub p2p_listen_on: Vec<Multiaddr>,
     /// Maximum number of headers in batch while syncing.
     pub sync_batch_size: u64,
+    /// Syncing window size, pruning starts one hour after syncing window end
```
zvolin, line 78: same here, please add shortly what syncing window is

zvolin: also I'm wondering if we could use Option there to automatically choose default one in case it's not provided by user. But no hard opinion, will defer to @oblique :p

zvolin (approved):
looks nice, left some nits

node/src/node.rs
oblique: I agree with `Option` but what `None` means in this case? Maybe some users will think that it will disable syncing window entirely. Maybe `custom_syncing_window: Option<Duration>` will make more sense.

oblique: We need to revisit builder pattern.

node-wasm/src/client.rs
fl0rek: Used the description from cli everywhere, as it's most descriptive :+1:

node/src/node.rs
fl0rek: Switched to custom_syncing_window and updated the comment. Plus one on the builder pattern xd

node-wasm/src/client.rs
```diff
@@ -373,6 +378,7 @@ impl WasmNodeConfig {
             bootnodes: canonical_network_bootnodes(network.into())
                 .map(|addr| addr.to_string())
                 .collect::<Vec<_>>(),
+            syncing_window_secs: None,
```
oblique, line 381: Should we add `custom_` here too?

fl0rek: I initially skipped it, as field could just be omitted in js, but I guess it won't hurt for consistency.

oblique (approved):

PR #444, opened 2024-10-18
fix(node-wasm): workaround to clean extra message from worker

zvolin:


node-wasm/src/ports.rs
```diff
@@ -211,13 +211,17 @@ impl WorkerClient {
...
+            let worker_response = response_channel
+                .recv()
+                .await
+                .expect("response channel should never drop")
+                .context("error executing command")?;
 
-        Ok(worker_response)
+            if !worker_response.is_internal_pong() {
+                return Ok(worker_response);
+            }
```
oblique, line 223: I think this is more correct. In this case `InternalPong` will be returned if ping was requested.

```suggestion
            // Skip InternalPong if requested command was not InternalPing.
            // We use this because ping is meant to be used with timeout but the server might
            // reply with pong after the timeout is reached.
            if worker_response.is_internal_pong() && !matches!(&command, NodeCommand::InternalPing) {
                continue;
            }
            
            return Ok(worker_response);
```

zvolin: nice catch

oblique (approved):

fl0rek (approved):
:ship:

PR #450, opened 2024-10-22
feat(types)!: add blob reconstruction from shares

zvolin:


types/src/blob.rs
```diff
@@ -135,6 +138,133 @@ impl Blob {
...
+
+        loop {
+            let mut blob = {
+                // find next share from blobs namespace that is sequence start
+                let Some(start) = shares.find(|&shr| shr.sequence_length().is_some()) else {
+                    break;
+                };
+                iter::once(start).chain(&mut shares)
+            };
+            blobs.push(Blob::reconstruct(&mut blob)?);
```
zvolin, line 263: should this error out if shares iterator ends while we are in the middle of some blob?
I'm wondering if it shouldn't be a valid usecase where user passes eg. 3 rows and we get all blobs that they contain, without failing if 3rd row only has part of some next blob

rpc/tests/share.rs
```diff
@@ -29,10 +29,14 @@ async fn get_share() {
 async fn get_shares_by_namespace() {
     let client = new_test_client(AuthLevel::Write).await.unwrap();
     let namespace = random_ns();
-    let data = random_bytes(1024);
-    let blob = Blob::new(namespace, data.clone()).unwrap();
+    let blobs: Vec<_> = (0..rand::random::<usize>() % 4 + 1)
```
oblique, line 32: i think we can remove random here

types/src/blob.rs
```diff
@@ -135,6 +138,137 @@ impl Blob {
...
+    where
+        I: IntoIterator<Item = &'a Share>,
+    {
+        let mut shares = shares.into_iter();
+        let Some(first_share) = shares.next() else {
+            return Err(Error::MissingShares);
+        };
+        let Some(blob_len) = first_share.sequence_length() else {
+            return Err(Error::NotSequenceStart);
+        };
```
oblique, line 176:
```suggestion
        let first_share = shares.next().ok_or(Error::MissingShares)?;
        let blob_len = first_share
            .sequence_length()
            .ok_or(Error::NotSequenceStart)?;

```

types/src/blob.rs
```diff
@@ -135,6 +138,137 @@ impl Blob {
...
+
+        let shares_needed = blob_len_in_shares(blob_len as usize);
+        let mut data =
+            Vec::with_capacity(shares_needed * appconsts::CONTINUATION_SPARSE_SHARE_CONTENT_SIZE);
+        data.extend_from_slice(first_share.payload().expect("non parity"));
+
+        for _ in 1..shares_needed {
+            let Some(share) = shares.next() else {
+                return Err(Error::MissingShares);
+            };
```
oblique, line 191:
```suggestion
            let share = shares.next().ok_or(Error::MissingShares)?;

```

oblique (approved):

types/src/blob.rs
```diff
@@ -135,6 +138,133 @@ impl Blob {
...
+
+        // remove padding
+        data.truncate(blob_len as usize);
+
+        Self::new(namespace, data)
+    }
+
+    /// Reconstructs all blobs from shares.
+    ///
+    /// This function will seek shares that indicate start of next blob (with
```
fl0rek, line 216:
```suggestion
    /// This function will seek shares that indicate start of the next blob (with
```

types/src/error.rs
```diff
@@ -172,13 +176,29 @@ pub enum Error {
...
     #[error("Invalid dimensions of EDS")]
     EdsInvalidDimentions,
 
     /// Zero block height.
     #[error("Invalid zero block height")]
     ZeroBlockHeight,
+
+    /// Share is not a start of a new blob.
+    #[error("Share is not sequence start")]
+    NotSequenceStart,
```
fl0rek, line 189:
```suggestion
    /// Share is not a start of a new blob.
    #[error("Expected sequence starting share")]
    ExpectedShareSequenceStart,
```
nit: maybe? 
Otherwise, the error wording is a bit weird to me, could be
```suggestion
    /// Share is not a start of a new blob.
    #[error("Share does not start a sequence")]
    NotSequenceStart,
```

fl0rek (approved):
few grammar nits, otherwise ready to go!

PR #454, opened 2024-10-24
feat: Tonic gRPC

fl0rek:


rpc/src/lib.rs
```diff
@@ -9,6 +9,8 @@ mod header;
 mod p2p;
 pub mod share;
 mod state;
+#[cfg(all(feature = "tonic", not(target_arch = "wasm32")))]
+pub mod tonic;
```
fl0rek, line 13: RFC about the module layout

proto/vendor/celestia/da/data_availability_header.pb.go
zvolin, line 1: I think we shouldn't include any `.go` files in vendor

proto/Cargo.toml
```diff
@@ -14,17 +14,35 @@ keywords = ["blockchain", "celestia", "lumina"]
...
+celestia-tendermint-proto.workspace = true
+prost.workspace = true
+prost-types.workspace = true
+cosmos-sdk-proto = { version = "0.26.0", default-features = false }
 serde = { version = "1.0.203", features = ["derive"] }
+pbjson = "0.7.0"
+pbjson-types = "0.7.0"
+tonic = { version = "0.12.3", optional = true, default-features = false, features = [
+  "codegen", "prost"
+]}
```
zvolin, line 26: shouldn't all of them be optional?

proto/src/lib.rs
```diff
@@ -4,4 +4,8 @@
 
 pub mod serializers;
 
+#[cfg(not(feature = "tonic"))]
 include!(concat!(env!("OUT_DIR"), "/mod.rs"));
+
+#[cfg(feature = "tonic")]
+::tonic::include_proto!("mod");
```
zvolin, line 11: are those outputs compatible for the types that are common to them? I'm wondering if it's safe to mix crates that depends on celestia-proto with and without the tonic feature

rpc/Cargo.toml
zvolin, line 1: are changes in celestia-rpc a leftover?

tonic/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "celestia-tonic"
```
zvolin, line 2: I think I'd prefer `celestia-grpc` and `grpc` directory

tonic/src/tendermint.rs
zvolin, line 1: doesn't seem to be included

tonic/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+use celestia_tendermint_proto::v0_34::types::BlobTx;
+
+use cosmrs::crypto::PublicKey;
+use cosmrs::Any;
+
+use crate::Error;
+
+pub mod tx;
+
+pub struct Block(pub TendermintBlock);
```
zvolin, line 34: why do we need a newtype for Block?

tonic/src/client.rs
```diff
@@ -0,0 +1,75 @@
...
+                self.grpc_channel.clone(),
+                self.auth_interceptor.clone(),
+            );
+            let request = ::tonic::Request::new(( $($param),* ).into_parameter());
+            let response = client.$method(request).await;
+
+            Ok(response?.into_inner().try_from_response()?)
+        }
+    };
+}
```
zvolin, line 46:
```suggestion
macro_rules! make_method {
    ($grpc_client:ty, $original_method:ident, $renamed_method:ident( $($param:ident : $param_type:ty),* ) -> $ret:ty) => {
        pub async fn $renamed_method(&mut self, $($param: $param_type),*) -> $ret {
            let mut client = <$grpc_client>::with_interceptor(
                self.grpc_channel.clone(),
                self.auth_interceptor.clone(),
            );
            let request = ::tonic::Request::new(( $($param),* ).into_parameter());
            let response = client.$original_method(request).await;

            response?.into_inner().try_from_response()
        }
    };
}
```

I was thinking about removing this `::` pathing requirement, something like this seems to work. The invocation changes a bit tho, requires a placeholder generic, like:
```rs
    make_method!(config::service_client::ServiceClient<_>, config, get_min_gas_price() -> Result<f64, Error>);
```

types/Cargo.toml
```diff
@@ -14,17 +14,18 @@ keywords = ["blockchain", "celestia", "lumina"]
...
-nmt-rs = { workspace = true }
+celestia-tendermint-proto.workspace = true
+nmt-rs.workspace = true
 
 base64 = "0.22.1"
 bech32 = "0.11.0"
 bytes = "1.6.0"
 cid = { version = "0.11.1", default-features = false, features = ["std"] }
 const_format = "0.2.32"
+cosmrs = "0.21.0"
```
zvolin, line 28: I'm in general a bit concerned about cosmrs dependency, is it hardly needed? If so, maybe we should first resolve tendermint issue... :see_no_evil:

tools/update-proto-vendor.sh
```diff
@@ -57,21 +56,21 @@ rm -rf vendor/google
 mkdir -p vendor/google/api
 cp ../target/proto-vendor-src/googleapis-master/google/api/{annotations.proto,http.proto} vendor/google/api
 
-rm -rf vendor/header
```
zvolin, line 60: why is this removed?

tonic/src/client.rs
fl0rek: I've replaced this macro with procedural macro in `grpc_method`

tonic/Cargo.toml
fl0rek: good call, done :+1:

proto/Cargo.toml
fl0rek: yes, changed to optional

proto/src/lib.rs
fl0rek: Good question, I believe so as in our crates compile for both tonic and non-tonic. I'll try to get a project with mixed tonic-ness and test that too.

rpc/Cargo.toml
fl0rek: other than the reordering, yes. Reverted.

tonic/src/tendermint.rs
fl0rek: yes, removed :+1:

tonic/src/types.rs
fl0rek: Actually we don't strictly need it, removed.

grpc/src/types/tx.rs
```diff
@@ -0,0 +1,120 @@
...
+            //logs: response.logs
+            info: response.info,
+            gas_wanted: response.gas_wanted,
+            gas_used: response.gas_used,
+            //tx: response.tx
+            timestamp: response.timestamp,
+            //events: response.events
+        })
+    }
+}
```
oblique, line 89: cleanup commented code

grpc/src/client.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Get latest block
+    #[grpc_method(TendermintServiceClient::get_latest_block)]
+    async fn get_latest_block(&mut self) -> Result<Block, Error>;
+
+    /// Get block by height
+    #[grpc_method(TendermintServiceClient::get_block_by_height)]
+    async fn get_block_by_height(&mut self, height: i64) -> Result<Block, Error>;
+
+    // TODO get_node_info
+    // make_method!(TendermintServiceClient::get_node_info; get_node_info() -> NodeInfo);
```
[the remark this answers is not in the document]
fl0rek: I've planned to implement few most obvious calls, but ended up not doing this one as it won't be needed by blob submission. It probably make sense to implement it at some point, thus TODO, but we don't have to track it here.

grpc/src/types/tx.rs
fl0rek: uncommented and cleaned up :+1:

proto/build.rs
```diff
@@ -63,26 +67,53 @@ static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
...
-            "vendor/cosmos/tx/v1beta1/tx.proto",
-            "vendor/go-header/p2p/pb/header_request.proto",
-        ],
-        ["vendor", "vendor/nmt"],
-    )
-    .expect("protox failed to build");
+    let fds = protox_compile();
+    prost_build(fds);
+    #[cfg(feature = "tonic")]
+    tonic_build(protox_compile())
```
fl0rek, line 109: deduplicate

.github/workflows/ci.yml
```diff
@@ -63,7 +63,7 @@ jobs:
     - name: Install wasm-pack
       uses: taiki-e/cache-cargo-install-action@v1
       with:
-        tool: wasm-pack@0.12.1
```
zvolin, line 66: we should bump it also in release-plz workflow, just check if it still works with next template

grpc/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+celestia-types = { workspace = true, features = [ "tonic" ] }
+celestia-proto = { workspace = true, features = [ "tonic", "cosmrs" ] }
+celestia-tendermint.workspace = true
+prost.workspace = true
+celestia-grpc-macros = { version = "0.1.0", path = "grpc-macros" }
+cosmos-sdk-proto = "0.26.0"
+
+hex = "0.4.3"
+serde = "1.0.215"
+cosmrs = "0.21.0"
```
zvolin, line 32: is this still used?

grpc/src/types/pagination.rs
```diff
@@ -0,0 +1,30 @@
...
+    query_start: KeyOrOffset,
+    limit: Option<u64>,
+    count_total: bool,
+    reverse: bool,
+}
+
+impl From<PageRequest> for RawPageRequest {
+    fn from(item: PageRequest) -> RawPageRequest {
+        RawPageRequest {
+
```
zvolin, line 22: is this intended to look like this?

grpc/src/types/tx.rs
```diff
@@ -0,0 +1,245 @@
...
+use celestia_tendermint_proto::v0_34::abci::Event;
+use celestia_tendermint_proto::Protobuf;
+use celestia_types::auth::{AccountKeypair, BaseAccount};
+use celestia_types::blob::{Blob, MsgPayForBlobs, RawBlob, RawBlobTx, RawMsgPayForBlobs};
+
+use crate::types::{FromGrpcResponse, IntoGrpcParam};
+use crate::Error;
+
+/// Response to a tx query
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
zvolin, line 27: we have `celestia_types::state::TxResponse`, I think we should either move this there to types or just use the one from types (although it's just re-export)

grpc/src/types/tx.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+impl IntoGrpcParam<RawGetTxRequest> for String {
+    fn into_parameter(self) -> RawGetTxRequest {
+        RawGetTxRequest { hash: self }
+    }
+}
+
+pub fn prep_signed_tx(
```
zvolin, line 160: this only works with message pay for blob. I'd rather make it more generic and accept any TxBody here, we can have another helper for making TxBody out of MsgPayForBlobs

zvolin: also would rename it to `sign_tx` or something

grpc/src/types/tx.rs
```diff
@@ -0,0 +1,245 @@
...
+    msg_pay_for_blobs: &MsgPayForBlobs,
+    base_account: &BaseAccount,
+    gas_limit: u64,
+    fee: u64,
+    chain_id: String,
+    account_keys: AccountKeypair,
```
zvolin, line 166: could we instead accept something like this?
```
public_key: VerifyingKey,
signer: impl Signer,
```

The reason for that is this enforces us to store sk there, while we can have different signers like hsm, or even use [api exposed by leap](https://docs.leapwallet.io/cosmos/for-dapps-connect-to-leap/api-reference#sign-direct-protobuf) for signing

grpc/tests/tonic.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    assert_eq!(&account, first_account);
+}
+
+#[tokio::test]
+async fn submit_blob() {
+    let mut client = new_test_client().await.unwrap();
+    let address = "celestia1rkfxnqt8wwu2vqgpa2ph84xa2ty0nseex4xqlc".to_string();
+    let private_key =
+        hex::decode("374b1d38f76c57fb6a1bb7bb840795239640441a37f506dc8de0d82b1ea9f690").unwrap();
```
zvolin, line 73: as discussed, we could load that from `ci/credentials`

grpc/tests/tonic.rs
```diff
@@ -0,0 +1,106 @@
...
+    let namespace = Namespace::new_v0(&[1, 2, 3]).unwrap();
+    let blob = Blob::new(namespace, "Hello, World!".into(), AppVersion::V1).unwrap();
```
zvolin, line 75:
```suggestion
    let blob = Blob::new(namespace, b"Hello, World!", AppVersion::V3).unwrap();
```

types/src/auth.rs
```diff
@@ -0,0 +1,161 @@
...
+/// [`BaseAccount`] defines a base account type.
+///
+/// It contains all the necessary fields for basic account functionality.
+///
+/// Any custom account type should extend this type for additional functionality
+/// (e.g. vesting).
+#[derive(Debug, Clone, PartialEq)]
+pub struct BaseAccount {
+    /// Bech32 `AccountId` of this account.
+    pub address: String,
```
zvolin, line 45: shouldn't we use Address type we have in types?

grpc/Cargo.toml
fl0rek: No, removed cosmrs entirely now and moved to our protobuf definitions and types in celestia_types and celestia_grpc depending on whether it's generic or grpc specific respectively.

grpc/src/types/pagination.rs
fl0rek: This was staged accidentally, removed this file.,

grpc/tests/tonic.rs
fl0rek: I didn't find reasonably straightforward way of loading the encrypted pem file in the tendermint format, so I've added saving also plaintext hex-encoded key version in `run-validator` and load that.

grpc/tests/tonic.rs
fl0rek: Changed to v3. `b"Hello, World!"` gives us slice, while blob seems to want a vec, so we'd need to call `to_vec` anyway so left it as is

.github/workflows/ci.yml
fl0rek: Interestingly in release-plz we install it via wasm-pack-action, and since we don't set version there it automatically downloads the latest one. 
I'll open a separate PR where we align the two, since I assume we want to have consistency there :P

types/src/auth.rs
fl0rek: I've missed that we already have it, thanks for the tip. Used that instead :+1:

grpc/src/types/tx.rs
fl0rek: Got rid of the `AccountKeypair` struct (outside of tests) and switched to accepting public key and `impl Singer`

ci/run-validator.sh
```diff
@@ -117,6 +120,7 @@ setup_private_validator() {
   celestia-appd keys add "$NODE_NAME" --keyring-backend="test"
   validator_addr="$(celestia-appd keys show "$NODE_NAME" -a --keyring-backend="test")"
   # Create a validator's genesis account for the genesis.json with an initial bag of coins
+  echo "VALIDATOR ADDR: $validator_addr"
```
zvolin, line 123: leftover?

grpc/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "celestia-grpc"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+description = "A collection of traits for interacting with Celestia validator nodes gRPC"
```
zvolin, line 6:
```suggestion
description = "A client for interacting with Celestia validator nodes gRPC"
```

ci/run-validator.sh
fl0rek: Yup, leftover from my furious debugging, removed :+1:

grpc/src/types/tx.rs
fl0rek: Moved the transaction related types to celestia types and updated already existing ones :+1:

proto/src/lib.rs
oblique: Since we use `include_file("mod.rs")` in prost and tonic builders, it is safe to just use `include!(concat!(env!("OUT_DIR"), "/mod.rs"));` for both of them.

 Also common types are compatible, because tonic for generates type definitions with prost and then generates additional code for the grpc services.

grpc/src/types/auth.rs
```diff
@@ -0,0 +1,92 @@
...
+};
+use celestia_types::state::Address;
+
+use crate::types::make_empty_params;
+use crate::types::{FromGrpcResponse, IntoGrpcParam};
+use crate::Error;
+
+/// Enum representing different types of account
+#[derive(Debug, PartialEq)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
fl0rek: My thinking was to be future compatible in case new account type pops up. Upon retrospection I'm not sure that's needed and we can just release breaking version in that case.

proto/src/lib.rs
fl0rek: Yes, I didn't comment to confirm back, but the code is mostly the same, with additional traits implemented. 

Also true, we can use `include!` in both, I was mostly looking for an excuse to use nicer looking tonic version :P Rolled back.

grpc/src/client.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Get accounts
+    #[grpc_method(AuthQueryClient::accounts)]
+    async fn get_accounts(&mut self) -> Result<Vec<Account>, Error>;
+
+    /// Broadcast Tx
+    #[grpc_method(TxServiceClient::broadcast_tx)]
+    async fn broadcast_tx(
+        &mut self,
+        tx: RawTx,
+        blobs: Vec<Blob>,
```
zvolin, line 79: broadcast_tx should be generic, it should only take a raw tx imo. If we want to submit blobs, we should prepare whole tx beforehand. Submitting blobs is just one of the available transactions, you still have transfers etc.

proto/build.rs
```diff
@@ -50,6 +53,7 @@ static CUSTOM_TYPE_ATTRIBUTES: &[(&str, &str)] = &[
 static CUSTOM_FIELD_ATTRIBUTES: &[(&str, &str)] = &[
     (".celestia.core.v1.da.DataAvailabilityHeader.row_roots", VEC_BASE64STRING),
     (".celestia.core.v1.da.DataAvailabilityHeader.column_roots", VEC_BASE64STRING),
+    #[cfg(not(feature = "tonic"))]
```
zvolin, line 56: do we need this conditionally? I think it can hurt us too

types/src/auth.rs
```diff
@@ -0,0 +1,157 @@
...
+
+pub use celestia_proto::cosmos::auth::v1beta1::BaseAccount as RawBaseAccount;
+pub use celestia_proto::cosmos::auth::v1beta1::ModuleAccount as RawModuleAccount;
+
+const COSMOS_ED25519_PUBKEY: &str = "/cosmos.crypto.ed25519.PubKey";
+const COSMOS_SECP256K1_PUBKEY: &str = "/cosmos.crypto.secp256k1.PubKey";
+
+/// Params defines the parameters for the auth module.
+#[derive(Debug)]
+pub struct AuthParams {
```
zvolin, line 25: couldn't we just re-export `Params` as `AuthParams`? or is prost type tricky to work with?

types/src/bit_array.rs
zvolin, line 1: could we move this and auth to the `state/`?

zvolin (commented):
left some last comments, otherwise lgtm

types/Cargo.toml
```diff
@@ -60,10 +63,12 @@ default = ["p2p"]
...
 wasm-bindgen = ["celestia-tendermint/wasm-bindgen"]
+tonic = ["dep:pbjson-types", "celestia-proto/tonic"]
 
 [package.metadata.docs.rs]
 features = ["p2p", "test-utils"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.cargo-udeps.ignore]
 development = ["indoc"]
+normal = ["prost-types"] # unused if tonic flag is enabled
```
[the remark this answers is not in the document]
fl0rek: afaik, feature flags can only enable optional dependencies, so this would require adding `prost` flag. This in conjunction with logic to have only one of prost and tonic enabled, and propagation seemed more convoluted than having the dependency ignored here, though I'm not wedded to that conclusion :p

types/src/bit_array.rs
fl0rek: yup, moved :+1:

proto/build.rs
fl0rek: This was needed because I wanted to avoid the pbjson/prost::Any swicheroo (and this didn't seem strictly needed for tonic). Since we'll be moving to tendermint's Any everywhere, I think it can stay unconditional.

grpc/src/client.rs
fl0rek: Thanks for noting that, I was a bit confused on what tx is apparently :P 
I've split raw `broadcast_tx`, which accepts any transaction bytes and a helper method which accepts typed arguments for blob submission.

types/src/auth.rs
fl0rek: In my fury of adding types from protobufs I didn't notice we might as well use that one directly, switched :+1:

types/src/bit_array.rs
zvolin: I think you forgot to move bit_array

zvolin (approved):
lgtm

types/src/bit_array.rs
fl0rek: I thought of it as a container type, which is why I put it in root, but actually we can move it to state since it's only used there.

fl0rek: moved

Cargo.toml
```diff
@@ -1,20 +1,23 @@
 [workspace]
 resolver = "2"
-members = ["cli", "node", "node-wasm", "proto", "rpc", "types"]
+members = ["cli", "grpc", "node", "node-wasm", "proto", "rpc", "types", ]
```
oblique, line 3:
```suggestion
members = ["cli", "grpc", "node", "node-wasm", "proto", "rpc", "types"]
```

proto/CHANGELOG.md
```diff
@@ -6,6 +6,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Any->Any
+- blob submission
+- lasts tests
+- doctests
+- missed serialisation
+- Merge remote-tracking branch 'upstream/main' into feat/grpc
+- cleanup, updates
+- missing rpc methods
+- types cleanup
+
```
oblique, line 39: what is this?

proto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "celestia-proto"
-version = "0.5.0"
+version = "0.5.1"
```
oblique, line 3: Wouldn't this be increased by release-plz?

proto/CHANGELOG.md
fl0rek: commited by mistake when I was testing release-plz, removing.

proto/Cargo.toml
fl0rek: commited by mistake when I was testing release-plz, removing.

zvolin (approved):

oblique (approved):

PR #464, opened 2024-11-11
fix(node): Increase sleep in `head_selection_with_multiple_peers` test case

oblique:
This test was sometimes failing because the `network_head` of height 20 was returned.


zvolin (approved):

fl0rek (approved):
LGTM!

PR #467, opened 2024-11-15
Correct Prost build issue.

musitdev:
Using the current main branch, if the Cargo.lock file is removed the build fails with the error:
```
error[E0277]: the trait bound `celestia_tendermint_proto::v0_34::types::ValidatorSet: prost::Message` is not satisfied
    --> target/debug/build/celestia-proto-d9e58f87878dccfd/out/header.pb.rs:5:28
     |
5    | #[derive(Clone, PartialEq, ::prost::Message)]
     |                            ^^^^^^^^^^^^^^^^ the trait `prost::Message` is not implemented for `celestia_tendermint_proto::v0_34::types::ValidatorSet`
     |
     = help: the following other types implement trait `prost::Message`:
```
The issue comes from the fact that the crate `celestia-tendermint-proto` use the post version 0.13.3 and the proto crate the version 0.12.6.
I've changed the `dependencies` version to 0.13.3 but kept the `build-dependencies `to the old one because otherwise, build.rs  doesn't build.


zvolin:
Yeah, we saw it. The issue is that the celestia-tendermint-proto doesn't follow semver. We were discussing it and unless it's an issue for you, we'd fixing it by merging #

musitdev:
Ok I see. No need to merge if you have another plan to correct it. For now, we'll use our fork and when it will be solved, we'll switch. I'm little surprised that no other users find this issue. Perhaps it's still hidden for them, but you're going to have more and more user that get it.

PR #468, opened 2024-11-28
feat(node-wasm, types)!: Add method to get blobs for wasm

fl0rek:


types/src/blob/commitment.rs
```diff
@@ -50,8 +52,19 @@ use crate::{InfoByte, Share};
 /// [`Nmt`]: crate::nmt::Nmt
 /// [`ExtendedDataSquare`]: crate::ExtendedDataSquare
 /// [`share commitment rules`]: https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/data_square_layout.md#blob-share-commitment-rules
+
+#[cfg(not(feature = "wasm-bindgen"))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Commitment {
```
zvolin, line 58: `cfg_attr` doesn't work in this case?

types/src/blob/msg_pay_for_blobs.rs
```diff
@@ -80,7 +81,7 @@ impl From<MsgPayForBlobs> for RawMsgPayForBlobs {
         let share_commitments = msg
             .share_commitments
             .into_iter()
-            .map(|c| c.0.to_vec())
+            .map(|c| Hash::from(c).to_vec())
```
oblique, line 84:
```suggestion
            .map(|c| c.hash().to_vec())
```

very minor thing but i prefer this

types/src/blob/msg_pay_for_blobs.rs
```diff
@@ -106,9 +107,8 @@ impl TryFrom<RawMsgPayForBlobs> for MsgPayForBlobs {
             .share_commitments
             .into_iter()
             .map(|c| {
-                Ok(Commitment(
-                    c.try_into().map_err(|_| Error::InvalidComittmentLength)?,
-                ))
+                let hash = Hash::try_from(c).map_err(|_| Error::InvalidComittmentLength)?;
+                Ok(hash.into())
```
oblique, line 111: what about `Ok(Commitment::new(hash))`?

grpc/tests/tonic.rs
zvolin, line 1: I think we can revert changes there and in `utils.rs`. We've already resolved ci failures about missing docs and stuff at another level

types/src/blob/commitment.rs
```diff
@@ -50,10 +54,19 @@ use crate::{InfoByte, Share};
 /// [`Nmt`]: crate::nmt::Nmt
 /// [`ExtendedDataSquare`]: crate::ExtendedDataSquare
 /// [`share commitment rules`]: https://github.com/celestiaorg/celestia-app/blob/main/specs/src/specs/data_square_layout.md#blob-share-commitment-rules
+#[cfg_attr(all(feature = "wasm-bindgen", target_arch = "wasm32"), wasm_bindgen)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Commitment(pub merkle::Hash);
+pub struct Commitment {
+    ///  hash of the commitment
```
zvolin, line 60:
```suggestion
    /// Hash of the commitment
```

types/src/blob/commitment.rs
```diff
@@ -101,7 +114,28 @@ impl Commitment {
 
         let hash = merkle::simple_hash_from_byte_vectors::<crypto::default::Sha256>(&subtree_roots);
 
-        Ok(Commitment(hash))
+        Ok(Commitment { hash })
+    }
+
+    ///  hash of the commitment
```
zvolin, line 120:
```suggestion
    /// Hash of the commitment
```

types/src/blob/commitment.rs
```diff
@@ -101,7 +114,28 @@ impl Commitment {
...
+    pub fn hash(&self) -> &merkle::Hash {
+        &self.hash
+    }
+}
+
+#[cfg(all(feature = "wasm-bindgen", target_arch = "wasm32"))]
+#[wasm_bindgen]
+impl Commitment {
+    ///  hash of the commitment
```
zvolin, line 129:
```suggestion
    /// Hash of the commitment
```

types/src/blob/commitment.rs
```diff
@@ -8,12 +8,16 @@ use nmt_rs::NamespaceMerkleHasher;
...
+#[cfg(all(feature = "wasm-bindgen", target_arch = "wasm32"))]
+use wasm_bindgen::prelude::*;
 
 use crate::consts::appconsts;
 use crate::nmt::{Namespace, NamespacedHashExt, NamespacedSha2Hasher, Nmt, RawNamespacedHash};
 use crate::{Error, Result};
 use crate::{InfoByte, Share};
 
+// TODO: once https://github.com/rustwasm/wasm-bindgen/pull/4351 is merged,
+// this can be replaced with a single common type definition
```
zvolin, line 20: there is only one type

types/src/nmt.rs
```diff
@@ -91,6 +93,7 @@ pub type Proof = nmt_rs::simple_merkle::proof::Proof<NamespacedSha2Hasher>;
 ///  - secondary reserved namespaces - those use version `0xff` so they are always placed after
 ///    user-submitted data.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Hash)]
+#[cfg_attr(all(feature = "wasm-bindgen", target_arch = "wasm32"), wasm_bindgen)]
```
zvolin, line 96: I'm wondering how usage of namespace works in js, as we don't expose any methods iiuc

oblique (approved):

types/src/nmt.rs
fl0rek: good point, added getters for id and version as well as to_byte equivalent.

zvolin: so user wasn't able to create a namespace, added that too, together with blob creation

oblique (approved):
LGTM

oblique (approved):

PR #472, opened 2024-12-04
feat(node)!: Implement `NodeBuilder` and remove `NodeConfig`

oblique:


node/src/network.rs
```diff
@@ -16,69 +18,180 @@ pub enum Network {
...
-        Network::Arabica => "arabica-11",
-        Network::Mocha => "mocha-4",
-        Network::Private => "private",
-        Network::Mainnet => "celestia",
+impl fmt::Display for Network {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            Network::Mainnet => "mainnet",
+            Network::Arabica => "arabica",
+            Network::Mocha => "mocha",
```
oblique, line 146: Should I capitalize them?

node/src/node/builder.rs
```diff
@@ -0,0 +1,353 @@
...
+        } else {
+            DEFAULT_PRUNING_DELAY
+        };
+
+        if syncing_window < MIN_SYNCING_WINDOW {
+            return Err(NodeBuilderError::VerySmallSynincWindow(syncing_window));
+        }
+
+        // Truncate sampling window if needed.
+        let sampling_window = sampling_window.min(syncing_window);
```
oblique, line 323: I was debating if we should truncate sampling window or return an error if it is bigger than syncing window. What do you think?

Wondertan:
OOC, whats the motivation behind this change?

node/src/node/builder.rs
```diff
@@ -0,0 +1,353 @@
...
+    sampling_window: Option<Duration>,
+    pruning_delay: Option<Duration>,
+}
+
+/// Representation of all the errors that can occur when interacting with the [`NodeBuilder`].
+#[derive(Debug, thiserror::Error)]
+pub enum NodeBuilderError {
+    /// Syncing window is smaller than [`MIN_SYNCING_WINDOW`].
+    #[error("Syncing window is {0:?} but cannot be smaller than {MIN_SYNCING_WINDOW:?}")]
+    VerySmallSynincWindow(Duration),
```
zvolin, line 56: typo
```suggestion
    VerySmallSyncingWindow(Duration),
```

node/src/node/builder.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// Sampling window is smaller than [`MIN_SAMPLING_WINDOW`].
+    #[error("Sampling window is {0:?} but cannot be smaller than {MIN_SAMPLING_WINDOW:?}")]
+    VerySmallSamplingWindow(Duration),
+
+    /// Sampling window is smaller than [`MIN_PRUNING_DELAY`].
+    #[error("Pruning delay is {0:?} but cannot be smaller than {MIN_PRUNING_DELAY:?}")]
+    VerySmallPruningDelay(Duration),
+}
```
zvolin, line 65: I think I'd prefer `s/Very/Too/` on all errors

oblique: Done. Also made `TooSmall` a suffix

node/src/network.rs
oblique: Network ID is returned with `id()` method. This is on purpose because the real network ID a bit different, for example network ID of Arabica is `arabica-11` etc.

zvolin: ah okay, then idc, up to you I guess

oblique: I capitalize them

node/src/node/builder.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// **Minimum:** 60 seconds.
+    pub fn pruning_delay(self, dur: Duration) -> Self {
+        NodeBuilder {
+            pruning_delay: Some(dur),
+            ..self
+        }
+    }
+
+    fn build_config(self) -> Result<NodeConfig<B, S>, NodeBuilderError> {
+        let network = self.network.expect("todo");
```
zvolin, line 240: todo

zvolin (approved):
looks good

node/src/node/builder.rs
oblique: fixed

node/src/node/builder.rs
```diff
@@ -0,0 +1,309 @@
...
+    #[error("Syncing window is {0:?} but cannot be smaller than {MIN_SYNCING_WINDOW:?}")]
+    SyncingWindowTooSmall(Duration),
+
+    /// Pruning delay is smaller than [`MIN_PRUNING_DELAY`].
+    #[error("Pruning delay is {0:?} but cannot be smaller than {MIN_PRUNING_DELAY:?}")]
+    PruningDelayTooSmall(Duration),
+}
+
+impl NodeBuilder<InMemoryBlockstore, InMemoryStore> {
+    /// Creates a new [`NodeBuilder`] that is using in-memory stores.
```
fl0rek, line 62:
```suggestion
    /// Creates a new [`NodeBuilder`] which uses in-memory stores.
```

node/src/node/builder.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///
+    /// After the creation you can use [`NodeBuilder::blockstore`]
+    /// and [`NodeBuilder::store`] to set other stores.
```
fl0rek, line 65:
```suggestion
    /// After the creation you can call [`NodeBuilder::blockstore`]
    /// and [`NodeBuilder::store`] to use other stores.
```

node/src/node/builder.rs
```diff
@@ -0,0 +1,309 @@
...
+{
+    /// Creates and starts a new Celestia [`Node`].
+    pub async fn start(self) -> Result<Node<B, S>> {
+        let (node, _) = self.start_subscribed().await?;
+        Ok(node)
+    }
+
+    /// Creates and starts a new Celestia [`Node`].
+    ///
+    /// Returns [`Node`] alogn with [`EventSubscriber`]. Use this to avoid missing
```
fl0rek, line 115:
```suggestion
    /// Returns [`Node`] along with [`EventSubscriber`]. Use this to avoid missing
```

node/src/node/builder.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub fn syncing_window(self, dur: Duration) -> Self {
+        NodeBuilder {
+            syncing_window: Some(dur),
+            ..self
+        }
+    }
+
+    /// Set pruning delay.
+    ///
+    /// Pruning delay how much time the pruner should wait after syncing window in
```
fl0rek, line 230:
```suggestion
    /// Pruning delay defines how much time the pruner should wait after syncing window in
```

node/src/node/builder.rs
```diff
@@ -0,0 +1,309 @@
...
+            // It is a valid scenario for user to create a node without any bootnodes
+            // and listening addresses. However it may not be what they wanted. Because
+            // of that we display a warning.
+            warn!("Node has empty bootnodes and listening addresses. It will never connect to another peer.");
+        }
+
+        // `Node` is memory hungry when in-memory stores are used and the user most
+        // probably will never know that they need to have a smaller syncing window
+        // to avoid it. In order to make `lumina` more user friendly we decided to
+        // have a different default when our in-memory stores are used.
```
fl0rek, line 262:
```suggestion
        // `Node` is memory hungry when in-memory stores are used and the user may not
        // expect they should set a smaller syncing window to reduce that. For user-friendliness
        // sake, use smaller default syncing window, if we're running in memory.
```
Maybe? Reworded for conciseness.

fl0rek (approved):
Few suggestions wrt comments, otherwise looking good!

zvolin (approved):

fl0rek (approved):
LGTM

PR #473, opened 2024-12-10
feat: adding uniffi bindings to support android and ios

sebasti810:
### Mobile bindings for Lumina node via UniFFI

This PR adds iOS and Android bindings for the node using Mozilla's UniFFI. 
I followed the approach from wasm and tested it successfully in the iOS and Android simulator.

#### Current State
- Basic node operations (start, stop, querying, etc.)  
- Platform-specific storage paths
- Event handling
- Example projects for [iOS](https://github.com/sebasti810/lumina-swift) and [Android](https://github.com/sebasti810/lumina-android)

#### TODOs
- [x] Improve runtime (using `#[uniffi::export(async_runtime = "tokio")]`)
- [x] Improve error handling
- [x] Improve types
- [x] Add docs
- [x] Finish and test example projects


oblique:
We refactored how we construct a `Node`. We introduced `NodeBuilder` and made private the `NodeConfig`. So this PR needs an update.

sebasti810:
@oblique i rebased, now it works again. test-wasm will still fail at the moment. Should I make the wasm crate compatible or should we just ignore it and use something like:
`cargo clippy --workspace --exclude native --all-targets --target=wasm32-unknown-unknown -- -D warnings`
instead of:
`cargo clippy --all --all-targets --target=wasm32-unknown-unknown -- -D warnings`

Cargo.toml
```diff
@@ -27,6 +27,7 @@ prost-types = "0.13.3"
 #libp2p = { path = "../../rust-libp2p/libp2p" }
 #libp2p-core = { path = "../../rust-libp2p/core" }
 #libp2p-swarm = { path = "../../rust-libp2p/swarm" }
+system-configuration = { git = "https://github.com/mullvad/system-configuration-rs", tag = "v0.6.1" }
```
zvolin, line 30: is this used? we generally don't allow git based deps as they are no-go when publishing to crates

node-native/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "native"
```
zvolin, line 2: we were thinking about something like `lumina-node-uniffi` and `node-uniffi` directory. wdyt?
We already use `native` usually when we talk on just running lumina on host / desktop and I think it's a wording used in `lumina-cli`

node-native/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+    pub fn new(network: Network) -> Result<Arc<Self>> {
+        Ok(Arc::new(Self {
+            node: Arc::new(Mutex::new(None)),
+            network,
+            events_subscriber: Arc::new(Mutex::new(None)),
+        }))
+    }
+
+    /// Starts the Lumina node. Returns true if successfully started.
+    pub async fn start(&self) -> Result<bool> {
```
zvolin, line 141: could we accept here some config struct that will allow for more control / tuning node?

node-native/src/types.rs
```diff
@@ -0,0 +1,421 @@
...
+            Network::Custom { id } => {
+                network::Network::Custom(network::NetworkId::new(id).expect("invalid network id"))
+            }
+        }
+    }
+}
+
+/// Statistics of the connected peers
+#[derive(Debug, Clone, Default, Record)]
+pub struct PeerTrackerInfo {
```
zvolin, line 43: we were thinking to add `uniffi` feature in both `celestia-types` and `lumina-node` crates and try to wrap as many stuff as possible directly instead of reintroducing this here. Is there any drawback with such approach? We don't have much experience with uniffi but [here](https://mozilla.github.io/uniffi-rs/latest/tutorial/Rust_scaffolding.html) it mentions that it should be possible with `uniffi_reexport_scaffolding`

zvolin (commented):
thank you for the contribution, we're supper happy about it and would love to getting it upstream :bow:

zvolin:
> test-wasm ci will still fail at the moment. Should I make the crate wasm compatible or should we just ignore it and use something like

I think we don't care about wasm in this new crate so you could just put this in top of `lib.rs`
```rs
#![cfg(not(target_arch = "wasm32"))]
```

sebasti810:
> I think we don't care about wasm in this new crate so you could just put this in top of `lib.rs`
> 
> ```rust
> #![cfg(not(target_arch = "wasm32"))]
> ```

true, that's enough. I thought we would then have to update the Cargo.toml etc. as well with target-specific dependencies, but that is not the case. Thank you

node-native/Cargo.toml
sebasti810: yeah i just saw that. Its definitely better, i'll rename it

node-native/src/lib.rs
sebasti810: sure, like the wasm `NodeStartConfig`? Or what configurations do you think of?

node-uniffi/src/types.rs
```diff
@@ -0,0 +1,516 @@
...
+    /// Maximum number of headers in batch while syncing. Default is 128.
+    pub batch_size: Option<u64>,
+    /// Optional Set the keypair to be used as Node's identity. If None, generates a new Ed25519 keypair.
+    pub ed25519_secret_key_bytes: Option<Vec<u8>>,
+}
+
+impl NodeStartConfig {
+    /// Convert into NodeBuilder for the implementation
+    pub(crate) async fn into_node_builder(self) -> Result<NodeBuilder<RedbBlockstore, RedbStore>> {
+        let base_path = get_base_path()?;
```
oblique, line 90: IMO, this is too high-level api for a library. I think it would be better for the user to select the base path thought `NodeStartConfig`.

This removes the responsibility from us to figure out the project dir or make use of environment variables within the library.

node-uniffi/src/types.rs
oblique, line 1: This file is big and it can get much more bigger in the future. I suggest you to create a `types` directory and split it in multiple files.

node-uniffi/src/types.rs
```diff
@@ -0,0 +1,516 @@
...
+
+#[derive(Record)]
+pub struct ShareCoordinate {
+    pub row: u16,
+    pub column: u16,
+}
+
+/// Events emitted by the node.
+#[derive(uniffi::Enum)]
+pub enum NodeEvent {
```
oblique, line 277: Can we use the one from `lumina-node` crate?

node-uniffi/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+tendermint.workspace = true
+libp2p.workspace = true
+redb = "2.1.1"
+thiserror = "1.0.61"
+serde_json = "1.0.64"
+uniffi = { version = "0.28.3", features = ["bindgen", "tokio", "cli"] }
+tokio = { version = "1.38.0", features = ["macros", "sync"] }
+
+[target.'cfg(target_os = "ios")'.dependencies]
+directories = "5.0.1"
```
oblique, line 26: Add a new line to remove the no-newline icon. Fix other files too.

node-uniffi/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+            network,
+            events_subscriber: Arc::new(Mutex::new(None)),
+        }))
+    }
+
+    /// Start the node without optional configuration.
+    /// UniFFI needs explicit handling for optional parameters to generate correct bindings for different languages.
+    pub async fn start(&self) -> Result<bool> {
+        self.start_with_config(None).await
+    }
```
oblique, line 54: I think this should be remove and rename `start_with_config` to `start`. In general we should give control to the user when they have to choose non-optional configs. For example a user should explicitly set the `network`.

sebasti810: I see. I thought that since the user already has to set the network in the `new` function, we can make a `start() `function that doesn't change anything in the rest of the optional configuration, but I'll change that. Apart from that, are the configuration options sufficient?

node-uniffi/src/types.rs
sebasti810: We can, but I haven't yet found a solution to make PeerId from libp2p and Vec<u16,u16> compatible with uniffi without the specific wrappers i wrote in types.rs. Afaik, e.g. tuple types are not directly supported because they cannot be cleanly mapped to equivalent types in different target languages. Of course, I could bring the wrappers into the lumina node if that's wanted :)

node-uniffi/src/lib.rs
oblique: Another approach, which I think we prefer, is to have the whole `NodeStartConfig` in `new` and then just have `start` (but not `start_with_config`). Also in this case it would make sense to rename the config struct to `NodeConfig`.

> Apart from that, are the configuration options sufficient?

Yes looks ok.

zvolin: > to have the whole NodeStartConfig in new and then just have start without the config

I'm curious. If we could also remove `start` and start the node right away within the `new`, and then make `stop` take `self`, we could get rid of `Mutex<Option<...>>` entirely.

I know we use the separate `new` and `start` pattern but that's because we use a `WebWorker` there. Would this work in case of uniffi? or the semantics of other languages like swift and kotlin don't play well with `stop(self)`?

sebasti810: I think the main problem here is that uniffi automatically wraps objects in an Arc if there are functions that need ownership. We'd probably need a construction like 
```rust
pub async fn stop(self: Arc<Self>) ... {
   match Arc::try_unwrap(self) { 
      Ok(node) =>  node.node.stop().await
      ...
} 
```
But even then, I'm not sure if it would work. Should I give it a try?

sebasti810: i agree, i played around with it, it's getting messy 😅

node-uniffi/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+        })
+    }
+
+    /// Starts the node and connects to the network.
+    pub async fn start(&self) -> Result<bool> {
+        let mut node_lock = self.node.write().await;
+        if node_lock.is_some() {
+            return Err(LuminaError::NetworkError {
+                msg: "Node is already running".to_string(),
+            });
```
oblique, line 52: I think it would less boilerplate to do one of the following in all errors that have a description message:

1. Use tuple instead. For example `LuminaError::NetworkError(String)`

2. If you want to keep the `msg` field then write some helpers. For example:

```rust
impl LuminaError {
    pub(crate) fn network_error(msg: &str) -> LuminaError {
        LuminaError::NetworkError { msg: msg.to_owned() }
    }
}
```

And then use it as:

```rust
return Err(LuminaError::network_error("Node is already running"));
```

Also, another thing, since `LuminaError` contains the `Error` in its name I prefer one of the two:

1. `LuminaError::Network`
2. `LuminaError::NetworkFailure`

node-uniffi/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+        let node = node
+            .as_ref()
+            .ok_or(LuminaError::network("Node not initialized"))?;
+
+        let metadata = node.get_sampling_metadata(height).await?;
+        Ok(metadata.map(|m| serde_json::to_string(&m).unwrap()))
+    }
+
+    /// Returns the next event from the node's event channel.
+    pub async fn events_channel(&self) -> Result<Option<NodeEvent>> {
```
zvolin, line 303: nit
```suggestion
    pub async fn next_event(&self) -> Result<Option<NodeEvent>> {
```

node-uniffi/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+        let mut events_subscriber = self.events_subscriber.lock().await;
+        match events_subscriber.as_mut() {
+            Some(subscriber) => match subscriber.try_recv() {
+                Ok(event) => Ok(Some(event.event.into())),
+                Err(e) => Err(LuminaError::network(e.to_string())),
+            },
+            None => Err(LuminaError::network("Node is not running")),
```
zvolin, line 310: since this function is async already and you used an async mutex, wouldn't it be better to `await subscriber.recv()` and return a `Result<NodeEvent>`?

Iirc the this can only fail if publisher is dropped, meaning that node was stopped, so we could return here `LuminaError::NodeNotRunning` if this happens.

node-uniffi/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+            events_subscriber: Mutex::new(None),
+            config,
+        })
+    }
+
+    /// Starts the node and connects to the network.
+    pub async fn start(&self) -> Result<bool> {
+        let mut node_lock = self.node.write().await;
+        if node_lock.is_some() {
+            return Err(LuminaError::network("Node is already running"));
```
zvolin, line 50:
```suggestion
            return Err(LuminaError::AlreadyRunning);
```

node-uniffi/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /// Stops the running node and closes all network connections.
+    pub async fn stop(&self) -> Result<()> {
+        let mut node = self.node.write().await;
+        if let Some(node) = node.take() {
+            node.stop().await;
+            Ok(())
+        } else {
+            Err(LuminaError::network("Node is already stopped"))
```
zvolin, line 69:
```suggestion
            Err(LuminaError::NodeNotRunning)
```

node-uniffi/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+    pub async fn is_running(&self) -> bool {
+        self.node.read().await.is_some()
+    }
+
+    /// Gets the local peer ID as a string.
+    pub async fn local_peer_id(&self) -> Result<String> {
+        let node = self.node.read().await;
+        let node = node
+            .as_ref()
+            .ok_or(LuminaError::network("Node not initialized"))?;
```
zvolin, line 83:
```suggestion
            .ok_or(LuminaError::NodeNotRunning)?;
```
this also applies on all other occurences below

node-uniffi/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+    },
+
+    /// Error returned when trying to start a node that's already running
+    #[error("Node is already running")]
+    AlreadyRunning,
+
+    /// Error returned when a mutex lock operation fails
+    #[error("Lock error")]
+    Lock,
+
```
zvolin, line 35: I think we only use infallible locks now

zvolin (commented):
looks very nice already, some nits from me

node-uniffi/src/lib.rs
sebasti810: good catch! ty

zvolin:
don't bother with this clippy failure, we have a new toolchain an ci related changes elsewhere and gonna get it in soon. The team is back so we are gonna address this PR soon, thanks for patience :pray:

node-uniffi/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "lumina-node-uniffi"
+version = "0.1.0"
+edition = "2021"
+description = "Mobile bindings for Lumina node"
```
zvolin, line 5:
```suggestion
description = "Mobile bindings for Lumina node"
publish = false
```
Could you disable publishing for now? I'm not sure if there is a usecase for the uniffi crate being on crates.io, and the `release-plz` tool we are using would publish it right away if merged

zvolin:
Ooops, I wanted to add `lumina-node-uniffi` to the clippy and test jobs on ci and I forgot to write `git push sebasti810 main` but just typed `git push` and it merged the PR :man_facepalming:. The coincidence was that I was granted admin to tweak the CI required checks and admin role has ability to bypass required checks (like approvals). So the lesson is never give admin to anyone :)
Anyway, we have discussed the PR on daily and we agreed that we want to approve and merge it, so it is okay :upside_down_face: I wanted to approve it right after this push anyway.

Thank you for your contribution! It's an amazing feature :pray:

PR #476, opened 2024-12-12
feat: Add remaining node types for wasm

fl0rek:


types/src/extended_header.rs
```diff
@@ -61,14 +65,76 @@ const VERIFY_CLOCK_DRIFT: Duration = Duration::from_secs(10);
...
+///
+/// // fetch new header
+/// let fetched_header = some_untrusted_header();
+///
+/// fetched_header.validate().expect("Invalid block header");
+/// genesis_header.verify(&fetched_header).expect("Malicious header received");
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg(all(feature = "wasm-bindgen", target_arch = "wasm32"))]
+#[wasm_bindgen]
```
zvolin, line 125: I think the main point of making it wasm bindgenable was to also expose some methods on it, also getters for jsoned fields. without that, this looks less usable than just a json we had before

types/src/extended_header.rs
```diff
@@ -61,14 +65,76 @@ const VERIFY_CLOCK_DRIFT: Duration = Duration::from_secs(10);
...
+    #[wasm_bindgen(skip)]
+    pub header: Header,
+    /// Commit metadata and signatures from validators committing the block.
+    #[wasm_bindgen(skip)]
+    pub commit: Commit,
+    /// Information about the set of validators commiting the block.
+    #[wasm_bindgen(skip)]
+    pub validator_set: ValidatorSet,
+    /// Header of the block data availability.
+    #[wasm_bindgen(skip)]
```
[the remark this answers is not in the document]
fl0rek: Since DAH is our own struct, we could wasm_bindgen it, so getter with clone is correct here, I believe.

cli/js/index.js
```diff
@@ -26,8 +26,8 @@ async function showStats(node) {
     return;
   }
 
-  const squareRows = networkHead.dah.row_roots.length;
-  const squareCols = networkHead.dah.column_roots.length;
+  const squareRows = networkHead.dah.row_roots().length;
+  const squareCols = networkHead.dah.column_roots().length;
```
oblique, line 30: I'm not sure if I like this breaking change because JS it doesn't give even a runtime error if user does not migrate.

types/src/data_availability_header.rs
```diff
@@ -221,6 +227,64 @@ impl DataAvailabilityHeader {
     }
 }
 
+#[cfg(all(feature = "wasm-bindgen", target_arch = "wasm32"))]
+#[wasm_bindgen]
+impl DataAvailabilityHeader {
+    /// Merkle roots of the [`ExtendedDataSquare`] rows.
+    #[wasm_bindgen(js_name = row_roots)]
```
zvolin, line 234: all those `js_name`s should use camel case

types/src/extended_header.rs
```diff
@@ -391,6 +443,99 @@ impl ExtendedHeader {
...
+
+    /// Verify a chain of adjacent untrusted headers.
+    ///
+    /// # Errors
+    ///
+    /// If verification fails, this function will return an error with a reason of failure.
+    /// This function will also return an error if untrusted headers are not adjacent
+    /// to each other.
+    #[wasm_bindgen(js_name = verify_range)]
+    pub fn js_verify_range(&self, untrusted: Vec<ExtendedHeader>) -> Result<(), JsValue> {
```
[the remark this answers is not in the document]
zvolin: I gave it a shot and I'm afraid there's no way around this. Using approach from [here](https://github.com/rustwasm/wasm-bindgen/issues/3107#issuecomment-1729050276) and accepting `&ExtendedHeaderArray` preserves the value correctly on javascript side, however then there's no good way for getting `ExtendedHeader` from `JsValue`. Downcasting is [not currently possible](https://github.com/rustwasm/wasm-bindgen/issues/2231) and `serde_wasm_bindgen` doesn't seem to work well between wasm_bindgen derive and our serde magic here. I think we could go through string, which is far from ideal.
Maybe we should just document it and require a deep clone here?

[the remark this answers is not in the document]
zvolin: https://github.com/fl0rek/celestia-node-rs/compare/feat/wasm-types-coverage...zvolin:lumina:feat/wasm-types-coverage-test?expand=1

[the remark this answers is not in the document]
zvolin: Or, let's just keep a single header verify. Users can trust node stuff they get from it is verified and these methods are consumed by node itself

cli/js/index.js
zvolin: we're gonna need to document that in the changelog

types/src/extended_header.rs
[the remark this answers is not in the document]
zvolin: I added a clone and doc how to use it

types/src/extended_header.rs
```diff
@@ -391,6 +443,117 @@ impl ExtendedHeader {
...
+        Ok(self.verify(untrusted)?)
+    }
+
+    /// Verify a chain of adjacent untrusted headers.
+    ///
+    /// # Errors
+    ///
+    /// If verification fails, this function will return an error with a reason of failure.
+    /// This function will also return an error if untrusted headers are not adjacent
+    /// to each other.
```
oblique, line 526: Can you add the note also here?

zvolin: done

PR #477, opened 2024-12-12
chore(node,node-wasm)!: Rename `syncing_window` to `sampling_window`

oblique:


node/src/node.rs
```diff
@@ -158,14 +158,16 @@ where
             p2p: p2p.clone(),
             event_pub: event_channel.publisher(),
             batch_size: config.sync_batch_size,
-            syncing_window: config.syncing_window,
+            // We sync only what we need to sample. So syncing_window is
+            // the same as sampling_window.
+            syncing_window: config.sampling_window,
```
oblique, line 163: I kept the `syncing_window` name here because it is relevant to the `Syncer`. But I can change it if you prefer. Wdyt?

zvolin: it's ok

zvolin (approved):
:+1:

fl0rek (approved):
:raised_hands:

PR #480, opened 2024-12-12
feat(types): add serde feature for nmt-rs

S1nus:
Adding the `serde` feature to the `nmt-rs` dep allows me to use the serde methods of nmt-rs when `celestia-types` is imported as a library

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ tendermint = { version = "0.40.0", default-features = false }
 tendermint-proto = "0.40.0"
 
 libp2p = "0.54.0"
-nmt-rs = "0.2.1"
+nmt-rs = {version = "0.2.1", features=["serde"]}
 prost = "0.13.3"
 prost-build = "0.13.3"
 prost-types = "0.13.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ tendermint = { version = "0.40.0", default-features = false }
 tendermint-proto = "0.40.0"
 
 libp2p = "0.54.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -14,7 +14,7 @@ tendermint = { version = "0.40.0", default-features = false }
 tendermint-proto = "0.40.0"
 
 libp2p = "0.54.0"
-nmt-rs = "0.2.1"
+nmt-rs = {version = "0.2.1", features=["serde"]}
```
zvolin, line 17: a nitpick, could you instead add the feature flag only in `celestia-types`?

zvolin: + please add spaces around inside curly braces. it's not on ci unfortunately (yet) but we format with taplo

S1nus: > a nitpick, could you instead add the feature flag only in `celestia-types`?

is this even possible? how do i do that

zvolin: Something like `nmt-rs = { workspace = true, features = ["serde"] }` in celestia types cargo toml should do. It's likely gonna have same effect as your current change but if in future we add a crate that doesn't depend on celestia types, it won't automatically inherit this feature

zvolin: Tendermint dep is good example of this pattern here

S1nus: ok got it

zvolin (approved):

oblique (approved):

zvolin:
it looks like a commit signing check failed for https://github.com/eigerco/lumina/pull/480/commits/3511733fb58ed45bae50c7fe43f84d1841345e80

PR #493, opened 2024-12-24
fix(types): RowNamespaceData binary deserialization

citizen-stig:
I've encountered following error when have been deserializing `NamespaceData`. After some investigation I've noticed that problem stems from `RowNamespaceData`

```
called `Result::unwrap()` on an `Err` value: InvalidBoolEncoding(2)
```

Reproducible test is added as part of this PR. You can comment out actual changes and see how it fails.

I would suggest check all types that have `serde(into` to be checked against binary serialization and have symmetrical `from` or `try_from`. Those types are:

* `ExtendedDataSquare`
* `Row`
* `Sample`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -578,6 +578,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


zvolin (approved):
thanks. Feel free to open an issue about other types that you identified. We mainly support the binary encoding of those through `.encode()` and `.decode()`, serde is there just because it's reused in json rpc

PR #495, opened 2024-12-28
Adding symmetrical serialization to `eds`, `row` and `sample`

citizen-stig:
Follow up of https://github.com/eigerco/lumina/pull/493

### DIFF
--- a/types/src/eds.rs
+++ b/types/src/eds.rs
@@ -131,7 +131,7 @@ impl Display for AxisType {
 /// [`Share`]: crate::share::Share
 /// [`DataAvailabilityHeader`]: crate::DataAvailabilityHeader
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
-#[serde(into = "RawExtendedDataSquare")]
+#[serde(into = "RawExtendedDataSquare", try_from = "RawExtendedDataSquare")]
 pub struct ExtendedDataSquare {
     /// The raw data of the EDS.
     data_square: Vec<Share>,
@@ -465,6 +465,14 @@ impl From<ExtendedDataSquare> for RawExtendedDataSquare {
                 .collect(),
             codec: eds.codec,
         }
+    }
+}
+
+impl TryFrom<RawExtendedDataSquare> for ExtendedDataSquare {
+    type Error = Error;
+
+    fn try_from(raw: RawExtendedDataSquare) -> std::result::Result<Self, Self::Error> {
+        ExtendedDataSquare::from_raw(raw, AppVersion::V2)
     }
 }
 
--- a/types/src/row.rs
+++ b/types/src/row.rs
@@ -50,7 +50,7 @@ pub struct RowId {
 
 /// Row together with the data
 #[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(into = "RawRow")]
+#[serde(into = "RawRow", try_from = "RawRow")]
 pub struct Row {
     /// Shares contained in the row
     pub shares: Vec<Share>,
@@ -164,6 +164,18 @@ impl From<Row> for RawRow {
             shares_half,
             half_side: RawHalfSide::Left.into(),
         }
+    }
+}
+
+impl TryFrom<RawRow> for Row {
+    type Error = Error;
+
+    fn try_from(raw: RawRow) -> std::result::Result<Self, Self::Error> {
+        let mut shares = Vec::with_capacity(raw.shares_half.len());
+        for shr in raw.shares_half {
+            shares.push(Share::try_from(shr)?);
+        }
+        Ok(Row { shares })
     }
 }
 
--- a/types/src/sample.rs
+++ b/types/src/sample.rs
@@ -40,7 +40,7 @@ pub struct SampleId {
 
 /// Represents Sample, with proof of its inclusion
 #[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(into = "RawSample")]
+#[serde(into = "RawSample", try_from = "RawSample")]
 pub struct Sample {
     /// Indication whether proving was done row or column-wise
     pub proof_type: AxisType,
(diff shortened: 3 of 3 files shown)


citizen-stig:
Tagging @zvolin for visibility

types/src/eds.rs
```diff
@@ -468,6 +468,14 @@ impl From<ExtendedDataSquare> for RawExtendedDataSquare {
     }
 }
 
+impl TryFrom<RawExtendedDataSquare> for ExtendedDataSquare {
+    type Error = Error;
+
+    fn try_from(raw: RawExtendedDataSquare) -> std::result::Result<Self, Self::Error> {
+        ExtendedDataSquare::from_raw(raw, AppVersion::V2)
```
zvolin, line 475: I don't think we can assume any particular `AppVersion` here, and this is likely the reason why there is no `Deserialize` derived here nor the `try_from`. I don't have any idea if we could do better there than just require using `ExtendedDataSquare::from_raw`

types/src/row.rs
```diff
@@ -167,6 +167,18 @@ impl From<Row> for RawRow {
...
 }
 
+impl TryFrom<RawRow> for Row {
+    type Error = Error;
+
+    fn try_from(raw: RawRow) -> std::result::Result<Self, Self::Error> {
+        let mut shares = Vec::with_capacity(raw.shares_half.len());
+        for shr in raw.shares_half {
+            shares.push(Share::try_from(shr)?);
+        }
```
zvolin, line 177: this isn't correct and again I'm not sure if we can do better than `Row::from_raw`. Row is a structure which holds the whole row from EDS, but in it's serialized form, it drops a half which needs to be reconstructed using leopard RS. However with just a plain `shares_half`, it is not possible to know which side needs to be reconstructed.

types/src/sample.rs
```diff
@@ -210,6 +210,20 @@ impl From<Sample> for RawSample {
     }
 }
 
+impl TryFrom<RawSample> for Sample {
+    type Error = Error;
+
+    fn try_from(raw: RawSample) -> std::result::Result<Self, Self::Error> {
+        let share = raw.share.ok_or(Error::MissingShares)?.try_into()?;
```
zvolin, line 217: this won't work for parity shares, and there is again not much information to judge if share should be treated as parity or not

zvolin (commented):
After looking through the PR, I'm not sure if there is a way to support this correctly. It's a problem we're facing porting celestia to rust that the structures here aren't self contained and often require additional information to make sure everything is sound. See https://github.com/celestiaorg/celestia-node/issues/3977 for EDS and https://github.com/celestiaorg/CIPs/pull/128 for shwap types.

I'm happy to revisit this if you find a way to make it correct, but I'm afraid it's not possible with current state as types are not self contained. With current way celestia operates, the only correct (and binary) encoding of those is through `.encode` and `.decode`, passing extra infos through corresponding `Id` types

types/src/eds.rs
zvolin: In the [rpc](https://github.com/eigerco/lumina/blob/main/rpc/src/share.rs#L65) we work this around by requiring whole `ExtendedHeader` instead of just height when requesting EDS (and other share related stuff). We made an exclusion for the `RowNamespaceData` as requesting it for parity shares through celestia-node rpc is not supported

citizen-stig:
> After looking through the PR, I'm not sure if there is a way to support this correctly. It's a problem we're facing porting celestia to rust that the structures here aren't self contained and often require additional information to make sure everything is sound. See [celestiaorg/celestia-node#3977](https://github.com/celestiaorg/celestia-node/issues/3977) for EDS and [celestiaorg/CIPs#128](https://github.com/celestiaorg/CIPs/pull/128) for shwap types.
> 
> I'm happy to revisit this if you find a way to make it correct, but I'm afraid it's not possible with current state as types are not self contained. With current way celestia operates, the only correct (and binary) encoding of those is through `.encode` and `.decode`, passing extra infos through corresponding `Id` types


Ok, I see the issue with hard coding particular version of app into the deserialization.

But what can be workaround here? We rely on support of serde to serialize/deserialize those things in binary format. 

If `.encode` is used in reality anyway, is it possible to remove `#[serde(into = "RawExtendedDataSquare", try_from = "RawExtendedDataSquare")]` so at least simmetrical binary serialization/deserialization works? 

I guess we can use `.encode` and `.decode` in out types, if we manually implement serde derives. I will need to look into it. Thank you for advices!

zvolin:
> If .encode is used in reality anyway, is it possible to remove #[serde(into = "RawExtendedDataSquare", try_from = "RawExtendedDataSquare")] so at least simmetrical binary serialization/deserialization works?

There is no symmetry in case of EDS because `Deserialize` is not implemented for it at all. I think that'd be the correct way also for `Sample` and `Row`, to just remove the `Deserialize` from them. They all serialize into `Raw` format, and then can be deserialized into `Raw` again but requiring additional info to reconstruct original type using `from_raw`.

If you need serde, I think you could wrap them in a struct containing also additional information required to make the deserialization correct

zvolin:
We don't like this too really... and we would love to have all the types self contained. It's mostly an idiomacy differences between go and rust, where in Go they happily serialize / deserialize everything and validate constrains where needed, and in Rust the idiomatic way is to not allow constructing invalid state at all

citizen-stig:
> > If .encode is used in reality anyway, is it possible to remove #[serde(into = "RawExtendedDataSquare", try_from = "RawExtendedDataSquare")] so at least simmetrical binary serialization/deserialization works?
> 
> There is no symmetry in case of EDS because `Deserialize` is not implemented for it at all. I think that'd be the correct way also for `Sample` and `Row`, to just remove the `Deserialize` from them. They all serialize into `Raw` format, and then can be deserialized into `Raw` again but requiring additional info to reconstruct original type using `from_raw`.
> 
> If you need serde, I think you could wrap them in a struct containing also additional information required to make the deserialization correct

Good point about `EDS`, missed that it does not have Deserialize. 
And totally agree on removing `Deserialize` from `Row` and `Sample`.

I will try to go that route and encapsulate all needed information in custom deserialization for types that are used in our adapter

PR #538, opened 2025-02-07
feat(types): Make fields of `ShareProof` public

S1nus:
I need to construct a ShareProof from its parts. There is no constructor besides the unfriendly protobuf types.


zvolin (approved):

fl0rek (approved):
:guitar:

oblique (approved):

PR #541, opened 2025-02-12
test(types): fix flaky blob reconstruct when len == 0

zvolin:


types/src/blob.rs
```diff
@@ -489,7 +489,7 @@ mod tests {
 
     #[test]
     fn reconstruct_not_sequence_start() {
-        let len = rand::random::<usize>() % 1024 * 1024;
+        let len = rand::random::<usize>() % (1024 * 1024) + 1;
```
fl0rek, line 492: after double checking operator precedence etc.. maybe it could be just:
```suggestion
        let len : usize = rand::random_range(1..=(1024*1024) + 1);
```
Upper and lower limits are immediately visible this way, no need to do the maths.

zvolin: sounds good, but for this we'd need to upgrade rand

oblique: @fl0rek In your suggestion the `+ 1` should be removed

oblique (approved):

fl0rek (approved):
lol, nice catch :+1:

PR #543, opened 2025-02-20
feat(rpc): Add Blobstream API

Ferret-san:
Adds the `Blobstream` trait for the `Client` to be able to make blobstream related requests. The blobstream service for celestia-node can be found [here](https://github.com/celestiaorg/celestia-node/tree/main/nodebuilder/blobstream)

```
#[rpc(client)]
pub trait Blobstream {
    /// GetDataRootTupleRoot retrieves the data root tuple root for a given range from start to end
    #[method(name = "blobstream.GetDataRootTupleRoot")]
    async fn get_data_root_tuple_root(
        &self,
        start: u64,
        end: u64,
    ) -> Result<DataRootTupleRoot, Error>;

    /// GetDataRootTupleInclusionProof returns a data root tuple inclusion proof for a given height
    /// between a range from start to end
    #[method(name = "blobstream.GetDataRootTupleInclusionProof")]
    async fn get_data_root_tuple_inclusion_proof(
        &self,
        height: u64,
        start: u64,
        end: u64,
    ) -> Result<DataRootTupleInclusionProof, Error>;
}
```


rpc/tests/utils/client.rs
```diff
@@ -45,7 +45,7 @@ pub async fn new_test_client(auth_level: AuthLevel) -> Result<Client> {
     let client = Client::new(&url, token.as_deref()).await?;
 
     // minimum 2 blocks
-    client.header_wait_for_height(2).await?;
```
zvolin, line 48: in this block, the account we use for blob submission should be founded with tia, why this change?

rpc/src/blobstream.rs
```diff
@@ -0,0 +1,33 @@
+use bytes::Bytes;
+use celestia_types::MerkleProof;
+use jsonrpsee::proc_macros::rpc;
+use prost::bytes;
+
+// DataRootTupleRoot is the root of the merkle tree created
+// from a set of data root tuples.
+pub type DataRootTupleRoot = Bytes;
```
zvolin, line 8: if it is a hash, could we alias it to `celestia_types::Hash`?

rpc/src/blobstream.rs
```diff
@@ -0,0 +1,33 @@
...
+
+// DataRootTupleInclusionProof is the binary merkle
+// inclusion proof of a height to a data commitment.
+pub type DataRootTupleInclusionProof = MerkleProof;
```
zvolin, line 12:
```suggestion
/// DataRootTupleRoot is the root of the merkle tree created
/// from a set of data root tuples.
pub type DataRootTupleRoot = Bytes;

/// DataRootTupleInclusionProof is the binary merkle
/// inclusion proof of a height to a data commitment.
pub type DataRootTupleInclusionProof = MerkleProof;
```

you need 3`/` to make it a doc-comment

rpc/src/blobstream.rs
```diff
@@ -0,0 +1,33 @@
+use bytes::Bytes;
```
zvolin, line 1:
```suggestion
//! celestia-node rpc types and methods related to blobstream

use bytes::Bytes;
```

rpc/tests/blobstream.rs
```diff
@@ -0,0 +1,31 @@
...
+#[tokio::test]
+async fn get_data_root_tuple_inclusion_proof() {
+    let client = new_test_client(AuthLevel::Write).await.unwrap();
+
+    let result = client
+        .get_data_root_tuple_inclusion_proof(4211299, 4211298, 4211300)
+        .await
+        .unwrap();
+
+    println!("data root tuple root: {:x?}", result);
```
zvolin, line 30: I think I would combine those tests into a single one where we request a proof and the root and just call `proof.verify(root).unwrap()`

rpc/tests/blobstream.rs
```diff
@@ -0,0 +1,31 @@
...
+pub mod utils;
+
+use crate::utils::client::{new_test_client, AuthLevel};
+
+#[tokio::test]
+async fn get_data_root_tuple_root() {
+    let client = new_test_client(AuthLevel::Write).await.unwrap();
+
+    let result = client
+        .get_data_root_tuple_root(4211296, 4211298)
```
zvolin, line 14: those tests seem to fail on CI.
```
called `Result::unwrap()` on an `Err` value: Call(ErrorObject { code: ServerError(1), message: "end block 4211300 is higher than local chain height 28. Wait for the node until it syncs up to 4211300", data: None })
```
could you instead make `header_get_network_head()` request and adjust height accordingly?

zvolin (commented):
thanks! left some ideas

rpc/src/blobstream.rs
```diff
@@ -0,0 +1,33 @@
...
+    async fn get_data_root_tuple_root(
+        &self,
+        start: u64,
+        end: u64,
+    ) -> Result<DataRootTupleRoot, Error>;
+
+    /// GetDataRootTupleInclusionProof returns a data root tuple inclusion proof for a given height
+    /// between a range from start to end
+    #[method(name = "blobstream.GetDataRootTupleInclusionProof")]
+    async fn get_data_root_tuple_inclusion_proof(
```
zvolin, line 27: one more thing, since those will be implicitly exposed on any jsonrpsee client, we have a convention of prefixing methods with the modeule name. Could you rename those to start with `blobstream_`?

rpc/tests/utils/client.rs
Ferret-san: when testing against mocha, if I had my node sync from a recent height and trusted hash, waiting for height 2 would trigger a warning on the node `header not found` (celestia-node), so removing it allowed me to test this, but we could have an env for this height and that way the test does not assume you have a node synced from genesis etc

rpc/src/blobstream.rs
Ferret-san: yes can def do that

rpc/tests/blobstream.rs
Ferret-san: yeah I pushed and opened the PR before making the test more robust if you will, let me get this under a single test :saluting_face:

rpc/tests/blobstream.rs
Ferret-san: noted!

rpc/src/blobstream.rs
Ferret-san: so `blobstream_get_data_root_tuple_root`?

rpc/tests/utils/client.rs
zvolin: We can instead do a loop here, where we request network head and break if it is >2 and sleep otherwise

Ferret-san: revisiting here, I think the problem with that (from my own testing), is that if I sync a node from a trusted hash and you request to wait for the block height `2`, the node returns "header not found". Maybe having an env var to control which height to wait for gives more flexibility?

rpc/src/blobstream.rs
Ferret-san: returning hash wont work since its 32 bytes and celestia-node returns a base64 44 byte length string, so `Bytes` does the job here (a user just has to decode it accordingly)

oblique (approved):
LGTM

fl0rek (approved):
🎸

PR #544, opened 2025-02-23
chore(ci): fix outdated lockfiles when releasing to npm

zvolin:
Currently we update package.json with new version of node-wasm and commit that from release-plz PR. However we can't keep lock file up to date this way, because at this time, the new version is not yet in npm.

To fix that, we depend on node-wasm based on the path and update the lockfile in release-plz PR. The actual version-based dependency is only set temporarily when we are releasing to npm.


fl0rek (approved):
:crossed_fingers:

Issue #22, opened 2023-07-04
Implement RPC Celestia api in celestia-rpc


zvolin:

oblique:
TODO list:

- [x] [Blob](https://node-rpc-docs.celestia.org/#blob)
- [ ] [Daser](https://node-rpc-docs.celestia.org/#daser)
- [ ] [Fraud](https://node-rpc-docs.celestia.org/#fraud)
- [x] [Header](https://node-rpc-docs.celestia.org/#header)
- [ ] [Node](https://node-rpc-docs.celestia.org/#node)
- [ ] [P2P](https://node-rpc-docs.celestia.org/#p2p)
- [x] [Share](https://node-rpc-docs.celestia.org/#share)
- [ ] [State](https://node-rpc-docs.celestia.org/#state)

PR #557, opened 2025-03-02
feat(rpc, types): add DAS module with sampling stats and catch-up methods

YakupAltay:
Hey @zvolin!

While scanning through the issues, I noticed that not all Celestia [Node API methods](https://node-rpc-docs.celestia.org/?version=v0.20.4) have been implemented in [this issue](https://github.com/eigerco/lumina/issues/22).

I thought it would be helpful to add them step by step, so I started by implementing the two DAS methods. I’d love to get your feedback on whether this is the best approach moving forward.

Looking forward to your thoughts! I'm excited to contribute more.


fl0rek:
@YakupAltay thanks for the contribution! :guitar: 

Looking at the CI failure, it seems you need to add a light node to our [docker-compose](https://github.com/eigerco/lumina/blob/main/ci/docker-compose.yml) setup. Please see Dockerfiles and shell scripts in that directory for reference, light node setup should hopefully look very similar. Or ping us if you need assistance :)

Additionally we require commits to be [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), so would you mind setting that up on your end, rebasing to sign commits you've pushed and force pushing to replace them?

YakupAltay:
Hello @fl0rek!

Thanks for your insights.

I’ve tried to add a light node to the CI. Even though it should be very similar to the bridge node setup, I haven't been able to get the light node running yet. It keeps failing with the following error:
`Error: node: failed to start: no trusted peers`

It seems like the light node needs a boot node to start, and I’m trying to retrieve it from bridge-0. However, the following command fails:
`peer_id=$(celestia p2p info --url 'ws://bridge-0:26658' | jq -r '.result.id')`
It returns:
`Error: cant access the auth token: token/node-store flag was not specified: no opened Node Store found (no node is running)`

This happens even though the bridge node is definitely running.

I'd really appreciate any suggestions or pointers you might have! 🙏

fl0rek:
@YakupAltay 
I think I've been able to find the problem. `celestia p2p info` invocation calls bridge node which has a skip auth set, but this call still uses credentials that were generated for other bridge nodes. This works, because auth is skipped, but files itself are absent when running light node in your case, causing error. 
I've been able to work around that by passing a dummy auth token :sweat_smile: 
```
peer_id=$(celestia p2p info --token 'foo' --url 'ws://bridge-0:26658' | jq -r '.result.id')
```
[Link to the entire thing](https://github.com/eigerco/lumina/compare/main...fl0rek:celestia-node-rs:test/ci-add-light-node?expand=1), I _think_ it's working :P

YakupAltay:
Congrats!! 🎉

I have spent almost a day to it and could not realize the dummy auth token, haha.

When do you plan to merge it with the main branch? So that I can re-commit my changes with verified commits.

Would love to move forward to the other Node API methods! 🚀

fl0rek:
Ah I though you'd just copy whatever you need from my my branch. Not like I
did particularly original work there :p

Otherwise I can clean it up and open the PR and try to squeeze it in
tomorrow

YakupAltay:
Then, if it is okay for you @fl0rek, I can copy it from your branch and do it from scratch on my side.

Would that be a problem?

fl0rek:
Not at all, please do 👍

types/src/das.rs
zvolin, line 1: Since those types consists mostly of stuff that is an implementation detail of `DASer` of celestia-node, I think I would like to keep them directly in `rpc/src/das.rs`. They are not generic enough to re-use them for e.g. lumina's daser

types/src/das.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub current: u64,
+    /// The starting height of the job.
+    pub from: u64,
+    /// The ending height of the job.
+    pub to: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Contains statistics related to the DAS process.
+pub struct SamplingStats {
```
zvolin, line 18: can you reorder doc-comment and derive directive? I think everywhere else we keep doc comments first, and then derive

zvolin (commented):
Hey, I was sick last week. Thanks for the PR, definitely makes sense to start filling those gaps in celestia-rpc.

One note when implementing the docker compose changes to run light node, I think I would like to keep a single dockerfile and script for light and bridge nodes, and introduce a `NODE_TYPE` variable inside compose service to choose which node type to run

zvolin:
> I have spent almost a day to it and could not realize the dummy auth token, haha.

may be worth opening an issue in celestia-node, it's clearly a bug there

zvolin:
One more thing, we require commits to be signed to merge to main, could you get Verified status on yours?
image

YakupAltay:
Hey @zvolin,

I hope you’re feeling better now — wishing you a smooth recovery!

I've just pushed the latest changes as requested. The light node should now run properly, and _the DAS tests are passing with the correct configuration_.

**The main issue** at this point is that some tests need to run under the _bridge_ setup and others under the _light_ setup. I haven’t figured out how to separate those yet, but I’m working on it. **_Would love to hear your ideas though_**.

Regarding the errors in the last commit — they’re due to **not authorizing myself with the container registry**. I’ll fix that ASAP.

Once everything looks good on your end, it would be best to close this PR. I’ll then open a new one with verified commits.

Thanks!

zvolin:
I started checking the changes required to run light nodes in our CI setup and I think I'd prefer them to land in a separate PR. Let me see if I can open one quickly

> they’re due to not authorizing myself with the container registry.

I think the issue is elsewhere

YakupAltay:
I did not really change anything on my side then.

The only thing coming to my mind is that **bridge-0.key** and **light-0.key** are the same. However, I do not think that they created such a big issue.

I am frustrated to have these problems with my first PR, haha.

Waiting for your updates.

zvolin:
https://github.com/eigerco/lumina/pull/562

I opened this, it allows specifying node type by env var to avoid dockerfiles / scripts duplication. I need to find out what causes one test to fail, otherwise it seems to work

zvolin:
>  it would be best to close this PR. I’ll then open a new one with verified commits.

You could try running this to walk all your commits and sign them
```
git rebase --exec "git commit --amend --no-edit -n -S" -i e1a7644a
```
and then force history update in remote
```
git push --force-with-lease
```

> I am frustrated to have these problems with my first PR, haha.

Don't be, our setup is not straightforward and changing compose setup often breaks stuff in other places than you expect it to. I think this massive ci failure was because github ci wasn't aware of the need to build your new image [here](https://github.com/eigerco/lumina/blob/main/.github/actions/local-devnet/action.yml#L21), although more issues could be discovered later

> bridge-0.key and light-0.key are the same

this doesn't sound ideal, because they would end up with the same peer id? :thinking: 

> some tests need to run under the bridge setup and others under the light setup

I think it's fine assuming that DAS tests would be the only ones running with light node for now

YakupAltay:
> and then force history update in remote

Doing these right away.

> Don't be, our setup is not straightforward and changing compose setup often breaks stuff in other places than you expect it to.

Thanks for the explanation and understanding! 🙏

> this doesn't sound ideal, because they would end up with the same peer id?

I will now change the keys and re-commit. We shall see if the tests will fail or not.

YakupAltay:
Error again, @zvolin.

```
 light-0 Pulling 
 light-0 Warning pull access denied for light, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
 Network ci_default  Creating
 Network ci_default  Created
 Volume "ci_genesis"  Creating
 Volume "ci_genesis"  Created
 Volume "ci_credentials"  Creating
 Volume "ci_credentials"  Created
 Container ci-light-0-1  Creating
 Container ci-validator-1  Creating
 Container ci-bridge-0-1  Creating
 Container ci-grpcwebproxy-1  Creating
 Container ci-bridge-1-1  Creating
Error response from daemon: No such image: light:latest
```

What am I missing? It seems that every test is returning the same error.

zvolin:
We've merged #562. Could you remove the changes to the ci setup here and merge main? That should sort out the CI issues and allow you to use `node-2` which is light node for your tests.

zvolin:
> What am I missing? It seems that every test is returning the same error.

The error wasn't related to the keys, I just said that keys could cause other issues. As I wrote earlier, I think the error was coming from github actions setup:

> I think this massive ci failure was because github ci wasn't aware of the need to build your new image [here](https://github.com/eigerco/lumina/blob/main/.github/actions/local-devnet/action.yml#L21), although more issues could be discovered later

But it doesn't matter now as the needed CI changes are already on main

YakupAltay:
Hello hello @zvolin.

There should be no errors now, everything seems fine.

However, I changed the **celestia-rpc tests** to use _node-2 (light node)_ - which is more advantageous in the case of RPC tests _I presume_. And, as far as I know, there are no differences between light and bridge nodes other than the DAS.

What would you think? Is it applicable?

YakupAltay:
I also modified the _rpc test suite_ to automatically switch to **node-2** when testing DAS. Haven't committed those but they are _ready and working_.

Yet, I am **not** sure whether the RPC tests **should be separated**. Light node is handling them all.

zvolin:
>  I changed the celestia-rpc tests to use node-2 (light node) 

I think yeah, it makes sense that the most common use case for rpc would be speaking to the light node. I think it's fine since tests pass. I was a bit worried about stuff like `share.GetEDS` because light node doesn't store whole data squares, but it seem to work, at least for our small blocks in ci setup.

YakupAltay:
Nice!

Thanks for helping me out about the CI. I am happy that all are resolved now.

zvolin:
Sure, in case you missed it, there's still [one unattended comment](https://github.com/eigerco/lumina/pull/557#discussion_r1989001406)

YakupAltay:
> there's still https://github.com/eigerco/lumina/pull/557#discussion_r1989001406

There is nothing to be done specifically for that I guess. So, I resolved it. Hope that's not a problem.

YakupAltay:
Ah, sorry @zvolin. My bad. I overlooked the comment. 🙏

I have now moved `das-types` to `celestia-rpc`. It might not be the best way. Waiting for your response.

YakupAltay:
Silly me! 😄

I think there it is now in the way that you want @zvolin.

I am really sorry for being such a headache to you.

zvolin (approved):
nice! thanks for the work you put in the PR. Please don't feel discouraged, we can be picky on PRs but once you get used to our style, it should become more seamless :+1:

oblique (approved):
Amazing! Thanks a lot @YakupAltay!

PR #558, opened 2025-03-04
fix: Rework NodeWorker to use oneshot response channels

fl0rek:


fl0rek:
closing in favour of https://github.com/eigerco/lumina/pull/563

PR #559, opened 2025-03-11
feat(grpc): increase max message size to handle big celestia blocks

zvolin:


grpc/grpc-macros/src/lib.rs
```diff
@@ -58,9 +58,14 @@ impl GrpcMethod {
...
             pub #signature {
+                // 256 mb, future proof as celesita blocks grow
+                const MAX_MSG_SIZE: usize = 256 * 1024 * 1024;
+
                 let mut client = #grpc_client_struct :: new(
                     self.transport.clone(),
-                );
+                )
+                .max_decoding_message_size(MAX_MSG_SIZE)
+                .max_encoding_message_size(MAX_MSG_SIZE / 2);
```
oblique, line 68: Why is this the half?

zvolin: set them both to be equal :+1:

oblique (approved):

fl0rek (approved):
big blocks :ship:

PR #564, opened 2025-03-12
feat: lumina-utils crate

fl0rek:


zvolin (approved):
nice :fist_oncoming:

utils/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//! Utilities providing platform abstraction used across lumina project
+
+/// async executor platform independent utilities
+#[cfg(feature = "executor")]
+pub mod executor;
+#[cfg(test)]
```
oblique, line 6: i prefer a feature flag here

utils/src/token.rs
```diff
@@ -2,39 +2,41 @@ use std::fmt::{self, Debug};
...
     /// Returns a guard that will trigger the token on drop.
-    pub(crate) fn trigger_drop_guard(&self) -> TokenTriggerDropGuard {
+    pub fn trigger_drop_guard(&self) -> TokenTriggerDropGuard {
         TokenTriggerDropGuard {
             token: Some(self.token.clone()),
         }
     }
 
     /// Returns if event is triggered or not.
     #[allow(dead_code)]
```
oblique, line 32: this is not needed anymore

utils/src/token.rs
```diff
@@ -45,13 +47,15 @@ impl Debug for Token {
...
 
-pub(crate) struct TokenTriggerDropGuard {
+/// A wrapper for cancellation token which automatically cancels it on drop.
+pub struct TokenTriggerDropGuard {
     token: Option<CancellationToken>,
 }
 
 impl TokenTriggerDropGuard {
+    /// Disarms the guard, making it so that dropping it won't trigger cancellation
     #[allow(dead_code)]
```
oblique, line 57: this is not needed anymore

grpc/Cargo.toml
```diff
@@ -28,6 +28,7 @@ celestia-types.workspace = true
 prost.workspace = true
 tendermint-proto.workspace = true
 tendermint.workspace = true
+lumina-utils = { workspace = true, features = ["time", "test-utils"] }
```
oblique, line 31: here we should have only `time` and in the `dev-dependencies` the `test-utils`

grpc/tests/tonic.rs
```diff
@@ -5,20 +5,13 @@ use celestia_proto::cosmos::bank::v1beta1::MsgSend;
...
 
 use crate::utils::{new_grpc_client, new_tx_client, spawn};
 
-#[cfg(not(target_arch = "wasm32"))]
-use tokio::test as async_test;
-#[cfg(target_arch = "wasm32")]
-use wasm_bindgen_test::wasm_bindgen_test as async_test;
-
-#[cfg(target_arch = "wasm32")]
-wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
```
oblique, line 20: I think this is still needed

node/Cargo.toml
```diff
@@ -33,6 +33,7 @@ libp2p = { workspace = true, features = [
   "request-response",
   "kad",
 ] }
+lumina-utils = { workspace = true, features = ["executor", "token", "time", "test-utils"] }
```
oblique, line 36: Same as my comment above

utils/src/token.rs
```diff
@@ -2,39 +2,40 @@ use std::fmt::{self, Debug};
 
 use tokio_util::sync::CancellationToken;
 
-#[derive(Clone)]
-pub(crate) struct Token {
+/// A token which can be used to signal a cancellation request
```
oblique, line 5: We are wrapping `CancellationToken` mainly to make its name less confusing in the cases we don't use it for cancelling futures. I think we shouldn't use the word `cancel` in our docs. Check other places too

zvolin: maybe something along the lines "A token which can be used for signalling something exactly once"

fl0rek: Vague, I like it :P

oblique (approved):

PR #574, opened 2025-03-14
feat(types)!: add support for share v1 and signer in blob

zvolin:
- add `Blob::new_with_signer`
- add `Share::validate`
- add `Share::signer`
- add `From<[u8; 20]>`, `TryFrom<&[u8]>`, `TryFrom<Vec<u8>>` for `Address` types
- add `consts::appconsts::{SHARE_VERSION_ONE, SIGNER_SIZE}`

Breaking:
-  new pub field `Blob::signer`
-  `Commitment::from_blob` now requires `app_version` and `signer`
-  `Commitment::from_shares` now requires `app_version`
-  new errors `Error::{MissingSigner, SignerNotSupported}`


types/src/state/address.rs
```diff
@@ -157,173 +172,98 @@ impl From<Address> for Raw {
...
-    }
+            fn try_from(value: Vec<u8>) -> Result<Self> {
+                let len = value.len();
+                let id = value
+                    .try_into()
+                    .map_err(|_| Error::InvalidAddressSize(len))?;
+                Ok(Self::new(id))
+            }
+        }
+    };
```
zvolin, line 261: this was added

types/src/share.rs
```diff
@@ -252,6 +295,62 @@ mod tests {
...
+        assert_eq!(shares[0].namespace(), ns);
+        assert_eq!(shares[1].namespace(), ns);
+
+        assert_eq!(shares[0].info_byte().unwrap().version(), 1);
+        assert_eq!(shares[1].info_byte().unwrap().version(), 1);
+
+        assert!(shares[0].info_byte().unwrap().is_sequence_start());
+        assert!(!shares[1].info_byte().unwrap().is_sequence_start());
+
+        // share v0 doesn't have signer
```
oblique, line 316:
```suggestion
        // share v1 has signer only if it's the first share of the sequence
```

types/src/blob.rs
```diff
@@ -374,38 +422,129 @@ impl Blob {
...
+
+    /// This is the copy of the `Blob` struct, to perform additional checks during deserialization
+    #[derive(Serialize, Deserialize)]
+    pub(super) struct SerdeBlob {
+        namespace: Namespace,
+        #[serde(with = "base64string")]
+        data: Vec<u8>,
+        share_version: u8,
+        commitment: Commitment,
+        // note: celestia supports deserializing blobs without index, so we should too
```
oblique, line 511:
```suggestion
        // NOTE: celestia supports deserializing blobs without index, so we should too
```

types/src/share.rs
```diff
@@ -134,16 +145,44 @@ impl Share {
         }
     }
 
+    /// Get the `signer` part of share if it's first in the sequence of sparse shares and `share_version` supprots
+    /// it. Otherwise returns `None`
+    pub fn signer(&self) -> Option<AccAddress> {
+        let info_byte = self.info_byte()?;
+        if info_byte.is_sequence_start()
+            && info_byte.version() == appconsts::SHARE_VERSION_ONE
+            && !self.namespace().is_reserved()
```
oblique, line 154: I prefer something like `!self.is_compact()` or a comment in both cases where is_reserved is used. It tooks me a bit of time to cross check this line with the specification

types/src/blob.rs
```diff
@@ -256,7 +295,13 @@ impl Blob {
         // remove padding
         data.truncate(blob_len as usize);
 
-        Self::new(namespace, data, app_version)
+        if share_version == appconsts::SHARE_VERSION_ZERO {
+            Self::new(namespace, data, app_version)
+        } else if share_version == appconsts::SHARE_VERSION_ONE {
+            Self::new_with_signer(namespace, data, signer.expect("must be Some"), app_version)
```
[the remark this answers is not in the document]
zvolin: unwrap here is safe because we check if share is in user-namespace, it's v1 and sequence start, so it must be Some. But I agree returning error here will be more future-proof

oblique (approved):
LGTM

PR #580, opened 2025-03-22
feat: remove Deserialize from Row & Sample

### LINKED ISSUE #496: Remove `Deserialize` for `Row` and `Sample`
> If .encode is used in reality anyway, is it possible to remove #[serde(into = "RawExtendedDataSquare", try_from = "RawExtendedDataSquare")] so at least simmetrical binary serialization/deserialization works?

There is no symmetry in case of EDS because `Deserialize` is not implemented for it at all. I think that'd be the correct way also for `Sample` and `Row`, to just remove the `Deserialize` from them. They all serialize into `Raw` format, and then can be deserialized into `Raw` again but requiring additional info to reconstruct original type using `from_raw`.

If you need serde, I think you could wrap them in a struct containing also additional information required to make the deserialization correct

_Originally posted by @zvolin in https://github.com/eigerco/lumina/issues/495#issuecomment-2567627539_

varun-doshi:
Closes #496

### DIFF
--- a/types/src/row.rs
+++ b/types/src/row.rs
@@ -14,7 +14,7 @@ use celestia_proto::shwap::{row::HalfSide as RawHalfSide, Row as RawRow, Share a
 use cid::CidGeneric;
 use multihash::Multihash;
 use prost::Message;
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 
 use crate::consts::appconsts::SHARE_SIZE;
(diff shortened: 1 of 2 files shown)


fl0rek (approved):
Thanks for picking that up @varun-doshi! Looks good to me.

oblique (approved):
LGTM

Issue #570, opened 2025-03-14
rpc tests set auth tokens but were redirected to node with `skip-auth`


zvolin:
We should add tests which verify the auth token is used correctly,
and ethier remove[ `skip-auth` on `node-2`](https://github.com/eigerco/lumina/blob/c8f6ba9ce5ee5426cfcd3877c2a30173b93b5b43/ci/docker-compose.yml#L73) or remove logic of picking auth level in tests that are fine to run with `skip-auth`. Currently `gen_auth_tokens.sh` [generates tokens for `node-1`](https://github.com/eigerco/lumina/blob/c8f6ba9ce5ee5426cfcd3877c2a30173b93b5b43/tools/gen_auth_tokens.sh#L47), but in rpc tests [we hit `node-2`](https://github.com/eigerco/lumina/blob/c8f6ba9ce5ee5426cfcd3877c2a30173b93b5b43/rpc/tests/utils/client.rs#L12)

zvolin:
cc @YakupAltay

YakupAltay:
Hey @zvolin, thanks for pinging!

I've reviewed this and have some suggestions—would love your input on these:

We can enhance `gen_auth_tokens.sh` by generating auth tokens and removing `skip-auth` for both **node-1** (_bridge_) and **node-2** (_light_). This approach will simplify managing and tracking tokens. It will also help clarify [instructions in `README.md` for testing](https://github.com/eigerco/lumina/blob/main/README.md#running-go-celestia-node-for-integration) scenarios using either bridge or light nodes. Additionally, having both sets of tokens will streamline the use of tokens across different contexts (see [this](https://github.com/eigerco/lumina/blob/main/cli/src/native.rs#L194)), particularly for RPC methods exclusive to the light node.

Here's a proposed structure:

- node-1 (_bridge_): `CELESTIA_BRIDGE_NODE_AUTH_TOKEN_*`
- node-2 (_light_): `CELESTIA_LIGHT_NODE_AUTH_TOKEN_*`

Let me know your thoughts!

zvolin:
I think `node-1` doesn't have skip-auth set. We also need to keep skip-auth on `node-0` to hit it from wasm. I think I'd lean towards just adding tests (`rpc/tests/auth_token.rs`) which check the auth tokens generated for `node-1` to keep it simple, and remove the tokens logic from `rpc/tests/utils/client.rs`

zvolin:
[...]

PR #581, opened 2025-03-26
Auth Level Fix

YakupAltay:
Hi @zvolin,

Following my review of [issue #570](https://github.com/eigerco/lumina/issues/570), I've slightly adjusted the `auth_level` logic within `new_test_client`. The tests now correctly handle both scenarios:

- Light node (**node-2**) with the _skip-auth_ flag enabled.
- Bridge node (**node-1**) without the _skip-auth_ flag.

I'd greatly appreciate your feedback on this update before I move forward with the remaining RPC methods.

Thanks!


YakupAltay:
Oops, @zvolin. I could not understand why all the previous commits got added here.

Do you have a solution for that?

PR #582, opened 2025-03-26
test(rpc): Use `AuthLevel::Skip` and provide `new_test_client_with_url`

YakupAltay:
Hi @zvolin,

Following my review of https://github.com/eigerco/lumina/issues/570, I've slightly adjusted the `auth_level` logic within `new_test_client`. The tests now correctly handle both scenarios:

- Light node (`node-2`) with the **_skip-auth_** flag enabled.
- Bridge node (`node-1`) without the **_skip-auth_** flag.

I'd greatly appreciate your feedback on this update before I move forward with the remaining RPC methods.

Thanks!

### DIFF
--- a/rpc/tests/auth_token.rs
+++ b/rpc/tests/auth_token.rs
@@ -17,8 +17,8 @@ async fn blob_submit_using_bridge_node() {
     let data = random_bytes(5);
     let blob = Blob::new(namespace, data, AppVersion::V2).unwrap();
 
-    // AuthLevel::Skip
-    let unauthorized_client = new_test_client_with_url(AuthLevel::Skip, CELESTIA_BRIDGE_RPC_URL).await;
+    let unauthorized_client =
+        new_test_client_with_url(AuthLevel::Skip, CELESTIA_BRIDGE_RPC_URL).await;
     assert!(unauthorized_client.is_err());
 
     for auth_level in [AuthLevel::Read, AuthLevel::Write, AuthLevel::Admin] {
--- a/rpc/tests/auth_token.rs
+++ b/rpc/tests/auth_token.rs
@@ -6,7 +6,7 @@ use celestia_types::Blob;
 
 pub mod utils;
 
-use crate::utils::client::{blob_submit, new_test_client, AuthLevel};
+use crate::utils::client::{blob_submit, new_test_client_with_url, AuthLevel};
 use crate::utils::{random_bytes, random_ns};
 
 // Use node-1 (bridge node) as the RPC URL
@@ -14,7 +14,7 @@ const CELESTIA_BRIDGE_RPC_URL: &str = "ws://localhost:36658";
 
 #[tokio::test]
 async fn blob_submit_and_get_using_bridge_node() {
-    let client = new_test_client(AuthLevel::Write, Some(CELESTIA_BRIDGE_RPC_URL))
+    let client = new_test_client_with_url(AuthLevel::Write, CELESTIA_BRIDGE_RPC_URL)
         .await
         .unwrap();
     let namespace = random_ns();
--- a/rpc/tests/auth_token.rs
+++ b/rpc/tests/auth_token.rs
@@ -1,0 +1,63 @@
+#![cfg(not(target_arch = "wasm32"))]
+
+use celestia_rpc::prelude::*;
+use celestia_types::consts::appconsts::AppVersion;
+use celestia_types::Blob;
+
+pub mod utils;
+
+use crate::utils::client::{blob_submit, new_test_client, AuthLevel};
+use crate::utils::{random_bytes, random_ns};
+
+// Use node-1 (bridge node) as the RPC URL
+const CELESTIA_BRIDGE_RPC_URL: &str = "ws://localhost:36658";
+
+#[tokio::test]
+async fn blob_submit_and_get_using_bridge_node() {
+    let client = new_test_client(AuthLevel::Write, Some(CELESTIA_BRIDGE_RPC_URL))
+        .await
+        .unwrap();
+    let namespace = random_ns();
+    let data = random_bytes(5);
+    let blob = Blob::new(namespace, data, AppVersion::V2).unwrap();
+
+    let submitted_height = blob_submit(&client, &[blob.clone()]).await.unwrap();
+
+    let dah = client
+        .header_get_by_height(submitted_height)
+        .await
+        .unwrap()
+        .dah;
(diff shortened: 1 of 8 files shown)


YakupAltay:
Would love to have your inputs on the Swift tests, @zvolin & @fl0rek!

I could not find how my changes affected it.

rpc/tests/utils/client.rs
```diff
@@ -38,10 +38,16 @@ fn env_or(var_name: &str, or_value: &str) -> String {
     env::var(var_name).unwrap_or_else(|_| or_value.to_owned())
 }
 
-pub async fn new_test_client(auth_level: AuthLevel) -> Result<Client> {
+pub async fn new_test_client(
```
fl0rek, line 41: I don't love having to put `None` almost everywhere, since it's not obvious that it means 'use default'. Maybe we could have a pair `new_test_client` and `new_test_client_with_url` to parametrise the URL? That too could get messy with more params, but in this case it like it more. 
@zvolin @oblique thoughts?

rpc/tests/auth_token.rs
```diff
@@ -0,0 +1,63 @@
...
+    let leaves = blob.to_shares().unwrap();
+
+    proofs[0]
+        .verify_complete_namespace(&root_hash, &leaves, namespace.into())
+        .unwrap();
+}
+
+/// Blobs received from chain have index field set, so to
+/// compare if they are equal to the ones we sent, we need
+/// to overwrite the index field with received one.
```
fl0rek, line 57: nit: I think this should be normal comment

fl0rek (commented):
@YakupAltay thanks! Left few nits, otherwise looks good. 🎸 
Our swift CI is currently flaky, I'll look into that. Until then please restart the failed builds (or ping us? Not sure about permissions), they usually pass on the second attempt.

rpc/tests/utils/client.rs
YakupAltay: I have now both `new_test_client` and `new_test_client_with_url` passing the tests. But I have not committed them yet.

If you have another idea, I would love to implement it.

oblique: I like the `new_test_client`  `new_test_client_with_url`. Push that

YakupAltay: Pushed it! 🫡

rpc/tests/utils/client.rs
```diff
@@ -51,6 +51,22 @@ pub async fn new_test_client(auth_level: AuthLevel) -> Result<Client> {
     Ok(client)
 }
 
+pub async fn new_test_client_with_url(
```
fl0rek, line 54: `new_test_client` could be implemented in terms of `new_test_client_with_url` with `CELESTIA_RPC_URL` passed. I don't think it'd have any knock-off effects?

YakupAltay: Yes, I tested it and it worked.

You can check out my last commit.

rpc/tests/utils/client.rs
```diff
@@ -39,9 +39,16 @@ fn env_or(var_name: &str, or_value: &str) -> String {
 }
 
 pub async fn new_test_client(auth_level: AuthLevel) -> Result<Client> {
+    new_test_client_with_url(auth_level, CELESTIA_RPC_URL).await
```
oblique, line 42: I think we still need to do the `env_or` otherwise we will not be able to override url from env.

```suggestion
    let url = env_or("CELESTIA_RPC_URL", CELESTIA_RPC_URL);
    new_test_client_with_url(auth_level, &url).await
```

YakupAltay: Oh, sorry about that. I did not think of it.

You can view the change in the latest commit.

oblique: i will fix them and open a PR

oblique: ok so I opened #594 for the clippy issue. the test-wasm looks like a temporary issue,

rpc/tests/utils/client.rs
```diff
@@ -54,6 +57,11 @@ pub async fn new_test_client(auth_level: AuthLevel) -> Result<Client> {
     Ok(client)
 }
 
+pub async fn new_test_client(auth_level: AuthLevel) -> Result<Client> {
+    let url = env_or("CELESTIA_RPC_URL", CELESTIA_RPC_URL);
+    new_test_client_with_url(auth_level, &url).await
```
oblique, line 62: sorry i wasn't thinking straight when i said `env_or` is needed.
your previous code was correct.

```suggestion
    new_test_client_with_url(auth_level, CELESTIA_RPC_URL).await
```

YakupAltay: Oh, okay!

Pushed it now.

oblique (approved):
LGTM. Thanks!

rpc/tests/auth_token.rs
zvolin: either way works, I think it was also just copy pasted as I recall this code :p

YakupAltay: This is the way, haha

zvolin:
btw sorry for late review but last week we were all on a retreat and this week I'm bit sick, thanks for the patience :+1:

YakupAltay:
That's okay. I hope you're feeling better.

rpc/tests/auth_token.rs
```diff
@@ -0,0 +1,63 @@
...
+#[tokio::test]
+async fn blob_submit_and_get_using_bridge_node() {
+    let client = new_test_client_with_url(AuthLevel::Write, CELESTIA_BRIDGE_RPC_URL)
+        .await
+        .unwrap();
+    let namespace = random_ns();
+    let data = random_bytes(5);
+    let blob = Blob::new(namespace, data, AppVersion::V2).unwrap();
+
+    let submitted_height = blob_submit(&client, &[blob.clone()]).await.unwrap();
```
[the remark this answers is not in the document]
YakupAltay: I think we have a problem here.

`client.header_network_head()` at _/rpc/tests/client.rs_ gives authentication error for `AuthLevel::Skip`. So, we cannot directly use `new_test_client_with_url` with that auth level.

What do you suggest for it?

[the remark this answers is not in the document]
YakupAltay: Just committed a change for that, @zvolin. It's not clean enough but it's a way, haha.

oblique (approved):

zvolin (approved):

Issue #173, opened 2024-01-09
be less restrictive on dependencies versions in libraries


zvolin:
we should allow wider ranges of libraries where possible, perfectly testing the minimal possible versions that we can be used with, with some MSRV on CI

citizen-stig:
MSRV for version of celestia_types is 1.82. This is error on rust 1.81:

```
error[E0658]: use of unstable library feature 'iter_repeat_n'
   --> /Users/nikolai/.cargo/registry/src/index.crates.io-6f17d22bba15001f/celestia-types-0.11.2/src/row.rs:126:42
    |
126 |                 let mut shares: Vec<_> = iter::repeat_n(vec![], data_shares)
    |                                          ^^^^^^^^^^^^^^
    |
    = note: see issue #104434 <https://github.com/rust-lang/rust/issues/104434> for more information
```

PR #598, opened 2025-04-03
chore(rpc): update jsonrpsee version

Ferret-san:
update jsonrpsee version to v0.24.9 in order to fix https://github.com/paritytech/jsonrpsee/issues/1536

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1146,6 +1146,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.10.0"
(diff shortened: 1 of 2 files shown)


zvolin:
it looks like a no-op change for me, except the lock file update but it doesn't affect others as libraries lock is not propagated

zvolin:
actually I forgot that this can break others' code who want to depend on exact older jsonrpsee versions, while doing nothing useful for those who want latest. Ideally we would want to be pretty open with regards to versions so ideally we would just set `"0.24"` there and do so for all dependencies (#173)

PR #601, opened 2025-04-07
chore: Add contributing.md and badges

fl0rek:


CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
...
+This requires using one of the specifically supported versions of those crates.
+
+To support that we've created `./tools/upgrade-deps.sh` script which upgrades all
+dependencies in `Cargo.toml` except the ones that are patched by risc0.
+
+How to upgrade:
+```bash
+./tools/upgrade-deps.sh -i  # `-i` upgrades incompatible versions too
+cargo update
+```
```
oblique, line 35: By seeing this in contribution.md as a user I would feel that I have to update the depencies on my PRs. Maybe we need to change the title or in the text start saying "If you need to update dependencies ..."

fl0rek: Haven't thought of that, good idea. Added clarification to the title and additional sentence in the middle to drive the point.

README.md
```diff
@@ -1,5 +1,17 @@
 # Lumina
 
+<a href="https://github.com/eigerco/lumina/actions/workflows/ci.yml?query=branch%3Amain"><img src="https://github.com/eigerco/lumina/actions/workflows/ci.yml/badge.svg?branch=main" alt="Build Status" /></a>
```
zvolin, line 3: I'm wondering if they wouldn't look better if centered

fl0rek: image
vs
image
vs
image

I think I like the fully centered (last one) best.

zvolin: Agreed

CONTRIBUTING.md
oblique: I think you forgot to push it

fl0rek: Indeed, thanks for noticing!

CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
...
+see [github's guide](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) 
+for reference.
+
+## Conventional Commits
+
+We follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
+specification when writing commits messages. Consistency helps with readability
+of git history, additionally we use it to generate our changelog.
+
+## Upgrading dependencies (should you need to)
```
oblique, line 21: Isn't "If you need to" more correct?

fl0rek: afaik both are correct https://ludwig.guru/s/should+you+need

zvolin: I prefer "if you need to" as it's simpler to understand for not so expert english speakers like me :p

oblique: i also prefer "if you need to"

fl0rek: done

oblique (approved):

PR #606, opened 2025-04-10
feat(node)!: Implement a more dynamic backward syncing

oblique:


oblique:
We are block by this: https://github.com/cberner/redb/issues/970. When pruning window is less than sampling window and Pruner kicks in, we get the panic described in the issue.

node/src/pruner.rs
```diff
@@ -125,78 +149,258 @@ where
...
+            .is_some_and(|updated_at| updated_at.elapsed() < update_afer)
+        {
+            // No need to update cached values
+            return Ok(());
+        }
+
+        let after_sampling_window =
+            find_height_after_window(&*self.store, stored_blocks, sampling_cutoff).await?;
+        let after_pruning_window =
+            find_height_after_window(&*self.store, stored_blocks, pruning_cutoff).await?;
```
zvolin, line 245: so those do the binary search, which involves potentially a lot of store access. Could we instead cache the next header's time (n + 1 height) and check if it fallen outside of the window to avoid repeating the binary search if we have continuous range of headers? Then we could only update caches with a single query of height + 1 each time.

node-wasm/src/client.rs
```diff
@@ -32,35 +32,38 @@ use crate::wrapper::node::{PeerTrackerInfoSnapshot, SyncingInfoSnapshot};
...
-    ///
-    /// The minimum value that can be set is 60 seconds.
+    /// **Default value:** 2592000 seconds (30 days)\
+    /// **Minimum:** 60 seconds
     #[wasm_bindgen(js_name = customSamplingWindowSecs)]
     pub custom_sampling_window_secs: Option<u32>,
-    /// Pruning delay defines how much time the pruner should wait after sampling window in
-    /// order to prune the block.
+
+    /// Pruning window defines maximum age of a block considered to be kept in store.
```
fl0rek:
```suggestion
    /// Pruning window defines maximum age of a block for it to be retained in store.
```
og makes it sound like we consider block to be in store solely by virtue of its age, rather than the (I think) intended "this defines how long we keep stuff".

cli/src/native.rs
```diff
@@ -73,16 +72,13 @@ pub(crate) async fn run(args: Params) -> Result<()> {
...
 
-    if let Some(pruning_delay) = args.pruning_delay {
-        node_builder = node_builder.pruning_delay(pruning_delay);
+    if let Some(pruning_window) = args.pruning_window {
+        node_builder = node_builder.pruning_window(pruning_window);
     } else if args.in_memory_store {
-        // In-memory stores are memory hungry, so we lower pruning window.
-        node_builder = node_builder.pruning_delay(MIN_PRUNING_DELAY);
+        // In-memory stores are memory hungry, so we prune blocks as soon as possible.
+        node_builder = node_builder.pruning_window(MIN_PRUNING_WINDOW);
```
zvolin, line 81: can we default to 0?

node/src/block_ranges.rs
```diff
@@ -43,8 +45,8 @@ pub(crate) trait BlockRangeExt {
     fn is_adjacent(&self, other: &BlockRange) -> bool;
     fn is_overlapping(&self, other: &BlockRange) -> bool;
     fn left_of(&self, other: &BlockRange) -> bool;
-    fn truncate_left(&self, limit: u64) -> Self;
-    fn truncate_right(&self, limit: u64) -> Self;
+    fn keep_head(&self, limit: u64) -> Self;
+    fn keep_tail(&self, limit: u64) -> Self;
```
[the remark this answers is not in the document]
zvolin: "keep" for me a bit sounds like it was taking `&mut self`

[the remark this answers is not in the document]
oblique: good point. we already have `head` `tail` tho. which returns a single height

cli/src/native.rs
```diff
@@ -52,11 +52,10 @@ pub(crate) struct Params {
     #[clap(value_parser = parse_duration::parse)]
     pub(crate) sampling_window: Option<Duration>,
 
-    /// Pruning delay defines how much time the pruner should wait after sampling window in
-    /// order to prune the block.
+    /// Pruning window defines maximum age of a block considered to be keeped in store.
```
fl0rek:
```suggestion
    /// Pruning window defines maximum age of a block for it to be retained in store.
```
og makes it sound like we consider block to be in store solely by virtue of its age, rather than the (I think) intended "this defines how long we keep stuff".

node-uniffi/src/types/event.rs
```diff
@@ -135,9 +135,11 @@ pub enum NodeEvent {
         /// A human readable error.
         error: String,
     },
-    /// Pruned headers up to and including specified height.
+    /// Range of headers that were pruned.
     PrunedHeaders {
-        /// Last header height that was pruned
+        /// Start of the range.
+        from_height: u64,
+        /// End of the range (included).
```
fl0rek:
```suggestion
        /// End of the range (inclusive).
```

node-wasm/src/client.rs
```diff
@@ -32,35 +32,36 @@ use crate::wrapper::node::{PeerTrackerInfoSnapshot, SyncingInfoSnapshot};
...
     ///
-    /// If this is not set, then default value will apply:
+    /// If pruning window is smaller than sampling window then blocks will be pruned
+    /// exactly after they get sampled. This is useful when you want to keep low
+    /// memory footprint but still validate the blockchain.
```
fl0rek:
```suggestion
    /// If pruning window is smaller than sampling window, then blocks will be pruned
    /// right after they are sampled. This is useful when you want to keep low
    /// memory footprint but still validate the blockchain.
```

node/src/block_ranges.rs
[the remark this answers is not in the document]
fl0rek: In similar cases core adds `n` suffix: https://doc.rust-lang.org/std/primitive.slice.html#method.splitn 
maybe `head_n`/`headn`?

node/src/block_ranges.rs
```diff
@@ -444,6 +451,153 @@ impl BlockRanges {
...
+                .expect("BlockRanges always holds valid ranges");
+            edges
+                .insert_relaxed(end..=end)
+                .expect("BlockRanges always holds valid ranges");
+        }
+
+        edges
+    }
+
+    /// Returns `(left, middle, right)` that can be used for binary search.
```
fl0rek:
```suggestion
    /// Returns `([left], middle, [right])` that can be used for binary search.
```
suggestion to make the returned type more obvious

node/src/events.rs
```diff
@@ -203,16 +203,16 @@ pub enum NodeEvent {
         row: u16,
         /// The column of the share.
         column: u16,
-        /// The result of the sampling of the share.
-        accepted: bool,
+        /// Sampling of the share timed out.
```
fl0rek:
```suggestion
        /// Share sampling timed out.
```
nit: less wordy

node/src/events.rs
```diff
@@ -276,9 +276,11 @@ pub enum NodeEvent {
         error: String,
     },
 
-    /// Pruned headers up to and including specified height.
+    /// Range of headers that were pruned.
     PrunedHeaders {
-        /// Last header height that was pruned
+        /// Start of the range.
+        from_height: u64,
+        /// End of the range (included).
```
fl0rek:
```suggestion
        /// End of the range (inclusive).
```

node/src/node/builder.rs
```diff
@@ -231,17 +230,19 @@ where
         }
     }
 
-    /// Set pruning delay.
+    /// Set pruning window.
     ///
-    /// Pruning delay defines how much time the pruner should wait after sampling window in
-    /// order to prune the block.
+    /// Pruning window defines maximum age of a block considered to be kept in store.
```
fl0rek:
```suggestion
    /// Pruning window defines maximum age of a block for it to be retained in store.
```
og makes it sound like we consider block to be in store solely by virtue of its age, rather than the (I think) intended "this defines how long we keep stuff".

node/src/node/builder.rs
```diff
@@ -231,17 +230,19 @@ where
...
     ///
-    /// **Default if [`InMemoryStore`]/[`InMemoryBlockstore`] are used:** 60 seconds.\
-    /// **Default:** 1 hour.\
-    /// **Minimum:** 60 seconds.
-    pub fn pruning_delay(self, dur: Duration) -> Self {
+    /// If pruning window is smaller than sampling window then blocks will be pruned
+    /// exactly after they get sampled. This is useful when you want to keep low
+    /// memory footprint but still validate the blockchain.
```
fl0rek:
```suggestion
    /// If pruning window is smaller than sampling window, then blocks will be pruned
    /// right after they get sampled. This is useful when you want to keep low
    /// memory footprint but still validate the blockchain.
```

node/src/pruner.rs
```diff
@@ -44,16 +52,19 @@ where
...
+    pub daser: Arc<Daser>,
     /// Headers storage.
     pub store: Arc<S>,
     /// Block storage.
     pub blockstore: Arc<B>,
     /// Event publisher.
     pub event_pub: EventPublisher,
-    /// interval at which pruner will run
-    pub pruning_interval: Duration,
+    /// After how much time a new block is produced.
```
fl0rek:
```suggestion
    /// Average block production interval.
```

node/src/pruner.rs
```diff
@@ -106,12 +117,25 @@ where
...
+    cache: Cache,
+}
+
+struct Cache {
+    /// When Cache was updated.
+    updated_at: Option<Instant>,
+    /// The first known height that is after the pruning window.
+    after_pruning_window: Option<u64>,
+    /// The first known height that is after the sampling window.
+    after_sampling_window: Option<u64>,
```
fl0rek, line 140: If I understand correctly, maybe using `stored` instead of `known` would be more accurate here?

node/src/pruner.rs
```diff
@@ -125,78 +149,258 @@ where
...
         Ok(())
     }
 
-    /// Get oldest header from the store to be pruned or None if there's nothing to prune
-    async fn get_tail_header_to_prune(&self, cutoff: &Time) -> Result<Option<ExtendedHeader>> {
-        let Some(current_tail_height) = self.store.get_stored_header_ranges().await?.tail() else {
-            // empty store == nothing to prune
-            return Ok(None);
-        };
+    /// Returns a batch of at most 512 blocks that are needed to be pruned.
```
fl0rek:
```suggestion
    /// Returns a batch of at most 512 blocks that are need to be pruned.
```

node/src/pruner.rs
```diff
@@ -125,78 +149,258 @@ where
...
+            self.daser
+                .update_number_of_prunable_blocks(num_of_prunable_blocks)
+                .await?;
+            self.prev_num_of_prunable_blocks = num_of_prunable_blocks;
+        }
+
+        let mut prunable_batch = prunable_and_sampled.keep_head(MAX_PRUNABLE_BATCH_SIZE);
 
-        let header = self.store.get_by_height(current_tail_height).await?;
+        // Daser needs to allow us to remove anything beyond sampling and pruning window.
```
fl0rek, line 438: I'm assuming this means "we need to ask daser to allow us to remove anything beyond sampling and pruning window". In this case, why? wouldn't stuff beyond sampling window be of no concern for daser?

node/src/store.rs
```diff
@@ -266,14 +251,10 @@ impl SamplingMetadata {
 
 #[derive(Message)]
 struct RawSamplingMetadata {
-    #[prost(bool, tag = "1")]
-    accepted: bool,
-
+    // Tags 1 and 3 are reseerved because they were used in previous versions
```
fl0rek:
```suggestion
    // Tags 1 and 3 are reserved because they were used in previous versions
```

fl0rek (commented):
I'll need second through pass to understand the pruner, leaving this for now :P

node/src/pruner.rs
oblique: not really. let say we `after_pruning_window` is height 150 and it gets pruned, the next height that is after pruning window will be 149, but because if is smaller than 150 we do not update the value. so `after_pruning_window` holds a height that is known but not stored.

node/src/block_ranges.rs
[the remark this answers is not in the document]
oblique: i will go with `headn`/`tailn`

node-wasm/src/client.rs
```diff
@@ -32,35 +32,36 @@ use crate::wrapper::node::{PeerTrackerInfoSnapshot, SyncingInfoSnapshot};
...
     ///
-    /// * If `use_persistent_memory == true`, default value is 1 hour.
-    /// * If `use_persistent_memory == false`, default value is 60 seconds.
+    /// If this is not set, then default value will apply:
     ///
-    /// The minimum value that can be set is 60 seconds.
-    #[wasm_bindgen(js_name = customPruningDelaySecs)]
-    pub custom_pruning_delay_secs: Option<u32>,
+    /// * If `use_persistent_memory == true`, default value is 30 days plus 1 hour.
+    /// * If `use_persistent_memory == false`, default value is 0 seconds.
```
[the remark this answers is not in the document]
oblique: if we want to do this, we need to decouple bitswap from `Store`, otherwise users will not be able to fetch blobs

node/src/daser.rs
```diff
@@ -408,20 +550,16 @@ where
...
-    /// failed is via timeout.
-    async fn populate_queue(&mut self) -> Result<()> {
+    async fn update_queue(&mut self) -> Result<()> {
         let stored = self.store.get_stored_header_ranges().await?;
-        let accepted = self.store.get_accepted_sampling_ranges().await?;
+        let sampled = self.store.get_sampled_ranges().await?;
 
-        self.queue = stored - accepted - &self.done - &self.ongoing;
+        self.head_height = stored.head();
+        self.queue = stored - &sampled - &self.done - &self.ongoing - &self.will_be_pruned;
```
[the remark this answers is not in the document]
oblique: done

node/src/daser.rs
```diff
@@ -85,6 +78,29 @@ where
...
+        value: u64,
+    },
+
+    /// Used by Pruner to tell Daser about a block that is going to be pruned.
+    /// Daser then replies with `true` if Pruner can do it. This is needed to
+    /// avoid race conditions between them when Daser has as ongoing sampling.
+    WantToPrune {
+        height: u64,
+        respond_to: oneshot::Sender<bool>,
+    },
```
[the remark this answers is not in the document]
oblique: Updated the comment

node/src/pruner.rs
oblique: So basically this is needed for not having race conditions with the ongoing data sampling. I updated the `WantToPrune` doc comment in `daser.rs` that explains the race condition.

node/src/syncer.rs
```diff
@@ -535,6 +568,28 @@ where
...
+                break;
+            }
+
+            // If `header` is after the pruning edge, we mark it as the
+            // `highest_slow_sync_height` and we stop checking lower headers.
+            if header.time() <= pruning_cutoff {
+                self.highest_slow_sync_height = Some(header.height().value());
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
fl0rek: Doesn't sound too bad? maybe just leave a note about it

node/src/pruner.rs
oblique: Implemented an algorithm that uses some cached values. Check `find_height_after_window_fast`

node/src/store/redb_store.rs
```diff
@@ -38,8 +36,9 @@ const SCHEMA_VERSION_TABLE: TableDefinition<'static, (), u64> =
 const RANGES_TABLE: TableDefinition<'static, &str, Vec<(u64, u64)>> =
     TableDefinition::new("STORE.RANGES");
 
-const ACCEPTED_SAMPING_RANGES_KEY: &str = "KEY.ACCEPTED_SAMPING_RANGES";
+const SAMPLED_RANGES_KEY: &str = "KEY.ACCEPTED_SAMPING_RANGES";
```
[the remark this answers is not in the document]
oblique: renamed

node/src/store/indexed_db_store.rs
```diff
@@ -36,8 +34,9 @@ const SCHEMA_STORE_NAME: &str = "schema";
 const HASH_INDEX_NAME: &str = "hash";
 const HEIGHT_INDEX_NAME: &str = "height";
 
-const ACCEPTED_SAMPLING_RANGES_KEY: &str = "accepted_sampling_ranges";
 const HEADER_RANGES_KEY: &str = "header_ranges";
+const SAMPLED_RANGES_KEY: &str = "accepted_sampling_ranges";
```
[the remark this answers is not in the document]
oblique: renamed

fl0rek (approved):
so big 🙈

zvolin (approved):
nice work, thanks

PR #607, opened 2025-04-14
f Update update-proto-vendor.sh

petryshkaCODE:
change old link to a new one


tools/update-proto-vendor.sh
```diff
@@ -26,7 +26,7 @@ extract_urls ../target/proto-vendor-src \
     https://github.com/celestiaorg/celestia-app/archive/refs/tags/v3.0.2.tar.gz \
     https://github.com/celestiaorg/celestia-node/archive/refs/heads/main.tar.gz \
     https://github.com/celestiaorg/cosmos-sdk/archive/refs/heads/release/v0.46.x-celestia.tar.gz \
-    https://github.com/cometbft/cometbft/archive/refs/tags/v0.34.35.tar.gz \
+    https://codeload.github.com/cometbft/cometbft/tar.gz/refs/tags/v0.34.35 \
```
fl0rek, line 29: Could you explain how is this better? They link to the exact same file as far as I can tell. Also, why change only this one?

PR #610, opened 2025-04-14
feat(network): add Mammoth testnet support

YakupAltay:
Hey @zvolin & @fl0rek! 👋

Just saw the [Celestia team’s announcement](https://x.com/celestia/status/1911782025100771603) about their new public testnet Mammoth (`mamo-1`).
I’ve already added the network, but I’m a bit unsure about the next steps. How can I actually see it in action and get the light node running in my browser?

Would love your guidance on getting it fully live!

### DIFF
--- a/node-wasm/js/README.md
+++ b/node-wasm/js/README.md
@@ -3475,9 +3475,21 @@ lumina\_node\_wasm.d.ts:18
 
 ***
 
+#### Mammoth
+
+> **Mammoth**: `2`
+
+Mammoth testnet.
+
+##### Defined in
+
+TODO: Ask the Eiger team.
+
+***
+
 #### Mocha
 
-> **Mocha**: `2`
+> **Mocha**: `3`
 
 Mocha testnet.
 
@@ -3489,7 +3501,7 @@ lumina\_node\_wasm.d.ts:26
 
 #### Private
 
-> **Private**: `3`
+> **Private**: `4`
 
 Private local network.
 
--- a/node-wasm/src/utils.rs
+++ b/node-wasm/src/utils.rs
@@ -31,6 +31,8 @@ pub enum Network {
     Mainnet,
     /// Arabica testnet.
     Arabica,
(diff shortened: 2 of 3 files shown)


jcstein:
since this is an ephemeral testnet, I'd recommend not to include this

oblique:
We decided to follow @jcstein recommentation.
So this is what you can do:

We support custom networks if you use Rust crate directly:

```rust
    let node = Node::builder()
        .store(store)
        .blockstore(blockstore)
        .network(Network::custom("mamo-1").unwrap())
        .bootnodes(["/dnsaddr/da-bridge-0.par.mamochain.com/p2p/12D3KooWNc3hDtzLvyKj8xbcE3SFMRy4uX5EojCScCuqYRrz4tzS".parse().unwrap()])
        .start().await?;
```

Now, in case of JS, we need a bit of refactoring in order to do it, because there we currently represent the networks with an integer, instead of a string. What you can do for now, is to use your fork until we make this more flexible for JS.

@YakupAltay Just wondering, how do you use our API? Are you using our npm package in JS, or are you doing something else?

YakupAltay:
Thanks for the clarification and the example code for adding custom networks via the Rust crate.

My initial goal was simply to view the Mammoth (`mamo-1`) testnet directly in the front end. Since it's possible to add it manually as a custom network, that's sufficient for now.

Regarding the API, I’m not currently using it, but I appreciate your guidance on the JS side and will keep that in mind if I explore integration further.

If there are no objections, I’ll go ahead and close this PR.

Thanks again!

oblique:
In case you want to use lumina-cli, you can also do:

```
cargo run --release -p lumina-cli -- node -n mamo-1 -b /dnsaddr/da-bridge-0.par.mamochain.com/p2p/12D3KooWNc3hDtzLvyKj8xbcE3SFMRy4uX5EojCScCuqYRrz4tzS
```

fl0rek:
per above, closing

PR #648, opened 2025-06-05
fix(node)!: Do not fail on dns-over-https

### LINKED ISSUE #632: bug: lumina NodeBuilder fails to start if one of the DNS of the Bootstrappers cannot be resolved
```
NodeBuilder: Could not resolve bootnode addresses: Could not parse response: Failed deserializing DoH response: Error: missing field `Answer`
```

Instead of failing at the start of the node, if only one of the DNS fails to resolve, the node should try all Bootnodes and use the one where the IP is available and the node can connect to.

oblique:
Fixes #632

### DIFF
--- a/node/src/utils/dns.rs
+++ b/node/src/utils/dns.rs
@@ -5,7 +5,7 @@ use js_sys::Promise;
 use libp2p::multiaddr::Protocol;
 use libp2p::{Multiaddr, PeerId};
 use serde::Deserialize;
-use tracing::error;
+use tracing::warn;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen_futures::JsFuture;
 use web_sys::{Request, RequestInit, RequestMode, Response};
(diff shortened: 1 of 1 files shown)


node/src/utils/dns.rs
```diff
@@ -20,17 +21,17 @@ pub(crate) enum Error {
...
-) -> Result<Vec<Multiaddr>, Error> {
+pub(crate) async fn resolve_bootnode_addresses(addrs: Vec<Multiaddr>) -> Vec<Multiaddr> {
     let mut bootnodes = Vec::with_capacity(addrs.len());
 
     for addr in addrs {
-        let resolved_addrs = resolve_dnsaddr_multiaddress(addr, DEFAULT_DNS_ADDR).await?;
-        bootnodes.extend(resolved_addrs.into_iter());
+        match resolve_dnsaddr_multiaddress(&addr, DEFAULT_DNS_ADDR).await {
+            Ok(resolved_addrs) => bootnodes.extend(resolved_addrs.into_iter()),
+            Err(e) => error!("Failed to resolve {addr}: {e}"),
```
zvolin: let's make it a warning

zvolin (approved):

node/src/utils/dns.rs
oblique: done

PR #653, opened 2025-06-11
fix(node)!: Make blockstore only store samples

zvolin:


node/src/node.rs
```diff
@@ -364,8 +364,12 @@ where
         Ok(self.p2p().get_row(row_index, block_height, timeout).await?)
     }
 
+    // TODO: Fix the storage cleanup after calling this method.
```
zvolin: I didn't try to resolve this to not conflict with #606, but happy to clean that up if we merge 606 first

node/src/blockstore.rs
```diff
@@ -1,20 +1,163 @@
...
 
 #[cfg(target_arch = "wasm32")]
 /// An [`IndexedDbBlockstore`].
 ///
 /// [`IndexedDbBlockstore`]: blockstore::IndexedDbBlockstore
-pub type IndexedDbBlockstore = blockstore::IndexedDbBlockstore;
+pub type IndexedDbBlockstore = SampleBlockstore<blockstore::IndexedDbBlockstore>;
+
+/// A blockstore which only stores samples and discards other CIDs.
+pub struct SampleBlockstore<B> {
```
oblique: I think we should keep this struct private and wrap blockstore later in `p2p.rs`

zvolin: do you mean to wrap it only for bitswap but not on `Node<B, S>` and in `Pruner` keep original blockstore?

oblique: I mean, in the `Node::start` we can wrap `config.blockstore` before we pass it to other components. Something like `Arc::new(SampleBlockstore::new(config.blockstore))`.

zvolin: sorry, had some brain fog, fixed

fl0rek (approved):
🎸

PR #654, opened 2025-06-13
feat: Wasm grpc client

fl0rek:


types/src/any.rs
```diff
@@ -0,0 +1,99 @@
+//! Helpers for dealing with `Any` types conversion
+
+use prost::Name;
+use tendermint_proto::google::protobuf::Any;
+
+/// Value convertion into protobuf's Any
+pub trait IntoAny {
```
oblique: Should we rename this to `IntoProtobufAny`?

types/src/block.rs
```diff
@@ -126,624 +149,160 @@ impl From<Block> for RawBlock {
...
-        //
-        // NOTE: We do not validate `app_hash` because if can be anything
-
-        Ok(())
+#[cfg(all(target_arch = "wasm32", feature = "wasm-bindgen"))]
+#[wasm_bindgen]
+impl Block {
+    /// Block header
+    #[wasm_bindgen(getter)]
+    pub fn get_header(&self) -> JsHeader {
```
oblique: If I'm not mistaken, wasm-bindgen suggests getters to be without `get_` prefix

utils/src/object.rs
```diff
@@ -0,0 +1,17 @@
+#[macro_export]
+/// Create a new javascript `Object` with given properties
```
oblique:
```suggestion
/// Create a new javascript `Object` with given properties
#[macro_export]
```

utils/src/lib.rs
```diff
@@ -12,3 +15,5 @@ pub mod time;
 /// platform independent cancellation token
 #[cfg(feature = "token")]
 pub mod token;
+//#[cfg(all(target_arch = "wasm32", feature = "make-object"))]
+//pub use crate::object::make_object;
```
oblique:
```suggestion
```

types/src/signature.rs
```diff
@@ -0,0 +1,69 @@
...
+#[cfg(all(target_arch = "wasm32", feature = "wasm-bindgen"))]
+mod wbg {
+    use js_sys::Uint8Array;
+    use tendermint::signature::Signature;
+    use wasm_bindgen::prelude::*;
+
+    /// Signature
+    #[derive(Clone, Debug)]
+    #[wasm_bindgen(getter_with_clone, js_name = "Signature")]
+    pub struct JsSignature(pub Uint8Array);
```
oblique, line 17: is `pub` needed for `Uint8Array`?

types/src/state/tx.rs
```diff
@@ -168,15 +254,72 @@ pub struct SignerInfo {
...
+#[wasm_bindgen]
+pub enum SignatureMode {
+    Single,
+    Multi,
+}
+
 /// sum is the oneof that specifies whether this represents a single or nested
 /// multisig signer
 #[derive(Debug, Clone, PartialEq)]
+#[non_exhaustive]
```
oblique: Why non-exhaustive was added?

types/src/wasm_types.rs
oblique, line 1: Remove this empty file

types/src/signature.rs
fl0rek: This is mostly meant as a type returned to JS, so I've left it public so that JS side can inspect the contents.

types/src/block.rs
fl0rek: get_ was added as not to clash with rust name, I'll add rename on js side

types/src/state/tx.rs
fl0rek: I swore I saw this in tendermint, so I've added it thinking we missed previously, but now I cannot find it. Removing.

types/src/any.rs
fl0rek: No strong feelings here, so I've changed it. One drawback is that it's public and in our bindings so it's breaking change for everybody.

fl0rek: also, @zvolin thoughts?

types/src/any.rs
```diff
@@ -0,0 +1,99 @@
...
+pub trait IntoProtobufAny {
+    /// Converts itself into protobuf's Any type
+    fn into_any(self) -> Any;
+}
+
+impl<T> IntoProtobufAny for T
+where
+    T: Name,
+{
+    fn into_any(self) -> Any {
```
oblique, line 16: Should we change also the method?

types/src/any.rs
fl0rek: Maciek mentioned during daily that he's ok with the change 🆗

types/src/any.rs
fl0rek: I felt that in context it's obvious what kind of `any` we're referering to here, so I opted to save few keystrokes. If you feel strongly we should change it, I can though.

types/src/lib.rs
```diff
@@ -1,13 +1,15 @@
...
 
+pub mod any;
 pub mod blob;
 pub mod block;
 mod byzantine;
 pub mod consts;
 mod data_availability_header;
 pub mod eds;
 mod error;
+pub mod evidence;
```
zvolin, line 16: it looks like it is going to be empty mod for non uniffi / non wasm cases

grpc/src/grpc.rs
```diff
@@ -15,6 +15,7 @@ use celestia_proto::cosmos::tx::v1beta1::service_client::ServiceClient as TxServ
 use celestia_types::blob::BlobParams;
 use celestia_types::block::Block;
 use celestia_types::hash::Hash;
+use celestia_types::state::auth::Account;
 use celestia_types::state::auth::AuthParams;
```
zvolin:
```suggestion
use celestia_types::state::auth::{Account, AuthParams};
```

grpc/src/js_client/grpc_client.rs
```diff
@@ -0,0 +1,137 @@
...
+            .client
+            .get_accounts()
+            .await?
+            .into_iter()
+            .map(Into::into)
+            .collect())
+    }
+
+    /// Get balance of coins with given denom
+    pub async fn get_balance(&self, address: &str, denom: String) -> Result<JsCoin> {
```
zvolin: I don't think that matters much but for consistency
```suggestion
    pub async fn get_balance(&self, address: &str, denom: &str) -> Result<JsCoin> {
```

grpc/src/js_client/grpc_client.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Get block by height
+    pub async fn get_block_by_height(&self, height: i64) -> Result<Block> {
+        self.client.get_block_by_height(height).await
+    }
+
+    /// Broadcast prepared and serialised transaction
+    pub async fn broadcast_tx(
+        &self,
+        tx_bytes: Vec<u8>,
+        mode: JsBroadcastMode,
```
zvolin: I'm a bit afraid this could nullify the value on javascript side if not passed by ref, since we defined it as wasm managed object
```suggestion
        mode: &JsBroadcastMode,
```

grpc/src/js_client/grpc_client.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Celestia GRPC client
+#[wasm_bindgen]
+pub struct GrpcClient {
+    client: InnerClient,
+}
+
+#[wasm_bindgen]
+impl GrpcClient {
+    /// Create a new client connected with the given `url`
+    pub async fn new(url: String) -> Result<Self> {
```
zvolin: again just for consistency, as we mostly use `&str` now
```suggestion
    pub async fn new(url: &str) -> Result<Self> {
```

grpc/src/js_client/tx_client.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// ```js
+    /// await window.leap.enable("mocha-4")
+    /// const keys = await window.leap.getKey("mocha-4")
+    ///
+    /// const signer = (signDoc) => {
+    ///   return window.leap.signDirect("mocha-4", keys.bech32Address, signDoc, { preferNoSetFee: true })
+    ///     .then(sig => Uint8Array.from(atob(sig.signature.signature), c => c.charCodeAt(0)))
+    /// }
+    ///
+    /// const tx_client = await new TxClient("http://127.0.0.1:18080", keys.bech32Address, keys.pubKey, signer)
```
zvolin: Examples should be aligned to not use addresses anymore

grpc/src/js_client/tx_client.rs
```diff
@@ -0,0 +1,332 @@
...
+use crate::{Result, TxClient};
+
+/// Celestia grpc transaction client.
+#[wasm_bindgen(js_name = "TxClient")]
+pub struct JsTxClient {
+    client: TxClient<Client, JsSigner>,
+}
+
+#[wasm_bindgen(js_class = "TxClient")]
+impl JsTxClient {
```
zvolin, line 27: I'm afraid it cannot be done without writing the actual boilerplate, but I think tx client should be a superset of grpc client also in wasm, so it should now get also all the methods that are present in the js version of grpc client

zvolin: we could have `common_client_impl!` macro to avoid boilerplate tho

zvolin: also I noticed the grpc client uses mostly `&str`s and here we use mostly `String`s, the common client impl would make that consistent too

types/src/block/commit.rs
```diff
@@ -0,0 +1,509 @@
...
+        BlockIdFlagNil,
+    }
+
+    #[derive(Clone, Debug)]
+    #[wasm_bindgen(getter_with_clone, js_name = "CommitVote")]
+    pub struct JsCommitVote {
+        pub validator_address: String,
+        pub timestamp: String,
+        pub signature: Option<JsSignature>,
+    }
```
zvolin: missing docs

types/src/lib.rs
zvolin: it looks like the same case is going to be with `signature` module

types/src/signature.rs
```diff
@@ -0,0 +1,69 @@
...
+pub mod uniffi_types {
+    use tendermint::signature::Signature as TendermintSignature;
+    use uniffi::Record;
+
+    use crate::UniffiConversionError;
+
+    /// Signature
+    #[derive(Record)]
+    pub struct Signature {
+        signature: Vec<u8>,
```
zvolin: shouldn't this be pub similarly to the wasm one? or `Record` enforces vis?

types/src/state/bit_array.rs
```diff
@@ -46,6 +46,30 @@ impl From<BitVector> for CompactBitArray {
...
+pub use wbg::*;
+
+#[cfg(all(target_arch = "wasm32", feature = "wasm-bindgen"))]
+mod wbg {
+    use super::BitVector;
+    use wasm_bindgen::prelude::*;
+
+    #[wasm_bindgen(getter_with_clone)]
+    pub struct JsBitVector(pub Vec<u8>);
+
```
zvolin, line 61: missing docs

zvolin: same in uniffi version :p

types/src/state/tx.rs
```diff
@@ -714,6 +962,82 @@ impl TryFrom<RawCoin> for Coin {
...
+    use wasm_bindgen::prelude::*;
+
+    use lumina_utils::make_object;
+
+    #[derive(Clone)]
+    #[wasm_bindgen(getter_with_clone)]
+    pub struct JsEvent {
+        pub r#type: String,
+        pub attributes: Vec<JsEventAttribute>,
+    }
```
zvolin, line 985: missing docs

types/src/state/tx.rs
```diff
@@ -714,6 +962,82 @@ impl TryFrom<RawCoin> for Coin {
...
+        }
+    }
+
+    #[derive(Clone)]
+    #[wasm_bindgen(getter_with_clone)]
+    pub struct JsEventAttribute {
+        pub key: Vec<u8>,
+        pub value: Vec<u8>,
+        pub index: bool,
+    }
```
zvolin, line 1002: missing docs

zvolin (commented):
nice job, doing all the binds is the 13th Hercules's labor :p left few nits / ideas

grpc/src/js_client/grpc_client.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Get block by height
+    pub async fn get_block_by_height(&self, height: i64) -> Result<Block> {
+        self.client.get_block_by_height(height).await
+    }
+
+    /// Broadcast prepared and serialised transaction
+    pub async fn broadcast_tx(
+        &self,
+        tx_bytes: Vec<u8>,
+        mode: JsBroadcastMode,
```
zvolin: this should still be passed by ref
```suggestion
        mode: &JsBroadcastMode,
```

zvolin (approved):
LGTM

grpc/src/js_client/tx_client.rs
zvolin: we agreed that this will be done in follow up for both uniffi and js

oblique (approved):

PR #655, opened 2025-06-16
feat(grpc)!: Expose entire GrpcClient API plus required type changes 

fl0rek:
Changelist:
- use addresses serialised as bech32 strings in uniffi, both for input and output
  - remote Address uniff type
- enable all grpc calls by making grpc client methods pub
- add grpc tests in swift
  - they require local-devnet setup which is not possible for mac, so they are _disabled_


grpc/src/uniffi_client/grpc_client.rs
```diff
@@ -59,8 +59,8 @@ impl GrpcClient {
     }
 
     /// Get account
-    pub async fn get_account(&self, account: &Address) -> Result<Account> {
-        Ok(self.client.get_account(account).await?)
+    pub async fn get_account(&self, account: &str) -> Result<Account> {
```
oblique, line 64: I'm a bit confused. Below you defined `AddressObject`. Aren't we going to use that?

fl0rek: I'm a bit torn here. str makes it easy and in line with wasm interface. But we also need `AddressObject` as a way to interface with `Address` in returned types. 
If you have strong feelings either way I can go with your suggestion.

.github/workflows/ci.yml
```diff
@@ -201,9 +201,9 @@ jobs:
     - name: xcode clean build
       working-directory: examples/ios/LuminaDemo
       run: xcodebuild -scheme LuminaDemo clean build CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO
-    - name: xcode sim test
-      working-directory: examples/ios/LuminaDemo
-      run: xcodebuild -scheme LuminaDemo -sdk iphonesimulator -destination "platform=iOS Simulator,name=iPhone 16" test
+#    - name: xcode sim test
+#      working-directory: examples/ios/LuminaDemo
+#      run: xcodebuild -scheme LuminaDemo -sdk iphonesimulator -destination "platform=iOS Simulator,name=iPhone 16" test
```
zvolin, line 206: do we want this to stay disabled?

types/src/state/address.rs
```diff
@@ -311,12 +311,37 @@ fn string_to_kind_and_id(s: &str) -> Result<(AddressKind, Id)> {
...
 pub(crate) mod uniffi_types {
     use super::{AccAddress, Address as RustAddress, ConsAddress, Id, ValAddress};
-    use uniffi::{Enum, Record};
+    use uniffi::{Enum, Object, Record};
 
-    use crate::error::UniffiConversionError;
+    use crate::{error::UniffiConversionError, UniffiError};
+
+    #[derive(Object, Clone)]
+    struct AddressObject(Address);
```
zvolin: I don't really understand, why is this needed?

zvolin (commented):
could you please modify title (and add some description) what this PR is for? I only see tests, some changes wrt to address passing and some pubs added, not really the tx broadcast changes

.github/workflows/ci.yml
fl0rek: Yes. Until there's a way to run docker builder on macos github action, we cannot setup local-devnet, so no grpc tests for now.

grpc/src/uniffi_client/grpc_client.rs
oblique: `str` sounds good for users

types/src/state/address.rs
fl0rek: Switched to using addresses as bech32 strings everywhere and removed this type

zvolin (approved):

oblique (approved):

PR #658, opened 2025-06-17
feat: make Merkle proof fields pub for Blobstream validation

odesenfans:
In the context of implementing DA bridging with Blobstream, it is necessary to access specific fields of the `MerkleProof` and `RowProof` structs returned by the RPC client.

This PR makes the fields we found to be necessary public.

### DIFF
--- a/types/src/data_availability_header.rs
+++ b/types/src/data_availability_header.rs
@@ -346,7 +346,7 @@ impl ValidateBasicWithAppVersion for DataAvailabilityHeader {
 #[serde(try_from = "RawRowProof", into = "RawRowProof")]
 pub struct RowProof {
     row_roots: Vec<NamespacedHash>,
-    proofs: Vec<MerkleProof>,
+    pub proofs: Vec<MerkleProof>,
     start_row: u16,
     end_row: u16,
 }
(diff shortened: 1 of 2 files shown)


types/src/data_availability_header.rs
```diff
@@ -346,7 +346,8 @@ impl ValidateBasicWithAppVersion for DataAvailabilityHeader {
 #[serde(try_from = "RawRowProof", into = "RawRowProof")]
 pub struct RowProof {
     row_roots: Vec<NamespacedHash>,
-    proofs: Vec<MerkleProof>,
+    /// Proofs of inclusion of each row root in the data availability header.
+    pub proofs: Vec<MerkleProof>,
```
zvolin: could we make it a getter instead, similar to the `row_roots`?

odesenfans: Of course!

odesenfans: Fixed in f6e9120bdd0e7a95cc1d4f76067fad6dce129a31.

zvolin (approved):

zvolin:
thanks

Issue #666, opened 2025-06-24
Add a TxClient constructor with the signing key bytes


zvolin:

fl0rek:
closed in #673

PR #673, opened 2025-06-27
feat(grpc): Streamline TxClient creation API, add docs with example

fl0rek:
related #666

### DIFF
--- a/grpc/src/tx.rs
+++ b/grpc/src/tx.rs
@@ -135,7 +135,7 @@ where
     /// let address = Address::from_account_veryfing_key(*signing_key.verifying_key());
     /// let grpc_url = "public-celestia-mocha4-consensus.numia.xyz:9090";
     ///
-    /// let tx_client = TxClient::with_url_and_signer_keypair(grpc_url, signing_key)
+    /// let tx_client = TxClient::with_url_and_keypair(grpc_url, signing_key)
     ///     .await
     ///     .unwrap();
     ///
@@ -199,7 +199,7 @@ where
     /// let address = Address::from_account_veryfing_key(*signing_key.verifying_key());
     /// let grpc_url = "public-celestia-mocha4-consensus.numia.xyz:9090";
     ///
-    /// let tx_client = TxClient::with_url_and_signer_keypair(grpc_url, signing_key)
+    /// let tx_client = TxClient::with_url_and_keypair(grpc_url, signing_key)
     ///     .await
     ///     .unwrap();
     ///
@@ -443,7 +443,7 @@ where
     /// # const GRPC_URL : &str = "http://localhost:19090";
     /// const HEX_SIGNING_KEY: &str = "393fdb5def075819de55756b45c9e2c8531a8c78dd6eede483d3440e9457d839";
     /// let signing_key = SigningKey::from_slice(&hex::decode(HEX_SIGNING_KEY).unwrap()).unwrap();
-    /// let client = TxClient::with_url_and_signer_keypair(GRPC_URL, signing_key).await.unwrap();
+    /// let client = TxClient::with_url_and_keypair(GRPC_URL, signing_key).await.unwrap();
     /// # }
     /// ```
     pub async fn with_url_and_keypair(url: impl Into<String>, signer_keypair: S) -> Result<Self> {
--- a/grpc/src/tx.rs
+++ b/grpc/src/tx.rs
@@ -88,24 +88,6 @@ where
     S: DocSigner,
 {
     /// Create a new transaction client with the specified account.
-    ///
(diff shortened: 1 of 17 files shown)


grpc/src/tx.rs
```diff
@@ -85,15 +87,11 @@ where
...
-        account_address: &Address,
-        account_pubkey: VerifyingKey,
-        signer: S,
-    ) -> Result<Self> {
+    /// Create a new transaction client with the specified account.
+    pub async fn new(transport: T, account_pubkey: VerifyingKey, signer: S) -> Result<Self> {
         let client = GrpcClient::new(transport);
-        let account = client.get_account(account_address).await?;
+        let account_address = Address::AccAddress(AccAddress::from(account_pubkey));
+        let account = client.get_account(&account_address).await?;
```
zvolin: I think I accepted the address here explicitly rather than deducing it from pubkey, because there are many address types, and I thought that each of them could be used there. But since we call `get_account`, I guess it's cool to only have account addresses here :facepalm:

grpc/src/tx.rs
```diff
@@ -420,12 +416,42 @@ where
...
+    /// # use celestia_types::state::Address;
+    /// # use k256::ecdsa::SigningKey;
+    /// # use celestia_grpc::TxClient;
+    /// # const GRPC_URL : &str = "http://localhost:19090";
+    /// const HEX_SIGNING_KEY: &str = "393fdb5def075819de55756b45c9e2c8531a8c78dd6eede483d3440e9457d839";
+    /// let signing_key = SigningKey::from_slice(&hex::decode(HEX_SIGNING_KEY).unwrap()).unwrap();
+    /// let client = TxClient::with_url_and_signer_keypair(GRPC_URL, signing_key).await.unwrap();
+    /// # }
+    /// ```
+    pub async fn with_url_and_signer_keypair(
```
oblique: This name feels too explicit. I think I prefer something like `with_url_and_signer`.

grpc/src/tx.rs
```diff
@@ -130,15 +128,14 @@ where
...
     /// use celestia_grpc::{TxClient, TxConfig};
     /// use celestia_proto::cosmos::bank::v1beta1::MsgSend;
-    /// use celestia_types::state::{AccAddress, Coin};
+    /// use celestia_types::state::{Address, AccAddress, Coin};
     /// use tendermint::crypto::default::ecdsa_secp256k1::SigningKey;
     ///
     /// let signing_key = SigningKey::random(&mut rand_core::OsRng);
-    /// let public_key = *signing_key.verifying_key();
-    /// let address = AccAddress::new(public_key.into()).into();
+    /// let address : Address = AccAddress::new((*signing_key.verifying_key()).into()).into();
```
oblique: This doesn't look convinient. I suggest these explicit methods:

* `AccAddress::from_verifying_key`
* `ValAddress::from_verifying_key`
* `ConsAddress::from_verifying_key`
* `Address::from_account_verifying_key`
* `Address::from_validator_verifying_key`
* `Address::from_consensus_verifying_key`


Also, should we reexport `VerifyingKey` from `k256`?

grpc/src/tx.rs
fl0rek: It's touch clunky, agree. Problem is _both_ `with_url*` constructors take `Signer` and the difference is whether this signer implements Keypair, or do we need public key passed separately. 
I've shortened it to `with_url_and_keypair`, wdyt?

grpc/src/tx.rs
fl0rek: Added the `from`s for `Address` and reexported `VerifyingKey` 👌 
I didn't add `from_verifyng_key` for `{AccAddress, ValAddress, ConsAddress}` since there's `From<VerifyingKey>` for those, which feels more idiomatic.

grpc/src/tx.rs
```diff
@@ -85,15 +87,11 @@ where
...
-        transport: T,
-        account_address: &Address,
-        account_pubkey: VerifyingKey,
-        signer: S,
-    ) -> Result<Self> {
+    /// Create a new transaction client with the specified account.
+    pub async fn new(transport: T, account_pubkey: VerifyingKey, signer: S) -> Result<Self> {
         let client = GrpcClient::new(transport);
-        let account = client.get_account(account_address).await?;
+        let account_address = Address::from_account_veryfing_key(account_pubkey);
```
zvolin: sorry for the late notice, but I think we should just use `AccAddress` everywhere directly

grpc/src/tx.rs
oblique: sure

oblique (approved):

PR #677, opened 2025-07-04
feat(grpc)!: Trustless balance queries

zvolin:


grpc/src/error.rs
```diff
@@ -65,6 +67,15 @@ pub enum Error {
     #[error("Provided public key differs from one associated with account")]
     PublicKeyMismatch,
 
+    /// ABCI proof verification has failed
+    #[error("ABCI proof verification has failed: {0}")]
+    AbciProof(#[from] ProofError),
+
+    // TODO: error code after wrapping of AbciQueryResponse
```
zvolin: let's wait with this for #654 to follow the same conventions it introduces

grpc/src/grpc.rs
```diff
@@ -90,6 +94,65 @@ where
...
 
+    /// Get balance of coins with [`BOND_DENOM`] for the given address, together with a proof,
+    /// and verify the returned balance against the corresponding block's [`AppHash`].
+    ///
+    /// NOTE: the balance returned is the balance reported by the parent block of
+    /// the provided header. This is due to the fact that for block N, the block's
+    /// [`AppHash`] is the result of applying the previous block's transaction list.
+    ///
+    /// [`AppHash`]: tendermint::hash::AppHash
+    pub async fn get_verified_balance(
```
zvolin, line 104: should this be renamed to `balance_for_account` to match celestia grpc client name?

oblique: yes, but `balance_for_address`

grpc/src/abci_proofs.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    /// ABCI proof is missing
+    #[error("ABCI proof is missing")]
+    AbciProofMissing,
+
+    /// Unsupported proof spec
+    #[error("Unsupported proof spec: {0}")]
+    UnsupportedSpec(String),
+
+    ///Decoding from proto failed
```
oblique:
```suggestion
    /// Decoding from proto failed
```

grpc/tests/utils/mod.rs
```diff
@@ -99,18 +105,24 @@ mod imp {
     use std::future::Future;
 
     use celestia_grpc::{GrpcClient, TxClient};
+    use celestia_rpc::Client as JrpcClient;
```
oblique: This is very minor, but I prefer `RpcClient`. The same goes for all the helpers and variable names.
Up to you tho

grpc/src/grpc.rs
oblique: Actually we can do this only in `Client`.

grpc/src/grpc.rs
```diff
@@ -87,6 +93,63 @@ where
...
+    /// NOTE: the balance returned is the balance reported by the parent block of
+    /// the provided header. This is due to the fact that for block N, the block's
+    /// [`AppHash`] is the result of applying the previous block's transaction list.
+    ///
+    /// [`AppHash`]: ::tendermint::hash::AppHash
+    pub async fn get_verified_balance(
+        &self,
+        address: &Address,
+        header: &ExtendedHeader,
+    ) -> Result<Coin> {
```
oblique, line 108: Since this is only for bond denom coin, we can return just the amount.

grpc/src/abci_proofs.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub fn verify_membership(
+        &self,
+        root: impl AsRef<[u8]>,
+        keys: impl IntoIterator<Item = impl AsRef<[u8]>>,
+        leaf: impl AsRef<[u8]>,
+    ) -> Result<(), ProofError> {
+        let root = root.as_ref();
+        let mut current_leaf = leaf.as_ref();
+
+        for (i, key) in keys.into_iter().enumerate() {
```
zvolin: This need to check if length of proofs and keys is equal, otherwise allows for proof crafting

grpc/src/grpc.rs
zvolin: we agreed to for now keep `Coin` there and use `Amount` directly only in `celestia-client`

oblique: after discussing this, we will keep it as `Coin`

oblique (approved):

fl0rek (approved):
nice!

PR #680, opened 2025-07-08
feat(grpc)!: Add support for Gas Estimation Service

fl0rek:
Few things here:
- update our proto vendor script to gather all the versions at the top
- update proto to pull gas estimation service defintions
- changes to grpc proc macro to support multiline comments
- Add Gas Estimation Service to GrpcClient
- Replace local gas estimation with querying Gas Estimation Service
- Update comments for GrpcClient to match ones in proto definitions


grpc/src/tx.rs
```diff
@@ -244,15 +238,12 @@ where
...
+        self.client.get_min_gas_price().await
     }
 
-    /// Set current gas price used by the client
-    async fn update_gas_price(&self) -> Result<f64> {
-        let gas_price = self.client.get_min_gas_price().await?;
-        *self.gas_price.write().expect("lock poisoned") = gas_price;
-        Ok(gas_price)
+    pub async fn get_estimate_gas_price(&self, priority: TxPriority) -> Result<f64> {
+        self.client.get_estimate_gas_price(priority).await
```
fl0rek: Since we're moving to a place where TxClient does not do any gas estimation, I've removed the min gas price caching from it. This means that `last_seen_gas_price` no longer makes sense, but I thought I'd add these two as a substitute. Alternative would be expecting user to create a `GrpcClient` and query the appropriate functions there directly.
wdyt? @zvolin @oblique

grpc/src/grpc.rs
```diff
@@ -128,25 +133,50 @@ where
...
+
+    // celestia.core.gas_estimation
+
+    /// get_estimate_gas_price takes a transaction priority and estimates the gas price based
+    /// on the gas prices of the transactions in the last five blocks.
+    ///
+    /// If no transaction is found in the last five blocks, return the network
+    /// min gas price.
+    #[grpc_method(GasEstimatorClient::estimate_gas_price)]
+    async fn get_estimate_gas_price(&self, priority: TxPriority) -> Result<f64>;
```
zvolin: I think I would remove the `get_` prefix here and below, sounds cleaner without it

grpc/src/grpc.rs
```diff
@@ -76,33 +80,34 @@ where
 
     // cosmos.auth
 
-    /// Get auth params
+    /// Params queries all parameters.
```
zvolin: I think for most of the docs, I liked ours more, although I'm okay with the change

grpc/src/tx.rs
```diff
@@ -652,22 +708,12 @@ pub async fn sign_tx(
     };
     let signature = signer.try_sign(doc).await?;
 
+    //println!("{:?}", signature.to_bytes().to_vec().len());
+    //panic!("AAAA");
+
```
zvolin:
```suggestion
```

grpc/src/tx.rs
```diff
@@ -553,17 +561,56 @@ mod wbg {
...
+      /**
+       * Estimated gas price is the mean of all gas prices from the last 5 blocks.
+       */
+      Medium = 2,
+      /**
+       * Estimated gas price is the price at the start of the top 10% of transactions’ gas prices from the last 5 blocks.
+       */
+      High = 3,
+    }
+
```
zvolin: I think this conflicts with the `wasm_bindgen` derive above. We should choose one or the other, but not both at the same time

grpc/src/js_client.rs
```diff
@@ -71,10 +71,20 @@ impl JsClient {
...
+        self.client.get_min_gas_price().await
+    }
+
+    /// get_estimate_gas_price takes a transaction priority and estimates the gas price based
+    /// on the gas prices of the transactions in the last five blocks.
+    ///
+    /// If no transaction is found in the last five blocks, return the network
+    /// min gas price.
+    #[wasm_bindgen(js_name = getEstimateGasPrice)]
+    pub async fn get_estimate_gas_price(&self, priority: TxPriority) -> Result<f64> {
```
zvolin: same about removing `get_` prefix

grpc/src/tx.rs
zvolin: it shouldn't be needed, because `TxClient` implements `Deref<Target = GrpcClient>`

fl0rek: Nice, didn't notice that 👍 
I'm worried this doesn't work for uniffi/wasm_bindgen. I'll check.

grpc/src/grpc.rs
fl0rek: Also notfeeling too string about it, but I got the impression we should align docs/api with celestia for consistency.

grpc/src/grpc.rs
```diff
@@ -111,11 +116,11 @@ where
 
     // cosmos.base.tendermint
 
-    /// Get latest block
+    /// get_latest_block returns the latest block.
```
oblique: Not sure if I like the function name as a prefix

grpc/src/tx.rs
```diff
@@ -610,6 +653,7 @@ mod wbg {
                 gas_limit: value.gas_limit(),
                 gas_price: value.gas_price(),
                 memo: value.memo(),
+                priority: value.priority().unwrap_or(TxPriority::Medium),
```
oblique:
```suggestion
                priority: value.priority().unwrap_or_default(),
```

grpc/src/tx.rs
```diff
@@ -263,6 +231,41 @@ where
...
+                    .await?;
+                let gas_limit = (usage as f64 * DEFAULT_GAS_MULTIPLIER) as u64;
+                (gas_limit, maybe_gas_price.unwrap_or(price))
+            }
+        })
+    }
+
     async fn sign_and_broadcast_tx(&self, tx: RawTxBody, cfg: TxConfig) -> Result<(Hash, u64)> {
         let account = self.account.lock().await;
         let sign_tx = |tx, gas, fee| {
```
oblique: we can avoid having a closure now

grpc/src/grpc.rs
fl0rek: Actually you may be right, reverted.

grpc/src/grpc.rs
fl0rek: Sure, reworded a bit.

grpc/src/tx.rs
fl0rek: Nice, didn't notice that 👍

grpc/src/tx.rs
fl0rek: Js wasn't complaining so I didn't notice it. Thanks, removed!

oblique (approved):

zvolin (approved):

PR #689, opened 2025-07-17
feat(grpc): gRPC client builder

fl0rek:


grpc/Cargo.toml
```diff
@@ -74,3 +74,5 @@ wasm-bindgen-test.workspace = true
 default = []
 wasm-bindgen = ["celestia-types/wasm-bindgen", "dep:js-sys", "dep:wasm-bindgen", "dep:wasm-bindgen-futures"]
 uniffi = ["dep:uniffi", "dep:async-trait", "celestia-types/uniffi"]
+tls-native-roots = ["tonic/tls-native-roots"]
+tls-webpki-roots = ["tonic/tls-webpki-roots"]
```
fl0rek: Issue here is that we cannot enable either of tls-* flags by default, as it'll enable it by default for all platforms - including breaking wasm32.

fl0rek: fixed by using roots manually and not relying on tonic.

grpc/src/js_client.rs
```diff
@@ -1,2 +1,116 @@
...
+///
+/// const signer = (signDoc) => {
+///   const bytes = protoEncodeSignDoc(signDoc);
+///   const sig = secp256k1.sign(bytes, privKey, { prehash: true });
+///   return sig.toCompactRawBytes();
+/// };
+///
+/// const txClient = await GrpcClientBuilder
+///   .withUrl("http://127.0.0.1:18080")
+///   .withPub signer)
```
zvolin, line 43: typo

grpc/src/builder.rs
```diff
@@ -0,0 +1,188 @@
...
+            connection: (endpoint, tls_config.trust_anchors(roots)),
+            ..self
+        }
+    }
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl<S> GrpcClientBuilder<(Endpoint, ClientTlsConfig), S> {
+    /// Connects to the endpoint, applying tls configuration
+    pub fn connect(self) -> Result<GrpcClientBuilder<Channel, S>> {
```
zvolin, line 90: I don't like having to call both connect and build, can we handle that in build?

grpc/src/builder.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl GrpcClientBuilder<(Endpoint, ClientTlsConfig), ()> {
+    /// Create a new client connected to the given `url` using [`Channel`] transport
+    ///
+    /// [`Channel`]: tonic::transport::Channel
+    pub fn with_url(url: impl Into<String>) -> Result<Self, tonic::transport::Error> {
+        let endpoint = Endpoint::from_shared(url.into())?.user_agent("celestia-grpc")?;
+        let tls_config = ClientTlsConfig::new();
```
zvolin, line 42: this applies to almost all methods, can we just store information we need and move all the fallible logic to happen at the final call to build?

grpc/src/builder.rs
```diff
@@ -0,0 +1,188 @@
...
+impl<T> GrpcClientBuilder<T, ()>
+where
+    T: GrpcService<BoxBody> + Clone,
+    T::ResponseBody: Body<Data = Bytes> + Send + 'static,
+    <T::ResponseBody as Body>::Error: Into<StdError> + Send,
+{
+    /// Build [`GrpcClient`]
+    pub fn build_client(self) -> GrpcClient<T> {
+        GrpcClient::new(self.connection)
+    }
```
zvolin, line 161: I think I would rename `build` to `connect`, and if we were to keep both grpc and tx client, I'd call them `connect` and `connect_read_only` /  `connect_ro`

grpc/src/error.rs
```diff
@@ -79,6 +79,10 @@ pub enum Error {
     /// Signing error
     #[error(transparent)]
     SigningError(#[from] SignatureError),
+
+    /// Builder error
+    #[error("Keys and signer not set when building transaction client")]
+    MissingKeysAndSinger,
```
zvolin, line 85: either just missing signer or missing keys **or** signer :thinking:

grpc/src/builder.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub fn with_transport(transport: T) -> Self {
+        Self {
+            connection: transport,
+            signer: (),
+            account_pubkey: None,
+        }
+    }
+
+    /// Add signer and a public key
+    pub fn with_pubkey_and_signer<S>(
```
zvolin, line 126: can we also include the key bytes and hex options similar to what `celestia-client` has? or just move it from the `celestia-client` here, if it gets merged first

grpc/src/js_client.rs
```diff
@@ -1,2 +1,116 @@
...
+        Ok(Self {
+            url: self.url,
+            signer,
+            account_pubkey,
+        })
+    }
+
+    /// build gRPC read-only client. If you need to send messages, use [`build_tx_client`]
+    #[wasm_bindgen(js_name = buildClient)]
+    pub async fn build_client(self) -> Result<GrpcClient> {
```
zvolin, line 98: similar to rust, I'd rename to connect / connect_read_only

grpc/src/js_client/grpc_client.rs
```diff
@@ -11,21 +11,14 @@ use crate::Result;
 
 type InnerClient = crate::GrpcClient<tonic_web_wasm_client::Client>;
 
-/// Celestia GRPC client
+/// Celestia GRPC client, for builder see [`GrpcClientBuilder`]
```
zvolin, line 14: we could add `pub fn builder` here and in the tx client

grpc/src/uniffi_client.rs
```diff
@@ -1,4 +1,190 @@
...
+    signer: Option<Arc<dyn UniffiSigner>>,
+    account_pubkey: Option<VerifyingKey>,
+    native_roots: bool,
+    webpki_roots: bool,
+}
+
+#[uniffi::export(async_runtime = "tokio")]
+impl GrpcClientBuilder {
+    /// Create a new builder for the provided url
+    #[uniffi::constructor(name = "create")]
```
zvolin, line 66:
```suggestion
    #[uniffi::constructor(name = "withUrl")]
```

grpc/src/uniffi_client.rs
```diff
@@ -1,4 +1,190 @@
...
+        })
+    }
+
+    // this function _must_ be async despite not awaiting, so that it executes in tokio runtime
+    // context
+    /// build gRPC read-only client. If you need to send messages, use [`build_tx_client`]
+    ///
+    /// [`build_tx_client`]: GrpcClientBuilder::build_tx_client
+    #[uniffi::method(name = "buildClient")]
+    pub async fn build_client(self: Arc<Self>) -> Result<GrpcClient, GrpcClientBuilderError> {
```
zvolin, line 102: same about `connect` and `connectReadOnly`

fl0rek:
closing in favour: #712

PR #700, opened 2025-07-28
feat: implement intoHeight trait

### LINKED ISSUE #693: Create `IntoHeight` trait
Currently the height can be 3 types: `Height`, `u64`, `i64`. It would be nice to have a `IntoHeight` trait will allow any of those types to be used.

varun-doshi:
Fixes #693

### DIFF
--- a/types/src/height.rs
+++ b/types/src/height.rs
@@ -1,45 +1,41 @@
+use serde::Serialize;
 use tendermint::block::Height;
 
-pub trait IntoHeight {
+pub trait TryIntoHeight: Serialize + Sized {
     // Convert self into a Height, returning an error if conversion fails
-    fn into_height(self) -> Result<Height, HeightConversionError>;
+    fn try_into_height(self) -> Result<Height, HeightConversionError>;
 }
 
 // Error type for conversion failures
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
 pub enum HeightConversionError {
-    NegativeValue,
+    #[error("Height cannot be zero")]
+    ZeroHeight,
+    #[error("Height cannot be negative")]
+    NegativeHeight,
+    #[error("Value too large for Height")]
     Overflow,
 }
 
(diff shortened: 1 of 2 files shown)


varun-doshi:
@oblique i can switch out existing height assignments with trait based if you'd like. Anywhere specific to change?

types/src/lib.rs
```diff
@@ -17,6 +17,7 @@ pub mod evidence;
 mod extended_header;
 pub mod fraud_proof;
 pub mod hash;
+pub mod height;
```
zvolin: let's make the mod private and re-export the trait down below

types/src/height.rs
```diff
@@ -0,0 +1,47 @@
+use tendermint::block::Height;
+
+pub trait IntoHeight {
```
zvolin: I think since this can be fallible, we should rename it to `TryIntoHeight` and `try_into_height` respectively

types/src/height.rs
```diff
@@ -0,0 +1,47 @@
...
+    // Convert self into a Height, returning an error if conversion fails
+    fn into_height(self) -> Result<Height, HeightConversionError>;
+}
+
+// Error type for conversion failures
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum HeightConversionError {
```
zvolin, line 11: We should also disallow `0` as height and have respective error

types/src/height.rs
```diff
@@ -0,0 +1,47 @@
...
+    NegativeValue,
```
zvolin:
```suggestion
    NegativeHeight,
```

types/src/height.rs
```diff
@@ -0,0 +1,47 @@
+use tendermint::block::Height;
+
+pub trait IntoHeight {
+    // Convert self into a Height, returning an error if conversion fails
+    fn into_height(self) -> Result<Height, HeightConversionError>;
+}
+
+// Error type for conversion failures
+#[derive(Debug, Clone, PartialEq, Eq)]
```
zvolin: can you use `thiserror` for creating the error type? that's what we do for all other error types

PR #712, opened 2025-07-31
feat(grpc): Merge TxClient and GrpcClient, add builder

fl0rek:


client/src/lib.rs
```diff
@@ -56,6 +56,7 @@ pub mod tx {
     pub use k256::ecdsa::{Signature, SigningKey, VerifyingKey};
 }
 
+use celestia_grpc::GrpcClientBuilderError;
```
oblique, line 59:
```suggestion
use celestia_grpc::GrpcClientBuilderError;

```

grpc/Cargo.toml
```diff
@@ -38,17 +38,25 @@ ics23 = { version = "0.12.0", default-features = false, features = ["std"] }
...
 ]}
+tracing = "0.1.40"
 uniffi = { version = "0.29.2", optional = true, features = ["tokio"] }
+futures = "0.3.30"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { version = "1.38.0", features = ["time"] }
-tonic = { version = "0.12.3", default-features = false, features = [ "transport" ] }
-signature = "2.2.0"
+tonic = { version = "0.12.3", default-features = false, features = [ "transport", "tls-native-roots" ] }
```
oblique: Why do we have `tls-native-roots` here? Shouldn't we enable it with our feature gate?

grpc/Cargo.toml
```diff
@@ -38,17 +38,25 @@ ics23 = { version = "0.12.0", default-features = false, features = ["std"] }
...
+rustls-native-certs = { version = "0.8.1", optional = true }
+webpki-roots = { version = "1.0.2", optional = true }
+rustls-webpki = { version = "0.103.4", optional = true }
+rustls = { version = "0.23.29", default-features = false, features = [
+  "std", "ring"
+]}
```
oblique: I think this should be optional

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+impl<T> GrpcClient<T> {
+    /// Get the underlying transport.
+    pub fn into_inner(self) -> T {
+        self.transport
+    }
+}
+
+impl<T> GrpcClient<T>
+where
+    //S: DocSigner,
```
oblique:
```suggestion
```

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+}
+
+/// Struct wrapping all the tonic types and doing type conversion behind the scenes.
+pub struct GrpcClient<T> {
+    transport: T,
+
+    account: Mutex<Option<AccountBits>>,
+
+    signer: Option<SignerBits>,
+}
```
oblique:
```suggestion
pub struct GrpcClient<T> {
    transport: T,
    account: Mutex<Option<AccountBits>>,
    signer: Option<SignerBits>,
}
```

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+
+    app_version: AppVersion,
+    chain_id: Id,
+}
+
+pub struct SignerBits {
+    pub(crate) signer: DispatchedDocSigner,
+
+    pub(crate) pubkey: VerifyingKey,
+}
```
oblique:
```suggestion
struct AccountBits {
    account: Account,
    app_version: AppVersion,
    chain_id: Id,
}

pub struct SignerBits {
    pub(crate) signer: DispatchedDocSigner,
    pub(crate) pubkey: VerifyingKey,
}
```

grpc/src/signer.rs
```diff
@@ -0,0 +1,390 @@
...
+/// Signature error
+pub type SignatureError = k256::ecdsa::signature::Error;
+
+/// Signer capable of producing ecdsa signature using secp256k1 curve.
+pub trait DocSigner: Send + Sync {
+    /// Try to sign the provided sign doc.
+    fn try_sign(
+        &self,
+        doc: SignDoc,
+    ) -> Pin<Box<dyn Future<Output = Result<DocSignature, SignatureError>> + Send>>;
```
oblique: We want user to be able to just use `async fn`. Another option is to use `async_trait` and which can be boxed

```suggestion
    fn try_sign(
        &self,
        doc: SignDoc,
    ) -> impl Future<Output = Result<DocSignature, SignatureError>> + Send;
```

grpc/src/signer.rs
```diff
@@ -0,0 +1,390 @@
...
+    pub fn new<S>(signer: S) -> DispatchedDocSigner
+    where
+        S: DocSigner + 'static,
+    {
+        DispatchedDocSigner(Box::new(signer))
+    }
+}
+
+impl DocSigner for DispatchedDocSigner {
+    //async fn try_sign(&self, doc: SignDoc) -> Result<DocSignature, SignatureError> {
```
oblique:
```suggestion
```

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+    T: GrpcService<BoxBody> + Clone,
+    T::Error: Into<StdError>,
+    T::ResponseBody: Body<Data = Bytes> + Send + 'static,
+    <T::ResponseBody as Body>::Error: Into<StdError> + Send,
+{
+    /// Create a new client wrapping given transport
+    pub(crate) fn new(transport: T, signer: Option<SignerBits>) -> Self {
+        Self {
+            transport,
+            account: None.into(),
```
oblique: I got a bit confused for a bit, I think i prefer:

```suggestion
            account: Mutex::new(None),
```

grpc/src/signer.rs
```diff
@@ -0,0 +1,390 @@
...
+        f.write_str("DispatchedDocSigner { .. }")
+    }
+}
+
+/// Sign `tx_body` and the transaction metadata as the `base_account` using `signer`
+pub async fn sign_tx(
+    tx_body: RawTxBody,
+    chain_id: Id,
+    base_account: &BaseAccount,
+    signer: &SignerBits,
```
oblique: `SignerBits` is not exposed. Also I think I don't like it exposed.

I think it would be cleaner for the user to know only about `DocSigner` and `Keypair`.

grpc/src/signer.rs
```diff
@@ -0,0 +1,390 @@
...
+        &self,
+        doc: SignDoc,
+    ) -> Pin<Box<dyn Future<Output = Result<DocSignature, SignatureError>> + Send>> {
+        let bytes = doc.encode_to_vec();
+        Box::pin(ready(self.try_sign(&bytes)))
+    }
+}
+
+/// Workaround for dispatching `DocSigner`
+pub struct DispatchedDocSigner(Box<dyn DocSigner>);
```
oblique: We should keep this `pub(crate)`

grpc/src/builder.rs
```diff
@@ -0,0 +1,189 @@
...
+        Self {
+            transport_setup: NativeTransportBits {
+                url,
+                load_native_roots,
+                load_webpki_roots: true,
+            },
+            ..self
+        }
+    }
+}
```
zvolin: let's have only `.with_default_tls()` and allow user to use feature flags to control which roots were selected

zvolin: or `with_tls()`. We can just deprecate it when we'll be adding custom tls support

client/src/client.rs
```diff
@@ -267,12 +267,14 @@ impl ClientBuilder {
         let (pubkey, grpc) = match (&self.grpc_url, signer) {
             (Some(url), Some((pubkey, signer))) => {
                 #[cfg(not(target_arch = "wasm32"))]
-                let tx_client = TxClient::with_url(url, pubkey, signer).await?;
+                let builder = GrpcClientBuilder::with_url(url).with_native_roots();
```
zvolin: I think we should re-expose `tls-{native,webpki}-roots` feature flags in client and have `native` as default feature.

Can we make it in a way that if none of the `tls-xyz-roots` is selected we create here client without tls?

grpc/src/builder.rs
```diff
@@ -0,0 +1,189 @@
...
+use tonic::client::GrpcService;
+#[cfg(not(target_arch = "wasm32"))]
+use tonic::transport::{Channel, ClientTlsConfig, Endpoint};
+
+use crate::client::SignerBits;
+use crate::client::StdError;
+use crate::signer::DispatchedDocSigner;
+use crate::DocSigner;
+use crate::GrpcClient;
+use crate::GrpcClientBuilderError;
```
zvolin: can you group those?

grpc/src/builder.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn with_url(url: impl Into<String>) -> Self {
+        GrpcClientBuilder {
+            transport_setup: NativeTransportBits {
+                url: url.into(),
+                load_native_roots: false,
+                load_webpki_roots: false,
+            },
+            signer_bits: None,
+        }
+    }
```
zvolin: Can we expose all those starting points from the `GrpcClient` instead?

So it would look like:
```
let client = GrpcClient::with_url("abc://def")
    .with_default_tls()
    .build();
```

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+const DEFAULT_GAS_MULTIPLIER: f64 = 1.1;
+
+struct AccountBits {
+    account: Account,
+
+    app_version: AppVersion,
+    chain_id: Id,
+}
+
+pub struct SignerBits {
```
zvolin: missing docs, is it intended to be pub?

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+                    retries += 1;
+                    continue;
+                }
+                Err(e) => return Err(e),
+            }
+        };
+        self.confirm_tx(tx_hash, sequence).await
+    }
+
+    async fn account_load(&self) -> Result<(MappedMutexGuard<'_, AccountBits>, &SignerBits)> {
```
zvolin: maybe `load_account`?

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+    pub async fn app_version(&self) -> Result<AppVersion> {
+        let (account, _) = self.account_load().await?;
+        Ok(account.app_version)
+    }
+
+    /// Get client's chain id
+    pub async fn chain_id(&self) -> Result<Id> {
+        let (account, _) = self.account_load().await?;
+        Ok(account.chain_id.clone())
+    }
```
zvolin: first I thought it's a bit weird that those require signer, but they are not needed otherwise so it makes sense kinda. Can you put a piece of a doc on all methods that require signer that it needs to be configured in order for them to work?

grpc/src/error.rs
```diff
@@ -79,6 +79,40 @@ pub enum Error {
...
+    #[error(transparent)]
+    #[cfg(not(target_arch = "wasm32"))]
+    Webpki(#[from] webpki::Error),
+
+    /// Could not import system certificates
+    #[cfg(not(target_arch = "wasm32"))]
+    #[error("Could not import platform certificates: {errors:?}")]
+    RustlsNativeCerts {
+        /// List of errors
+        errors: Vec<rustls_native_certs::Error>,
```
zvolin: since we reference types from those crates, those variants should be properly feature gated. However it is rather a bad practice in enums to have different variants depending on features being enabled, so I think we should stringify them instead or erase types in other way (not sure what they contain)

grpc/src/error.rs
```diff
@@ -93,3 +127,10 @@ impl From<Error> for wasm_bindgen::JsValue {
         error.to_string().into()
     }
 }
+
+#[cfg(not(target_arch = "wasm32"))]
+impl From<Vec<rustls_native_certs::Error>> for GrpcClientBuilderError {
```
zvolin: this should also be feature gated

grpc/src/client.rs
```diff
@@ -0,0 +1,624 @@
...
+
+impl Keypair for SignerBits {
+    type VerifyingKey = VerifyingKey;
+
+    fn verifying_key(&self) -> Self::VerifyingKey {
+        self.pubkey
+    }
+}
+
+/// Struct wrapping all the tonic types and doing type conversion behind the scenes.
```
zvolin: I know that it's just moved, but could we rephrase this doc? I'd like first line to just say that it's a grpc client for the celestia consensus node

grpc/src/js_client.rs
```diff
@@ -1,2 +1,100 @@
...
+#[wasm_bindgen]
+impl GrpcClientBuilder {
+    /// Create a builder for [`GrpcClient`] connected to `url`
+    #[wasm_bindgen(js_name = withUrl)]
+    pub fn with_url(url: String) -> Self {
+        Self {
+            inner: RustBuilder::with_grpcweb_url(url),
+        }
+    }
+
```
zvolin: can we move it to the client?

grpc/src/js_client.rs
```diff
@@ -1,2 +1,100 @@
...
+        signer_fn: JsSignerFn,
+    ) -> Result<Self> {
+        let signer = JsSigner::new(signer_fn);
+        let account_pubkey = VerifyingKey::try_from(account_pubkey.to_vec().as_slice())?;
+        Ok(Self {
+            inner: self.inner.with_pubkey_and_signer(account_pubkey, signer),
+        })
+    }
+
+    /// build gRPC read-only client. If you need to send messages, use [`build_tx_client`]
```
zvolin: leftover doc

grpc/src/js_client/grpc_client.rs
```diff
@@ -153,12 +165,113 @@ impl GrpcClient {
...
     pub async fn tx_status(&self, hash: &str) -> Result<TxStatusResponse> {
         self.client.tx_status(hash.parse()?).await
     }
+
+    /// estimate_gas_price takes a transaction priority and estimates the gas price based
+    /// on the gas prices of the transactions in the last five blocks.
+    ///
+    /// If no transaction is found in the last five blocks, return the network
+    /// min gas price.
+    #[wasm_bindgen(js_name = getEstimateGasPrice)]
```
zvolin:
```suggestion
    #[wasm_bindgen(js_name = estimateGasPrice)]
```

grpc/src/lib.rs
```diff
@@ -1,18 +1,24 @@
...
 mod js_client;
+pub mod signer;
 mod tx;
 #[cfg(all(not(target_arch = "wasm32"), feature = "uniffi"))]
 pub mod uniffi_client;
 mod utils;
 
+pub use crate::builder::GrpcClientBuilder;
+pub use crate::client::GrpcClient;
+pub use crate::error::GrpcClientBuilderError;
```
zvolin: merge the import with the one below

grpc/Cargo.toml
fl0rek: There was a weird situation here: tls-*-roots flags cannot be enabled for wasm32, this causes tonic to pull mio, which does not compile there. On the other hand having at least one of the flags is required for tonic to expose tls types, and we need that for non-wasm32. 
First workaround was the above, enable one of the flags for the appropriate platform. Switched to using meta-package wich pulls appropriate deps and flags for the appropriate platform.

grpc/Cargo.toml
fl0rek: Removed entirely.

grpc/src/error.rs
fl0rek: removed

grpc/src/error.rs
fl0rek: those errors got removed, using tonic directly now

grpc/src/client.rs
fl0rek: sure, same thing 👍

grpc/src/client.rs
fl0rek: added a note to both 👍

client/src/client.rs
fl0rek: Switched to `Client` using all roots that were enabled via feature flags (or none if none were).

grpc/src/builder.rs
fl0rek: Switched to single function and exact roots being loaded is driven by feature flags 👍

grpc/src/signer.rs
fl0rek: This is taken from celestia-client, so used across crates already

grpc/src/signer.rs
fl0rek: Pulled back the SignerBits expansion, it should be internal between the builder and the client now.

grpc/src/client.rs
fl0rek: per Yiannis request above, this got hidden now.

grpc/src/signer.rs
fl0rek: went with `async_trait` to alow boxing 👍

grpc/Cargo.toml
```diff
@@ -30,32 +30,37 @@ tendermint-proto.workspace = true
...
+tracing = "0.1.40"
 uniffi = { version = "0.29.2", optional = true, features = ["tokio"] }
+futures = "0.3.30"
+http = "1.3.1"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { version = "1.38.0", features = ["time"] }
-tonic = { version = "0.12.3", default-features = false, features = [ "transport" ] }
-signature = "2.2.0"
+tonic = { workspace = true, features = [ "transport", "_tls-any" ] }
```
zvolin: do we need `_tls-any` here now that we don't create tls providers on our end? it will pull `rustls` even with `default-features = false`

grpc/src/client.rs
```diff
@@ -0,0 +1,642 @@
...
+struct AccountBits {
+    account: Account,
+    app_version: AppVersion,
+    chain_id: Id,
+}
+
+pub(crate) struct SignerBits {
+    pub(crate) signer: DispatchedDocSigner,
+    pub(crate) pubkey: VerifyingKey,
+}
```
zvolin: could we `s/Bits/Config/g` or `Conf`? I tried to get used to this name but my mind keeps connecting it to bytes somehow

grpc/src/signer.rs
```diff
@@ -0,0 +1,399 @@
...
+impl<T> KeypairExt for T
+where
+    T: Keypair,
+    T::VerifyingKey: Into<AccAddress>,
+{
+    fn address(&self) -> AccAddress {
+        self.verifying_key().into()
+    }
+}
+
```
zvolin: this seem to only be used [here](https://github.com/eigerco/lumina/pull/712/files#diff-8b2379cf48ef774d4757c1ecd32705abea258152d0c229b2b8afde6628f28681R427), no?

In that case I'd prefer removing this and just using `self.signer.pubkey.into()` there

grpc/src/client.rs
```diff
@@ -0,0 +1,642 @@
...
+
+impl Keypair for SignerBits {
+    type VerifyingKey = VerifyingKey;
+
+    fn verifying_key(&self) -> Self::VerifyingKey {
+        self.pubkey
+    }
+}
```
zvolin: If we do [this](https://github.com/eigerco/lumina/pull/712/files#r2298279280), then this is only used in places we could as well pass `self.pubkey`, no? if yes, we could remove that

grpc/src/builder.rs
fl0rek: done

grpc/Cargo.toml
fl0rek: removed 👍

grpc/src/client.rs
fl0rek: Settled on `SignerConfig` and `AccountState`, since the latter isn't really a config, but a cached account state we're using for txns.

grpc/src/signer.rs
fl0rek: got carried away here, removed ;)

grpc/src/client.rs
fl0rek: Removed all the keypair stuff 👍

grpc/src/builder.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// Create a gRPC client builder using provided prepared transport
+    pub fn with_transport<B, T>(transport: T) -> Self
+    where
+        //B: Body<Data = Bytes> + Send + Sync + Unpin + 'static,
+        B: http_body::Body + AbstractBody + Send + Unpin + 'static,
```
oblique: I just tested locally and the line without the `AbstractBody` compiles. Is there any reason you left it there? I prefer all the boxed mechanism to be hidden from the user.

client/src/client.rs
```diff
@@ -1,7 +1,8 @@
 use std::fmt::{self, Debug};
 use std::sync::Arc;
 
-use celestia_grpc::TxClient;
+use celestia_grpc::signer::DispatchedDocSigner;
```
oblique: I prefer this to be hidden from the user. We can use generics in the `ClientBuilder` and keep the erasure only for `Client`.

client/src/client.rs
```diff
@@ -14,18 +15,11 @@ use crate::state::StateApi;
...
 use crate::types::state::AccAddress;
 use crate::types::ExtendedHeader;
-use crate::utils::DispatchedDocSigner;
 use crate::{Error, Result};
 
-#[cfg(not(target_arch = "wasm32"))]
-type Transport = tonic::transport::Channel;
-
-#[cfg(target_arch = "wasm32")]
-type Transport = tonic_web_wasm_client::Client;
```
oblique, line 24: I think it would be nice to be exposed with `ClientBuilder::grpc_transport`

grpc/Cargo.toml
```diff
@@ -30,32 +30,38 @@ tendermint-proto.workspace = true
...
 ]}
+tracing = "0.1.40"
 uniffi = { version = "0.29.2", optional = true, features = ["tokio"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { version = "1.38.0", features = ["time"] }
-tonic = { version = "0.12.3", default-features = false, features = [ "transport" ] }
-signature = "2.2.0"
+tonic = { workspace = true, features = [ "transport" ] }
+rustls-native-certs = { version = "0.8.1", optional = true }
```
oblique: I think this can be removed

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+        };
+        let Some(result) = ready else {
+            return Poll::Ready(None);
+        };
+
+        Poll::Ready(Some(result.map_err(box_err)))
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
```
zvolin:
```suggestion
```

zvolin: `http_body::Body` I think can cover everything

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+    T: http_body::Body<Data = Bytes>,
+    <T as http_body::Body>::Error: StdError + Send + Sync + 'static,
+{
+    fn poll_frame_inner(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Frame<Bytes>, BoxError>>> {
+        let Poll::Ready(ready) = self.poll_frame(cx) else {
+            return Poll::Pending;
+        };
```
zvolin: use `std::task::ready`
```suggestion
        let ready = ready!(self.poll_frame(cx));
```

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        Poll::Ready(Some(result.map_err(box_err)))
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+impl<T> AbstractBody for T
+where
+    T: http_body::Body<Data = Bytes>,
```
zvolin, line 66: we have tonic body as `TonicBody` so we could just replace qualified path with use everywhere :thinking:

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    fn poll_frame(
+        mut self: Pin<&mut BoxedBody>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Frame<Self::Data>, Self::Error>>> {
+        AbstractBody::poll_frame_inner(Pin::new(self.inner.deref_mut()), cx)
+    }
+}
+
+pub trait AbstractBody {
```
oblique: make this `pub(crate)`

grpc/src/builder.rs
```diff
@@ -0,0 +1,179 @@
...
+pub struct GrpcClientBuilder {
+    transport: TransportSetup,
+    signer_bits: Option<SignerConfig>,
+}
+
+impl GrpcClientBuilder {
+    /// Create a new client connected to the given `url` using [`Channel`] transport
+    ///
+    /// [`Channel`]: tonic::transport::Channel
+    pub fn with_url(url: impl Into<String>) -> Self {
```
oblique: Not a strong opinion: I think when `Builder` is in the struct name, we can avoid `with_`. I think `with_` is used mostly on structs with `Config` in their name and on constructors.

grpc/src/client.rs
```diff
@@ -0,0 +1,617 @@
...
+                .await
+            {
+                Ok(resp) => break resp,
+                Err(Error::TxBroadcastFailed(_, ErrorCode::InsufficientFee, _)) if retries < 3 => {
+                    retries += 1;
+                    continue;
+                }
+                Err(e) => return Err(e),
+            }
+        };
```
oblique: Don't forget to change this and other loops to the new flow

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+#[cfg(target_arch = "wasm32")]
+mod conditional_send {
+    /// Use [`ConditionalSend`] to mark an optional Send trait bound. Useful as on certain platforms (eg. WASM),
+    /// futures aren't Send.
+    pub trait ConditionalSend {}
+
+    impl<T> ConditionalSend for T {}
+}
+
+pub use conditional_send::*;
```
zvolin: can we copy it from `blockstore` instead to run only one impl?

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn call(&mut self, req: http::Request<TonicBody>) -> AbstractFuture {
+        Box::pin(Service::call(self, req).map(|response| match response {
+            Ok(response) => Ok(response.boxed()),
+            Err(e) => Err(box_err(e)),
+        }))
+    }
+}
+
+type BoxedResponse = http::Response<BoxedBody>;
+type AbstractFuture = BoxedFuture<'static, Result<BoxedResponse, BoxError>>;
```
zvolin: this breaks the convention with `Abstract` being a trait. Can we just define `BoxedFuture` here without that helper?

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+use bytes::Bytes;
+use dyn_clone::{clone_box, DynClone};
+use futures::FutureExt;
+use http_body::Frame;
+use tonic::body::Body as TonicBody;
+use tonic::codegen::Service;
+
+dyn_clone::clone_trait_object!(AbstractTransport);
+
+type BoxError = Box<dyn StdError + Sync + Send + 'static>;
```
zvolin:
```suggestion
type BoxedError = Box<dyn StdError + Sync + Send + 'static>;
```
let's keep names consistent, this way or the other

grpc/src/builder.rs
zvolin: yeah, I agree we should keep this hidden

grpc/src/error.rs
```diff
@@ -79,6 +79,31 @@ pub enum Error {
...
+#[derive(thiserror::Error, Debug)]
+pub enum GrpcClientBuilderError {
+    /// Error from tonic transport
+    #[error(transparent)]
+    #[cfg(not(target_arch = "wasm32"))]
+    TonicTransportError(#[from] tonic::transport::Error),
+
+    /// Attempted to enable tls when using pre-prepared transport
+    #[error("Cannot enable tls on manual transport")]
+    CannotEnableTlsOnManualTransport,
```
zvolin:
```suggestion
    #[error("Cannot enable tls on custom transport")]
    CannotEnableTlsOnCustomTransport,
```

grpc/src/signer.rs
```diff
@@ -0,0 +1,376 @@
...
+        /// signature bytes
+        pub bytes: Vec<u8>,
+    }
+
+    #[async_trait]
+    impl DocSigner for UniffiSignerBox {
+        async fn try_sign(
+            &self,
+            doc: SignDoc,
+            //) -> Pin<Box<dyn Future<Output = Result<Secp256k1Signature, K256Error>> + Send>> {
```
oblique:
```suggestion
```

grpc/src/signer.rs
```diff
@@ -0,0 +1,376 @@
...
+            }
+            /*
+            let signer = self.0.clone();
+            Box::pin(async move {
+                match signer.sign(doc).await {
+                    Ok(s) => Secp256k1Signature::try_from(s).map_err(K256Error::from_source),
+                    Err(e) => Err(K256Error::from_source(e)),
+                }
+            })
+                */
```
oblique:
```suggestion
```

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+            inner: clone_box(&*self.inner),
+        }
+    }
+}
+
+trait AbstractTransport: DynClone {
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), BoxError>>;
```
zvolin:
```suggestion
        cx: &mut Context<'_>,
    ) -> Poll<Result<(), BoxError>>;
```

grpc/src/boxed.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+pub use conditional_send::*;
+
+pub trait ConditionalSendFuture: Future + ConditionalSend {}
+
+impl<T: Future + ConditionalSend> ConditionalSendFuture for T {}
+
+/// An owned and dynamically typed Future used when you can't statically type your result or need to add some indirection.
+pub type BoxedFuture<'a, T> = core::pin::Pin<Box<dyn ConditionalSendFuture<Output = T> + 'a>>;
```
zvolin:
```suggestion
pub type BoxedFuture<'a, T> = Pin<Box<dyn ConditionalSendFuture<Output = T> + 'a>>;
```

node-uniffi/Cargo.toml
```diff
@@ -33,4 +33,12 @@ tokio = { version = "1.38.0", features = ["macros", "sync"] }
...
-tonic = "0.12.3"
+tonic.workspace = true
+
+[features]
+# webpki is needed for TLS on android and iOS
+# https://github.com/hyperium/tonic/issues/2369
+default = ['tls-webpki-roots']
+tls-native-roots = ["celestia-grpc/tls-native-roots"]
+tls-webpki-roots = ["celestia-grpc/tls-webpki-roots"]
+
```
oblique:
```suggestion
```

oblique:
Nice work!

grpc/src/error.rs
```diff
@@ -79,6 +79,31 @@ pub enum Error {
...
     #[error(transparent)]
     SigningError(#[from] SignatureError),
+
+    /// No account
+    #[error("no account")]
+    NoAccount,
+
+    /// Builder error
+    #[error("Keys and signer not set when building transaction client")]
+    MissingKeysAndSinger,
```
oblique:
```suggestion
    /// Signer was not set on construction
    #[error("Signer was not set on construction")]
    SignerNotSet,
```

grpc/src/signer.rs
```diff
@@ -0,0 +1,366 @@
...
+    T: Signer<DocSignature> + Send + Sync + 'static,
+{
+    async fn try_sign(&self, doc: SignDoc) -> Result<DocSignature, SignatureError> {
+        let bytes = doc.encode_to_vec();
+        self.try_sign(&bytes)
+    }
+}
+
+/// Workaround for dispatching `DocSigner`
+pub struct DispatchedDocSigner(Box<dyn DocSigner>);
```
oblique: Since `DocSigner` is now implemented with `async_trait`, we don't need the dispatching workaround, you can just `Box<dyn DocSigner>`

grpc/src/js_client/grpc_client.rs
```diff
@@ -1,42 +1,48 @@
...
-use crate::grpc::{GasInfo, GetTxResponse, JsBroadcastMode, TxStatusResponse};
+use crate::grpc::{GasInfo, GetTxResponse, JsBroadcastMode, TxPriority, TxStatusResponse};
+use crate::js_client::GrpcClientBuilder;
+use crate::tx::{JsTxConfig, JsTxInfo};
 use crate::Result;
 
-type InnerClient = crate::GrpcClient<tonic_web_wasm_client::Client>;
-
-/// Celestia GRPC client
+/// Celestia GRPC client, for builder see [`GrpcClientBuilder`]
```
oblique:
```suggestion
/// Celestia gRPC client, for builder see [`GrpcClientBuilder`]
```

grpc/src/js_client/grpc_client.rs
```diff
@@ -153,12 +170,113 @@ impl GrpcClient {
...
+    #[wasm_bindgen(js_name = txStatus)]
     pub async fn tx_status(&self, hash: &str) -> Result<TxStatusResponse> {
         self.client.tx_status(hash.parse()?).await
     }
+
+    /// estimate_gas_price takes a transaction priority and estimates the gas price based
+    /// on the gas prices of the transactions in the last five blocks.
+    ///
+    /// If no transaction is found in the last five blocks, return the network
+    /// min gas price.
```
oblique: Updated comment from `client.rs` 

```suggestion
    /// Estimate gas price for given transaction priority based
    /// on the gas prices of the transactions in the last five blocks.
    ///
    /// If no transaction is found in the last five blocks, return the network
    /// min gas price.
```

grpc/src/js_client/grpc_client.rs
```diff
@@ -153,12 +170,113 @@ impl GrpcClient {
...
+    #[wasm_bindgen(js_name = appVersion, getter)]
+    pub async fn app_version(&self) -> Result<JsAppVersion> {
+        Ok(self.client.app_version().await?.into())
+    }
+
+    /// Submit blobs to the celestia network.
+    ///
+    /// When no `TxConfig` is provided, client will automatically calculate needed
+    /// gas and update the `gasPrice`, if network agreed on a new minimal value.
+    /// To enforce specific values use a `TxConfig`.
```
oblique: This part was removed from `client.rs` because we don't do any calculation.

```suggestion
```

grpc/src/js_client/grpc_client.rs
```diff
@@ -153,12 +170,113 @@ impl GrpcClient {
...
+        let tx_config = tx_config.map(Into::into).unwrap_or_default();
+        let tx = self.client.submit_blobs(&blobs, tx_config).await?;
+        Ok(tx.into())
+    }
+
+    /// Submit message to the celestia network.
+    ///
+    /// When no `TxConfig` is provided, client will automatically calculate needed
+    /// gas and update the `gasPrice`, if network agreed on a new minimal value.
+    /// To enforce specific values use a `TxConfig`.
```
oblique:
```suggestion
```

grpc/src/uniffi_client/grpc_client.rs
```diff
@@ -190,6 +219,55 @@ impl GrpcClient {
...
+    /// AppVersion of the client
+    pub async fn app_version(&self) -> Result<AppVersion> {
+        Ok(self.client.app_version().await?)
+    }
+
+    /// Submit blobs to the celestia network.
+    ///
+    /// When no `TxConfig` is provided, client will automatically calculate needed
+    /// gas and update the `gasPrice`, if network agreed on a new minimal value.
+    /// To enforce specific values use a `TxConfig`.
```
oblique:
```suggestion
```

grpc/src/uniffi_client/grpc_client.rs
```diff
@@ -190,6 +219,55 @@ impl GrpcClient {
...
+        let blobs = Vec::from_iter(blobs.into_iter().map(Arc::<Blob>::unwrap_or_clone));
+        let config = config.unwrap_or_default();
+        Ok(self.client.submit_blobs(&blobs, config).await?)
+    }
+
+    /// Submit message to the celestia network.
+    ///
+    /// When no `TxConfig` is provided, client will automatically calculate needed
+    /// gas and update the `gasPrice`, if network agreed on a new minimal value.
+    /// To enforce specific values use a `TxConfig`.
```
oblique:
```suggestion
```

grpc/src/uniffi_client/grpc_client.rs
```diff
@@ -180,7 +209,7 @@ impl GrpcClient {
     /// min gas price.
     ///
     /// The gas used is estimated using the state machine simulation.
-    async fn estimate_gas_price_and_usage(
+    pub async fn estimate_gas_price_and_usage(
         &self,
         priority: TxPriority,
         tx_bytes: Vec<u8>,
```
oblique, line 236: Sync their doc comments with `client.rs`

grpc/src/signer.rs
fl0rek: hidden 👍

client/src/client.rs
fl0rek: moved the dispatched part to grpc per daily

grpc/src/boxed.rs
fl0rek: TonicBody is a struct, this here is a trait. Tonic re-exports http_body::Body as tonic::transport::Body, but this if feature flagged and enabling it pulls a bunch of wasm32 incompatible deps, so we cannot use it.

grpc/src/boxed.rs
fl0rek: I knew we had it somewhere, but totally forgot about blockstore, thanks!

grpc/src/boxed.rs
fl0rek: good point, updated

grpc/src/signer.rs
fl0rek: Per daily, due to dyn-incompatibility and bunch of other considerations, settled on hiding the dispatching in grpc crate entirely.

client/src/client.rs
fl0rek: Exposed grpc_transport (pulled few needed deps 😬 for trait bounds).
wdym about above types? they weren't pub above, should I just export them in lib.rs?

grpc/src/builder.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    use tonic::transport::{ClientTlsConfig, Endpoint};
+
+    pub(super) fn build_transport(
+        url: String,
+        tls: bool,
+    ) -> Result<BoxedTransport, tonic::transport::Error> {
+        let mut tls_config = ClientTlsConfig::new();
+
+        #[cfg(any(feature = "tls-native-roots", feature = "tls-webpki-roots"))]
```
zvolin:
```suggestion
```

grpc/src/error.rs
```diff
@@ -79,6 +79,39 @@ pub enum Error {
...
+/// Representation of all the errors that can occur when building [`GrpcClient`] using
+/// [`GrpcClientBuilder`]
+///
+/// [`GrpcClient`]: crate::GrpcClient
+/// [`GrpcClientBuilder`]: crate::GrpcClientBuilder
+#[derive(thiserror::Error, Debug)]
+pub enum GrpcClientBuilderError {
+    /// Error from tonic transport
+    #[error(transparent)]
+    #[cfg(not(target_arch = "wasm32"))]
```
zvolin, line 101: @oblique do you think it's okay to have this cfg in public enum? If some lib uses that, they'd need to also include it, no?

grpc/src/js_client.rs
```diff
@@ -1,2 +1,96 @@
...
+use grpc_client::GrpcClient;
+
+/// Builder for [`GrpcClient`] and [`TxClient`].
+///
+/// Url must point to a [grpc-web proxy](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md).
+///
+/// # Keyless client example
+///
+/// ```js
+/// const grpcClient = await GrpcClientBuilder
```
zvolin: rename all variables to just `client`, we have `grpcClient` here and `txClient` below

grpc/src/js_client.rs
```diff
@@ -1,2 +1,96 @@
...
+///   .withUrl("http://127.0.0.1:18080")
+///   .buildClient()
```
zvolin:
```suggestion
/// const client = await GrpcClient
///   .withUrl("http://127.0.0.1:18080")
///   .build()
```

grpc/src/js_client.rs
```diff
@@ -1,2 +1,96 @@
...
+/// const signer = (signDoc) => {
+///   return window.leap.signDirect("mocha-4", keys.bech32Address, signDoc, { preferNoSetFee: true })
+///     .then(sig => Uint8Array.from(atob(sig.signature.signature), c => c.charCodeAt(0)))
+/// }
+///
+/// const txClient = await GrpcClientBuilder
+///   .withUrl("http://127.0.0.1:18080")
+///   .withPubkeyAndSigner(keys.pubKey, signer)
+///   .buildTxClient()
+/// ```
```
zvolin:
```suggestion
/// const client = await GrpcClient
///   .withUrl("http://127.0.0.1:18080")
///   .withPubkeyAndSigner(keys.pubKey, signer)
///   .build()
/// ```
```

grpc/src/js_client.rs
```diff
@@ -1,2 +1,96 @@
...
+/// const privKey = "fdc8ac75dfa1c142dbcba77938a14dd03078052ce0b49a529dcf72a9885a3abb";
+/// const pubKey = secp256k1.getPublicKey(privKey);
+///
+/// const signer = (signDoc) => {
+///   const bytes = protoEncodeSignDoc(signDoc);
+///   const sig = secp256k1.sign(bytes, privKey, { prehash: true });
+///   return sig.toCompactRawBytes();
+/// };
+///
+/// const txClient = await GrpcClient
```
zvolin:
```suggestion
/// const client = await GrpcClient
```

grpc/src/js_client.rs
```diff
@@ -1,2 +1,96 @@
...
+        })
+    }
+
+    /// build gRPC client
+    #[wasm_bindgen(js_name = build)]
+    pub async fn build(self) -> Result<GrpcClient> {
+        Ok(self
+            .inner
+            .build()
+            .expect("client creation successful")
```
zvolin: can't we `?` here?

grpc/src/js_client.rs
```diff
@@ -1,2 +1,96 @@
...
+///   .buildTxClient()
+/// ```
+#[wasm_bindgen]
+pub struct GrpcClientBuilder {
+    inner: crate::GrpcClientBuilder,
+}
+
+#[wasm_bindgen]
+impl GrpcClientBuilder {
+    /// Add public key and signer to the client being built
```
zvolin, line 78: Consuming and returning may not be obvious to javascript guys :thinking: I'm not sure what would be the best angle here, but if we keep it that way, let's add this to all methods:

```suggestion
    /// Add public key and signer to the client being built
    ///
    /// Note that this method **consumes** builder and returns updated instance of it.
    /// Make sure to re-assign it if you keep builder in a variable.
```

grpc/src/uniffi_client.rs
```diff
@@ -1,4 +1,133 @@
...
+    /// Create a new builder for the provided url
+    #[uniffi::constructor(name = "withUrl")]
+    pub fn with_url(url: String) -> Self {
+        GrpcClientBuilder {
+            url,
+            signer: None,
+            account_pubkey: None,
+            tls: false,
+        }
+    }
```
zvolin: In case of uniffi, we cannot keep the flow of having this?

```
GrpcClient.withUrl(...).build()
```

zvolin (commented):
:clap:

grpc/src/uniffi_client.rs
fl0rek: Unfortunately, no. This is uniffi limitation - it does not support static class methods
https://github.com/mozilla/uniffi-rs/issues/1074

grpc/src/js_client.rs
fl0rek: Taking mutable reference would on the other hand break the more obvious use-case of chaining the calls, so this felt like a better compromise. Added a note to both methods.

grpc/src/error.rs
oblique: Yes they need to include it

grpc/src/uniffi_client.rs
zvolin: Can you maybe put such comment somewhere so I don't ask this question in the future? :p

fl0rek: Added 👍

client/src/client.rs
oblique: I don't remember why I wrote this comment on those types. I only wanted `ClientBuilder::grpc_transport`

Cargo.toml
```diff
@@ -9,17 +9,20 @@ lumina-node-wasm = { version = "0.10.3", path = "node-wasm" }
...
 nmt-rs = "0.2.1"
-prost = "0.13.3"
-prost-build = "0.13.3"
-prost-types = "0.13.3"
+prost = "0.13.5"
+prost-build = "0.13.5"
+prost-types = "0.13.5"
+tonic = { version = "0.13.1", default-features = false }
+tonic-web-wasm-client = "0.7.1"
+tonic-build = { version = "0.13.1" }
```
oblique:
```suggestion
tonic-build = "0.13.1"
```

grpc/src/builder.rs
```diff
@@ -0,0 +1,235 @@
...
+        tls: bool,
+    },
+    BoxedTransport(BoxedTransport),
+}
+
+/// Builder for [`GrpcClient`]
+#[derive(Default)]
+pub struct GrpcClientBuilder {
+    transport: TransportSetup,
+    signer_bits: Option<SignerKind>,
```
oblique:
```suggestion
    signer_kind: Option<SignerKind>,
```

grpc/src/signer.rs
```diff
@@ -0,0 +1,382 @@
...
+pub type SignatureError = k256::ecdsa::signature::Error;
+
+/// Helper struct used to pass type erased signer. To implement custom signer, see
+/// [`DocSigner`] trait.
+pub(crate) struct BoxedDocSigner(Box<dyn AbstractDocSigner>);
+
+/// Signer capable of producing ecdsa signature using secp256k1 curve.
+pub trait DocSigner: Send + Sync {
+    /// Try to sign the provided sign doc.
+    //async fn try_sign(&self, doc: SignDoc) -> Result<DocSignature, SignatureError>;
```
oblique:
```suggestion
```

grpc/Cargo.toml
```diff
@@ -30,32 +30,39 @@ tendermint-proto.workspace = true
 tendermint.workspace = true
 lumina-utils = { workspace = true, features = ["time"] }
 
-async-trait = { version = "0.1.88", optional = true }
+async-trait = { version = "0.1.88", optional = true}
```
oblique:
```suggestion
async-trait = { version = "0.1.88", optional = true }
```

client/src/client.rs
```diff
@@ -253,30 +258,19 @@ impl ClientBuilder {
...
-                #[cfg(target_arch = "wasm32")]
-                let tx_client = TxClient::with_grpcweb_url(url, pubkey, signer).await?;
-
-                (Some(pubkey), Some(tx_client))
-            }
-            (Some(_url), None) => return Err(Error::SignerNotSet),
-            (None, Some((_pubkey, _signer))) => return Err(Error::GrpcEndpointNotSet),
-            (None, None) => (None, None),
+            (false, false) => (None, None),
+            (false, true) => return Err(Error::SignerNotSet),
```
oblique, line 272: TODO for another PR. We should allow this and return an error if the gRPC call needs a signer

grpc/src/uniffi_client.rs
```diff
@@ -1,4 +1,134 @@
...
+    #[cfg(any(feature = "tls-native-roots", feature = "tls-webpki-roots"))]
+    #[uniffi::method(name = "withDefaultTls")]
+    pub fn default_tls(self: Arc<Self>) -> Self {
+        GrpcClientBuilder {
+            url: self.url.clone(),
+            signer: self.signer.clone(),
+            account_pubkey: self.account_pubkey,
+            tls: true,
+        }
+    }
```
oblique: This is not needed anymore

grpc/src/uniffi_client/grpc_client.rs
oblique: You forgot to sync the doc comment of `estimate_gas_price_and_usage`

client/src/client.rs
fl0rek: created https://github.com/eigerco/lumina/issues/745

grpc/src/uniffi_client/grpc_client.rs
fl0rek: thanks, fixed now

zvolin (approved):
:clap:

PR #719, opened 2025-08-09
Move to singular `Blob::new` constructor

nuke-web3:
### Motivation

In working constructing a `Blob` it's inconvenient to have to make logic to handle if a blob is signed or not, for us it lead to hours of headache trying to troubleshoot why some blobs would work and others would not [example here](https://github.com/celestiaorg/eq-service/pull/110).

To ensure it's not a footgun and make less boilerplate for users of `Blob`, this PR makes a singular interface to construct with `Blob::new`.

## What's included

- Removes `Blob::new_with_signer`
- Add optional `signer` to `Blob::new`

---

Note for maintainers: I am sure there was some reason to split it into two constructors... maybe because this is a type braking change and requires action from all users? Happy to have this closed if it's not aligned with goals, or continue to polish as needed.

### DIFF
--- a/types/src/blob.rs
+++ b/types/src/blob.rs
@@ -95,7 +95,7 @@ impl Blob {
     ///     }"#},
     /// );
     /// ```
-    pub fn new(namespace: Namespace, data: Vec<u8>, app_version: AppVersion) -> Result<Blob> {
+    pub fn new(namespace: Namespace, data: Vec<u8>, app_version: AppVersion, signer: Option<AppVersion>) -> Result<Blob> {
         let share_version = appconsts::SHARE_VERSION_ZERO;
         let commitment =
             Commitment::from_blob(namespace, &data[..], share_version, None, app_version)?;
@@ -106,7 +106,7 @@ impl Blob {
             share_version,
             commitment,
             index: None,
-            signer: None,
+            signer,
         })
     }
 
--- a/types/src/blob.rs
+++ b/types/src/blob.rs
@@ -68,11 +68,16 @@ pub struct BlobParams {
 }
 
 impl Blob {
-    /// Create a new blob with the given data within the [`Namespace`].
+    /// Create a new blob with the given data within the [`Namespace`], with optional signer.
+    ///
+    /// # Notes
+    ///
+    /// If present onchain, `signer` was verified by consensus node on blob submission.
     ///
     /// # Errors
     ///
     /// This function propagates any error from the [`Commitment`] creation.
+    /// To use `signer = Some(address)`, [`AppVersion`] must be at least [`AppVersion::V3`].
     ///
     /// # Example
     ///
@@ -80,11 +85,11 @@ impl Blob {
     /// use celestia_types::{AppVersion, Blob, nmt::Namespace};
     ///
     /// let my_namespace = Namespace::new_v0(&[1, 2, 3, 4, 5]).expect("Invalid namespace");
-    /// let blob = Blob::new(my_namespace, b"some data to store on blockchain".to_vec(), AppVersion::V2)
+    /// let blob_unsigned = Blob::new(my_namespace, b"some data to store on blockchain".to_vec(), AppVersion::V2, None)
     ///     .expect("Failed to create a blob");
     ///
     /// assert_eq!(
-    ///     &serde_json::to_string_pretty(&blob).unwrap(),
+    ///     &serde_json::to_string_pretty(&blob_unsigned).unwrap(),
     ///     indoc::indoc! {r#"{
     ///       "namespace": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQIDBAU=",
     ///       "data": "c29tZSBkYXRhIHRvIHN0b3JlIG9uIGJsb2NrY2hhaW4=",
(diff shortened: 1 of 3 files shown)


types/src/blob.rs
```diff
@@ -722,6 +721,62 @@ mod tests {
...
+          "commitment": "AUpLKHYnlrfK0cX6DcGyryFMld1nJia+cjkCwXFTFgA=",
+          "index": -1,
+          "signer": "j71qdnFJas04ncZ4/CazBpFlSWE="
+        }"#};
+
+        assert_eq!(
+            serde_json::from_str::<serde_json::Value>(&actual_signed).unwrap(),
+            serde_json::from_str::<serde_json::Value>(expected_signed).unwrap()
+        );
+    }
```
nuke-web3: Main thrust of changes here, reflected in all doc comments that are tested with `cargo test --doc` passing.

Still need to test if commitment logic is still correct. It should be, but needs e2e test really

nuke-web3:
chipping away at CI errors, but pausing here till I get feedback this PR is likely to be accepted or not at all :sweat_smile:

zvolin:
Hey, sorry for the delay there, we were discussing that topic, and we reached an agreement that your proposed changes are better in terms of API than what we had. And you were mostly right with this, that's the main reason we just added `_with_signer()` that time
> maybe because this is a type braking change and requires action from all users?

Anyways, this has green light from us, thanks :+1:

nuke-web3:
I think https://github.com/eigerco/lumina/actions/runs/17557648490/job/49865847903?pr=719 is OK to let fail, assuming it's fix in some other PR - lmk if I should go ahead and fix, otherwise I think this PR is good to go :crossed_fingers: 
Now just to decide when it's released - I can target some staging branch other than `main` if you need, lmk :grin:

rpc/README.md
```diff
@@ -11,13 +11,13 @@ use celestia_types::{AppVersion, Blob, nmt::Namespace};
 async fn submit_blob() {
     // create a client to the celestia node
     let token = std::env::var("CELESTIA_NODE_AUTH_TOKEN").expect("Token not provided");
-    let client = Client::new("ws://localhost:36658", Some(&token))
+    let client = Client::new("ws://localhost:36658")
```
fl0rek: This is causing ci failures. Why was the argument removed here, seems unrelated? re https://github.com/eigerco/lumina/pull/719#issuecomment-3267114443

types/src/blob.rs
zvolin: it is correct. The end to end test happens in `celestia-rpc`

types/src/blob.rs
```diff
@@ -722,6 +721,62 @@ mod tests {
...
+          "share_version": 0,
+          "commitment": "m0A4feU6Fqd5Zy9td3M7lntG8A3PKqe6YdugmAsWz28=",
+          "index": -1,
+          "signer": null
+        }"#};
+
+        assert_eq!(
+            serde_json::from_str::<serde_json::Value>(&actual_unsigned).unwrap(),
+            serde_json::from_str::<serde_json::Value>(expected_unsigned).unwrap()
+        );
```
zvolin: this test seems a little bit inconvenient as a unit test and not part of the docs. What I'd prefer here (and below) was to:

- create blob with(out) signer
- call the `.to_shares()`
- check if first share returns expected share version and signer

types/src/blob.rs
```diff
@@ -68,66 +68,65 @@ pub struct BlobParams {
...
-        let share_version = appconsts::SHARE_VERSION_ONE;
+        let mut share_version = appconsts::SHARE_VERSION_ZERO;
+
+        if signer.is_some() {
+            let app_version = app_version.as_u64();
+            if app_version < 3 {
+                return Err(Error::UnsupportedAppVersion(app_version));
+            }
+            share_version = appconsts::SHARE_VERSION_ONE;
+        }
```
zvolin: we can rely on `Commitment::from_blob` on performing all the validation

```suggestion
        let share_version = if signer.is_none() {
            appconsts::SHARE_VERSION_ZERO
        } else {
            appconsts::SHARE_VERSION_ONE
        };
```

types/src/blob.rs
```diff
@@ -68,66 +68,65 @@ pub struct BlobParams {
...
-    /// use celestia_types::{AppVersion, Blob, nmt::Namespace};
+    /// use celestia_types::{AppVersion, Blob, nmt::Namespace, state::AccAddress};
     ///
     /// let my_namespace = Namespace::new_v0(&[1, 2, 3, 4, 5]).expect("Invalid namespace");
-    /// let blob = Blob::new(my_namespace, b"some data to store on blockchain".to_vec(), AppVersion::V2)
-    ///     .expect("Failed to create a blob");
+    /// let signer: AccAddress = "celestia1377k5an3f94v6wyaceu0cf4nq6gk2jtpc46g7h"
+    ///     .parse()
+    ///     .unwrap();
+    /// let blob_signed = Blob::new(
```
zvolin: Could we remove all the occurrences of word `signed` in this example? It is a regular example on `Blob::new` and I don't think emphasizing this part helps in readability / understanding. Also the blob is not technically signed, it just stores info who intends to sign it.

types/src/blob.rs
```diff
@@ -68,66 +68,65 @@ pub struct BlobParams {
...
-    ///     indoc::indoc! {r#"{
-    ///       "namespace": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQIDBAU=",
-    ///       "data": "c29tZSBkYXRhIHRvIHN0b3JlIG9uIGJsb2NrY2hhaW4=",
-    ///       "share_version": 0,
-    ///       "commitment": "m0A4feU6Fqd5Zy9td3M7lntG8A3PKqe6YdugmAsWz28=",
-    ///       "index": -1,
-    ///       "signer": null
-    ///     }"#},
+    ///     serde_json::from_str::<serde_json::Value>(&actual_signed).unwrap(),
+    ///     serde_json::from_str::<serde_json::Value>(expected_signed).unwrap()
```
zvolin: could we maybe restore the original look of the assert?

```rust
assert_eq!(
    serde_json::to_string_pretty(&blob).unwrap(),
    indoc::indoc! {r#"{
      "namespace": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQIDBAU=",
      "data": "c29tZSBkYXRhIHRvIHN0b3JlIG9uIGJsb2NrY2hhaW4=",
      "share_version": 1,
      "commitment": "AUpLKHYnlrfK0cX6DcGyryFMld1nJia+cjkCwXFTFgA=",
      "index": -1,
      "signer": "j71qdnFJas04ncZ4/CazBpFlSWE="
    }"#},
);
```

I like it for readability, going back and forth through serde just adds noise, especially that we compare `Value`. In case there's some issue with that (idk, field order during serialization :thinking:) then I would prefer comparing `Blob`s with a single call to `from_str` on the hardcoded json

fl0rek (approved):
LGTM 🙌

Issue #683, opened 2025-07-11
Remove workarounds for `Header` and `Proofs` json encoding


zvolin:
We have workarounds to be able to deserialize `u64` in non-tendermint format (not stringified), that can be removed after next api breaking release of celestia-node https://github.com/celestiaorg/celestia-node/milestone/5.

https://github.com/celestiaorg/celestia-node/pull/3928
https://github.com/celestiaorg/celestia-node/pull/3930
https://github.com/celestiaorg/celestia-node/pull/3929

PR #720, opened 2025-08-12
feat(proto,types,rpc)!: upgrade to celestia-node v0.25

zvolin:
Aligns with the breaking changes in celestia-node v0.25.

It should be still compatible with older node versions (for now, until we pull the trigger and remove compat layer with #683) except for the `header.GetRangeByHeight` which will fail for older nodes as they cannot deserialize quoted height in commit.


proto/src/serializers.rs
```diff
@@ -3,6 +3,7 @@
 pub mod bytes;
 pub mod cow_str;
 pub mod from_str;
+pub mod maybe_from_str;
```
fl0rek: nit: I think it's more `from_maybe_str`

rpc/tests/utils/mod.rs
```diff
@@ -30,6 +30,7 @@ pub fn random_ns_range(start: Namespace, end: Namespace) -> Namespace {
 }
 
 pub fn random_bytes(length: usize) -> Vec<u8> {
+    let length = 1.max(length);
```
fl0rek: what's the motivation for the change here? All calls in this PR use literal values.

zvolin: idk why but I had in my mind that it creates array of random size up to length and slapped that quickly without even thinking :p reverted

proto/src/serializers.rs
zvolin: I was thinking about many names here, also `from_t_or_str`, which I love and hate kinda, wdyt?

oblique: or what about `maybe_quoted`?

zvolin: with `maybe_quoted` I'd like to rename `from_str` to `quoted`

rpc/src/share.rs
```diff
@@ -359,6 +368,18 @@ fn is_ods_square(row: u64, column: u64, square_width: u16) -> bool {
...
+    data: Vec<u8>,
+}
+
+// get samples version of sample, differs from [`RawSample`] in `share`
+#[derive(Default, Deserialize)]
+#[serde(default)]
+struct GetSamplesSample {
+    share: GetSamplesShare,
+    proof: RawProof,
+    proof_type: i32,
```
oblique: Move this above with the other types

zvolin: removed with new node version

PR #723, opened 2025-08-14
feat(node-wasm): Identity persistance in browser

fl0rek:


node-wasm/src/lock.rs
```diff
@@ -0,0 +1,210 @@
...
+        // https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API#advanced_use
+        let (promise, resolve_fn, _reject) = promise_with_resolvers();
+
+        let (tx, rx) = oneshot::channel();
+
+        let cb: Function = Closure::once_into_js(move |lock: JsValue| {
+            let _ = tx.send(lock);
+            promise
+        })
+        .unchecked_into();
```
zvolin: I think I'd prefer simplifying this to something like:
```suggestion
        let (unlock_tx, unlock_rx) = oneshot::channel();
        let (would_block_tx, would_block_rx) = oneshot::channel();

        let cb: Function = Closure::once_into_js(move |lock: JsValue| {
            future_to_promise(async move {
                let _ = would_block_tx.send(lock.is_falsy());
                let _ = unlock_rx.await;
                Ok(JsValue::null())
            })
        })
        .unchecked_into();
```

and then have the `Guard` contain sth like:
```rust
    unlock_tx: Option<oneshot::Sender<()>>,
```

This way we don't need to add `Promise` bindings and calling js function in drop

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,210 @@
...
+
+impl NamedLock {
+    pub fn create(name: impl Into<String>) -> Result<Self, Error> {
+        let lock_manager = get_lock_manager().map_err(Error::LockManagerUnavailable)?;
+
+        Ok(NamedLock {
+            lock_manager,
+            name: name.into(),
+        })
+    }
```
zvolin: I think this overall structure is mostly to eventually upstream it to the `named_lock`?
since for now we only use it with `KeyRegistry::lock_key(&keypair).await?` I would personally simplify this, remove `NamedLock` and keep only the guard (maybe rename it to `NamedLock`?) and have `lock` and `try_lock` as constructors.
Up to you, but imo we don't need to match the api exactly right now

node-wasm/src/key_registry.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        let db = open_request.await?;
+
+        Ok(Self {
+            db: SendWrapper::new(db),
+        })
+    }
+
+    async fn get_first_unused_key(&self) -> Result<Option<(Keypair, NamedLockGuard)>> {
+        const KEY_BATCH_SIZE: usize = 512;
```
zvolin: I'd rename it to `MAX_KEYS` and remove pagination. 512 tabs with lumina side by side should be enough :p

node-wasm/src/client.rs
```diff
@@ -363,20 +371,39 @@ impl WasmNodeConfig {
...
+    ) -> Result<(NamedLockGuard, NodeBuilder<WasmBlockstore, WasmStore>)> {
         let network = network::Network::from(self.network);
         let network_id = network.id();
 
+        let (keypair, guard) = if let Some(key) = self.identity_key {
+            let keypair = Keypair::ed25519_from_bytes(key).context("could not decode key")?;
+            let guard = KeyRegistry::lock_key(&keypair).await?;
+
+            (keypair, guard)
+        } else if self.use_persistent_memory {
```
zvolin: I think we should keep the key in persistent memory regardless of this setting, and we shouldn't have an ephemeral key option at all

node-wasm/src/key_registry.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// IndexedDb contains invalid data
+    #[error("could not deserialise keypair data from db")]
+    InvalidKeypairData,
+
+    /// IndexedDb error coming from idb
+    #[error("idb error: {0}")]
+    IdbError(#[from] idb::Error),
+
+    #[error("key already locked")]
+    KeyLocked,
```
zvolin, line 34: is it possible we ever return this variant to the user? for me it seems like we should remove this variant

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,210 @@
...
+    async fn lock_unlock() {
+        let lock = NamedLock::create("foo").expect("lock creation ok");
+
+        {
+            let _guard = lock.try_lock().await.expect("valid lock");
+            let _locked_guard = lock.try_lock().await.expect_err("locked lock");
+        }
+
+        // XXX: a bit nasty, but we need to yield back to js for it to register the unlock
+        sleep(Duration::from_millis(1)).await;
```
zvolin: maybe use `lumina_utils` there

```suggestion
        yield_now().await;
```

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,210 @@
...
+        return Err(JsError::new("`navigator` not found in global scope"));
+    };
+    match Reflect::get(&navigator, &lock_manager_key) {
+        Ok(manager) => Ok(manager.unchecked_into::<LockManager>()),
+        Err(_) => Err(JsError::new("`navigator.locks` not found in global scope")),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
zvolin, line 125: it'd be nice to see a test where a blocking call acquires lock after some previously locked one is dropped

node-wasm/src/key_registry.rs
```diff
@@ -0,0 +1,181 @@
...
+use serde_wasm_bindgen::to_value;
+use tracing::{info, warn};
+use wasm_bindgen::{JsCast, JsValue};
+
+use crate::lock::{Error as LockError, NamedLock, NamedLockGuard};
+
+const DB_NAME: &str = "identity_registry";
+const DB_VERSION: u32 = 1;
+
+const KEY_STORE_NAME: &str = "🗝️";
```
zvolin: can we use ascii there? :p

node-wasm/src/worker.rs
```diff
@@ -60,6 +61,7 @@ pub struct NodeWorker {
 struct NodeWorkerInstance {
     node: Node<WasmBlockstore, WasmStore>,
     events_channel_name: String,
+    _key_lock: NamedLockGuard,
```
zvolin: nit, maybe something more descriptive like `p2p_identity_lock`?

node-wasm/src/lock.rs
zvolin: or we can use `lumina_utils::Token` instead of oneshot, up to you

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,210 @@
...
+    #[wasm_bindgen(method, structural, js_class = "LockManager", js_name = query)]
+    /// The `query()` method.
+    /// [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/LockManager/query)
+    pub fn query(this: &LockManager) -> Promise;
+
+    #[wasm_bindgen(method, structural, js_class = "LockManager", js_name = request)]
+    /// The `request()` method.
+    /// [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/LockManager/request)
+    pub fn request_with_callback(this: &LockManager, name: &str, callback: &Function) -> Promise;
+
```
zvolin: let's remove the ones we don't use, wdyt?

node-wasm/src/key_registry.rs
fl0rek: This may be returned, if user provides custom libp2p key that is equal to already locked key (most probably starting node twice with the same config, with identity key set). It's a case of protecting the user from themselves, but it doesn't cost us much here I think.

node-wasm/src/lock.rs
fl0rek: Yea it was a bit of a streach to fit the named_lock API exactly, simplified it here to a single struct.

node-wasm/src/key_registry.rs
fl0rek: It'll be less cool, but I guess I can change it :P

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,186 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Call would block")]
+    WouldBlock,
+
+    #[error("Could not acquire lock manager")]
+    LockManagerUnavailable(JsError),
+}
+
+pub struct NamedLock {
```
zvolin, line 22: I can't see drop impl or any other unlocking, am I missing sth?

zvolin: oh, actually we unlock on drop, but returning an error on receiver side. I mean, it works :shrug: but it isn't very obvious

zvolin: actually we can keep that behavior if we use the  token lock guard from `utils`

node-wasm/src/lock.rs
fl0rek: added blocking test 👍

node-wasm/src/lock.rs
fl0rek: Switched to token 👍

node-wasm/src/key_registry.rs
```diff
@@ -0,0 +1,166 @@
...
+                if let Some(key_and_lock) = k {
+                    return Ok(key_and_lock);
+                }
+            }
+            Err(e) => {
+                info!("error loading keys from IndexedDb: {e}");
+            }
+        }
+
+        info!("Persisting new keypair to indexeddb");
```
zvolin: please make those `debug`

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        /// The Promise.withResolvers() static method returns an object containing a new
+        /// Promise object and two functions to resolve or reject it, corresponding to the
+        /// two parameters passed to the executor of the Promise() constructor.
+        ///
+        /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers
+        #[wasm_bindgen(static_method_of = Promise, js_name = "withResolvers")]
+        pub fn with_resolvers() -> JsValue;
+    }
+}
```
zvolin:
```suggestion
```

node-wasm/src/client.rs
```diff
@@ -363,20 +371,37 @@ impl WasmNodeConfig {
...
+
+            (keypair, guard)
+        } else {
+            let registry = KeyRegistry::new().await?;
+            registry.get_key().await?
+        };
+
         let mut builder = if self.use_persistent_memory {
-            let store = IndexedDbStore::new(network_id)
+            let peer_id = keypair.public().to_peer_id().to_base58();
```
oblique: `PeerId` implements `Display` which actually call `to_base58` internally. We can remove `to_base58` here.

node-wasm/src/key_registry.rs
```diff
@@ -0,0 +1,166 @@
...
+        store.add(&encode_keypair(&keypair), None)?.await?;
+
+        tx.commit()?;
+
+        Ok((keypair, guard))
+    }
+
+    pub async fn lock_key(keypair: &Keypair) -> Result<NamedLock> {
+        try_lock_key(keypair).await
+    }
```
oblique: I think we should keep only `try_lock_key` because `lock_key` doesn't block until the key is released

node-wasm/src/key_registry.rs
```diff
@@ -0,0 +1,166 @@
...
+                info!("error loading keys from IndexedDb: {e}");
+            }
+        }
+
+        info!("Persisting new keypair to indexeddb");
+
+        let keypair = Keypair::generate_ed25519();
+        let guard = try_lock_key(&keypair)
+            .await
+            .expect("newly generated key to be unique");
```
oblique:
```suggestion
            .expect("newly generated key needs to be unique");
```

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,170 @@
+// TODO: upstream to `named_lock`
```
oblique, line 1: ❤️

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,170 @@
...
+        };
+
+        // fn returns a promise that gets resolved when the callback returns,
+        // which is useless for us. Actual lock is gets passed to the callback
+        let _promise = lock_manager.request_with_options_and_callback(name, &opts, &cb);
+
+        let would_block = would_block_rx.await.expect("valid singleshot channel");
+
+        // only case where callback above is called with a falsy value is when we're in
+        // non-blocking mode and the lock is already taken
```
oblique: move this comment in the callback, above the line that contains `lock.is_falsy()`

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,170 @@
...
+        NamedLock::lock_impl(name, true).await
+    }
+
+    async fn lock_impl(name: &str, block: bool) -> Result<NamedLock, Error> {
+        let lock_manager = get_lock_manager().map_err(Error::LockManagerUnavailable)?;
+
+        let unlock_token = Token::new();
+        let unlock_token_guard = unlock_token.trigger_drop_guard();
+        let (would_block_tx, would_block_rx) = oneshot::channel();
+        let cb: Function = Closure::once_into_js(move |lock: JsValue| {
```
oblique, line 46: Add a comment here. Something like this:

> This callback is called in two cases:
>
> * If the lock is granted
> * If `ifAvailable` was set to `true` buck the lock was not granted. In this case however `lock.is_falsy()` will return `true`.

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,170 @@
...
+            future_to_promise(async move {
+                let _ = would_block_tx.send(lock.is_falsy());
+                unlock_token.triggered().await;
```
oblique, line 53: I prefer a comment here. Something like this:

```suggestion
                // The lock is held until this callback returns. This will keep the lock alive
                // until the `unlock_token` is triggered.
                unlock_token.triggered().await;
```

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,170 @@
...
+pub enum Error {
+    #[error("Call would block")]
+    WouldBlock,
+
+    #[error("Could not acquire lock manager")]
+    LockManagerUnavailable(JsError),
+}
+
+pub struct NamedLock {
+    _unlock_guard: TokenTriggerDropGuard,
```
oblique, line 23: If JS side crashes, and our code stops running, will the JS engine release the lock?

node-wasm/src/lock.rs
```diff
@@ -0,0 +1,170 @@
...
+        Ok(NamedLock {
+            _unlock_guard: unlock_token_guard,
+        })
+    }
+}
+
+impl fmt::Debug for NamedLock {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        //write!(f, "NamedLock { .. }", self.name)
+        f.write_str("NamedLock { .. }")
```
oblique:
```suggestion
        f.write_str("NamedLock { .. }")
```

node-wasm/src/lock.rs
zvolin: Until callback returns -> until future resolves 
Or sth like that

node-wasm/src/client.rs
fl0rek: TIL! Fixed thanks 👍

node-wasm/src/key_registry.rs
fl0rek: Good point, renamed to `try_lock_key` for consistency.

node-wasm/src/lock.rs
fl0rek: Moved and reworded a bit.

node-wasm/src/lock.rs
fl0rek: Added comment

node-wasm/src/lock.rs
fl0rek: Added comment, slightly different wording, same effect.

node-wasm/src/lock.rs
fl0rek: If the tab crashes (sad face on chrome, I've trigged it via simple loop that consumed all the memory), this releases all the locks. Tested it on chrome and opening a new tab after a crash causes it to correctly start with the same key as the one that just died.

PR #732, opened 2025-09-02
feat(grpc)!: expose whole node config instead just gas price

zvolin:


fl0rek (approved):

grpc/src/grpc/node.rs
```diff
@@ -1,21 +1,58 @@
...
+    pub pruning_interval: u64,
+
+    /// A height at which the node should stop advancing state.
+    pub halt_height: u64,
+}
+
+impl FromGrpcResponse<ConfigResponse> for RawConfigResponse {
+    fn try_from_response(self) -> Result<ConfigResponse> {
+        let minimum_gas_price = if self.minimum_gas_price.is_empty() {
+            None
```
oblique, line 33: Is there a valid case where a node will return an empty `minimum_gas_price`?

zvolin: Yes, in V6 nodes don't substitute default here so if user left this empty in config, it's gonna be none

PR #733, opened 2025-09-02
feat(types)!: Add support for app v6

zvolin:


types/src/consts.rs
```diff
@@ -80,12 +80,27 @@ pub mod appconsts {
...
+        pub const SQUARE_SIZE_UPPER_BOUND: usize = 128;
+        /// Maximum width of a single subtree root when generating blob's commitment.
+        pub const SUBTREE_ROOT_THRESHOLD: u64 = 64;
+        /// Cost of each byte in a transaction (in units of gas).
+        pub const TX_SIZE_COST_PER_BYTE: u64 = 10;
+        /// Cost of each byte in blob (in units of gas).
+        pub const GAS_PER_BLOB_BYTE: u64 = 8;
+    }
+
+    // celestia-app/pkg/appconsts/app_consts.go
```
fl0rek: I wonder whether it'd be better to link to a specific revision on gh here. Otherwise they all link to the same file but 
at different times, since celestia does not do the versioning the way we do, just updates.
Not necessarily pushing to update it here, more thinking aloud.

zvolin: good idea

fl0rek (approved):
🕸️

oblique (approved):

PR #734, opened 2025-09-02
chore(client,grpc,types): Make all the celestia-client returned types Serialisable

fl0rek:


types/src/state/query_delegation.rs
```diff
@@ -269,10 +348,64 @@ impl TryFrom<RawRedelegationEntry> for RedelegationEntry {
...
+                                creation_height: entry.creation_height.into(),
+                                completion_time: Some(entry.completion_time.into()),
+                                initial_balance: entry.initial_balance.to_string(),
+                                shares_dst: entry.dest_shares.to_string(),
+                                unbonding_id: entry.unbonding_id,
+                                unbonding_on_hold_ref_count: entry.unbonding_on_hold_ref_count,
+                            }
+                        }
+                        ).collect(),
+                        */
```
zvolin:
```suggestion
```

client/src/test_utils.rs
```diff
@@ -73,3 +74,6 @@ pub(crate) fn node0_address() -> AccAddress {
     let s = include_str!("../../ci/credentials/node-0.addr");
     s.trim().parse().expect("invalid account address")
 }
+
+pub(crate) fn is_serialisable_deserialisable<'a, T: Serialize + Deserialize<'a>>(_: T) {}
+pub(crate) fn is_serialisable<T: Serialize>(_: T) {}
```
zvolin: wow, I didn't see that coming, nice :+1: 
I thought that, maybe we want to add also negative ones, with `!`, and put them on the methods we are sure are not serializable / deserializable, to make it easier to not forget to cover something (by requiring adding methods to the tests) and not to reason if something should be covered or not in future?

types/src/state/query_delegation.rs
fl0rek: whoops, missed it during cleanup

client/src/test_utils.rs
fl0rek: There were only two methods that were not covered, `sync_wait` and `shares_available`, both in the await-for-the-condition variety which effectively return ().
I've added them as special cases with `let _ : () = ...`. Originally I didn't have them, as it tripped [unit_arg](https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg) lint, though we could silence it too.

So currently all the methods are covered in one way or another.

fl0rek: Not sure how we could enforce adding checks for future added methods.
Maybe some proc-macro? 🤔

zvolin: Sorry, I for some reason thought that for Shwap types we have no Serialize, ignore me on the negative bounds. And for enforcing I thought just a comment to add new ones on API or remembering that fact on reviews :p

client/src/header.rs
```diff
@@ -136,3 +136,39 @@ impl HeaderApi {
...
+
+        ensure_serializable_deserializable(api.network_head().await.unwrap());
+
+        let hash = ensure_serializable_deserializable(unimplemented!());
+        ensure_serializable_deserializable(api.get_by_hash(hash).await.unwrap());
+
+        ensure_serializable_deserializable(api.get_by_height(0).await.unwrap());
+        let header = ensure_serializable_deserializable(unimplemented!());
+
+        ensure_serializable_deserializable(api.get_range_by_height(&header, 0).await.unwrap());
```
oblique:
```suggestion

        let header = ensure_serializable_deserializable(unimplemented!());
        ensure_serializable_deserializable(api.get_range_by_height(&header, 0).await.unwrap());
```

grpc/src/tx.rs
```diff
@@ -483,7 +484,7 @@ where
 }
 
 /// A result of correctly submitted transaction.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
```
zvolin:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```

zvolin (approved):

types/src/state/query_delegation.rs
```diff
@@ -85,16 +92,37 @@ impl TryFrom<RawQueryDelegationResponse> for QueryDelegationResponse {
     }
 }
 
+impl From<QueryDelegationResponse> for RawQueryDelegationResponse {
+    fn from(value: QueryDelegationResponse) -> Self {
+        // XXX: can we assume denom is utia?
+        let balance = Coin::utia(value.response.balance);
```
oblique: Util now we can and if you notice in `celestia-client` I'm always returning `f64` instead of `Coin`. However, if you want to make it 1:1, you can use `Coin` only in `QueryDelegationResponse`, but keep the `f64` in `celestia-client`.

fl0rek: left it as is and removed the comment

PR #741, opened 2025-09-12
ci(uniffi): Fix (🤞) for swift test flakiness

fl0rek:


zvolin (commented):
I'm not exactly sure why this fixes the issue, since we connect to mainnet where current height will always be bigger than `2`

PR #742, opened 2025-09-14
feat!: unify and upgrade dependencies, add explicit msrv

zvolin:


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[target.wasm32-unknown-unknown]
+rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
```
fl0rek, line 2: 🙈

CONTRIBUTING.md
```diff
@@ -18,19 +18,18 @@ We follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
...
 
-How to upgrade:
-```bash
-./tools/upgrade-deps.sh -i  # `-i` upgrades incompatible versions too
-cargo update
-```
+We try not to enforce the latest versions of dependencies to give some freedom for the dependency resolver of dependant crates.
+When adding or upgrading dependencies, try to set the minimal required version of a crate, e.g. if current latest version of
+`serde-json` is `1.0.256`, you can try setting it to `1`. The CI job will then check what is the minimal required version and
+suggest e.g. `1.0.101`, which you then will have to update in the respective `toml` file.
```
fl0rek:
```suggestion
We don't enforce the latest dependency versions to give more flexibility in dependency resolution to the dependant crates.
When adding or upgrading dependencies, try setting the minimal required version of a crate, e.g. if current latest version of
`serde-json` is `1.0.256`, start with version `1`. The CI job will then check what is the minimal required version and
suggest e.g. `1.0.101`, which you then can use in the respective `toml` file.
```
High concentration of `try` and `depend` made it hard for me to parse. Couldn't get rid of dependency, but reduced the try. Feel free to apply/ignore parts you find appealing.

CONTRIBUTING.md
```diff
@@ -18,19 +18,18 @@ We follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
...
+
+If some dependency is used in multiple crates of a workspace, then try putting this dependency in `workspace.dependencies` section
+instead of repeating it in multiple crates.
```
fl0rek:
```suggestion
If a dependency is used in multiple crates of the workspace, set it in `workspace.dependencies` section
instead of repeating it in multiple crates.
```
I think we an just drop the try-indirection here

Cargo.toml
```diff
@@ -1,35 +1,94 @@
...
 
 [patch.crates-io]
 # Uncomment to apply local changes
-#beetswap = { path = "../beetswap" }
-#blockstore = { path = "../blockstore" }
+# beetswap = { path = "../beetswap" }
+# blockstore = { path = "../blockstore" }
 #nmt-rs = { path = "../nmt-rs" }
 #libp2p = { path = "../../rust-libp2p/libp2p" }
 #libp2p-core = { path = "../../rust-libp2p/core" }
```
fl0rek:
```suggestion
# nmt-rs = { path = "../nmt-rs" }
# libp2p = { path = "../../rust-libp2p/libp2p" }
# libp2p-core = { path = "../../rust-libp2p/core" }
```

Cargo.toml
```diff
@@ -1,38 +1,97 @@
...
-#beetswap = { path = "../beetswap" }
-#blockstore = { path = "../blockstore" }
-#nmt-rs = { path = "../nmt-rs" }
-#libp2p = { path = "../../rust-libp2p/libp2p" }
-#libp2p-core = { path = "../../rust-libp2p/core" }
+# beetswap = { path = "../beetswap" }
+# blockstore = { path = "../blockstore" }
+# nmt-rs = { path = "../nmt-rs" }
+# libp2p = { path = "../../rust-libp2p/libp2p" }
+# libp2p-core = { path = "../../rust-libp2p/core" }
```
oblique, line 94: fix also the two lines below

oblique (approved):

PR #748, opened 2025-09-17
feat(grpc)!: Add support for attaching metadata to requests

zvolin:


grpc/src/client.rs
```diff
@@ -358,7 +336,133 @@ impl GrpcClient {
...
+        AsyncGrpcCall::new(move |context| async move {
+            this.submit_blobs_impl(&blobs, cfg, &context).await
+        })
+        .context(&self.inner.context)
+    }
+
+    /// Get client's app version
+    ///
+    /// Note that this function _may_ try to load the account information from the network and as
+    /// such requires signed to be set up.
```
```suggestion
    /// such requires signing to be set up.
```

grpc/src/client.rs
```diff
@@ -358,7 +336,133 @@ impl GrpcClient {
...
+            let ChainState { app_version, .. } = this.load_chain_state(&context).await?;
+            Ok(app_version)
+        })
+        .context(&self.inner.context)
+    }
+
+    /// Get client's chain id
+    ///
+    /// Note that this function _may_ try to load the account information from the network and as
+    /// such requires signed to be set up.
```
```suggestion
    /// such requires signing to be set up.
```

grpc/src/client.rs
```diff
@@ -358,7 +336,133 @@ impl GrpcClient {
...
+}
+
+impl GrpcClient {
+    async fn get_verified_balance_impl(
+        &self,
+        address: &Address,
+        header: &ExtendedHeader,
+        context: &Context,
+    ) -> Result<Coin> {
+        // construct the key for querying account's balance from bank state
```
```suggestion
        // construct the key for querying account balance from bank state
```

grpc/src/client.rs
fl0rek: Also, with readonly grpc, we now load chain and account info separately, so I think this doesn't apply anyway.

fl0rek (approved):
👏 
just the note about the comment

grpc/src/client.rs
zvolin: yeah, we missed it in #755

grpc/src/grpc.rs
```diff
@@ -36,6 +42,177 @@ pub use crate::grpc::cosmos_tx::JsBroadcastMode;
...
+    /// Appends whole metadata map to the current metadata map.
+    pub(crate) fn append_metadata_map(&mut self, metadata: &MetadataMap) {
+        for key_and_value in metadata.iter() {
+            match key_and_value {
+                KeyAndValueRef::Ascii(key, val) => {
+                    self.metadata.append(key, val.clone());
+                }
+                KeyAndValueRef::Binary(key, val) => {
+                    self.metadata.append_bin(key, val.clone());
+                }
```
oblique, line 96: Should we also check for duplicates here?

grpc/src/grpc.rs
```diff
@@ -36,6 +42,177 @@ pub use crate::grpc::cosmos_tx::JsBroadcastMode;
...
+    /// Keys for binary metadata must have `-bin` suffix.
+    pub fn metadata_bin(mut self, key: &str, val: &[u8]) -> Result<Self, MetadataError> {
+        self.context.append_metadata_bin(key, val)?;
+        Ok(self)
+    }
+
+    /// Append a metadata map to the grpc request.
+    pub fn metadata_map(self, metadata: MetadataMap) -> Self {
+        let context = Context { metadata };
+        self.context(&context)
```
oblique: I think calling `self.context.append_metadata_map` is more correct. We may have more stuff in `Context` later on.

oblique (approved):

PR #751, opened 2025-09-19
feat(node): Node identity retention for native,ios,android

fl0rek:


node/src/store.rs
```diff
@@ -1208,6 +1229,47 @@ mod tests {
...
+    async fn libp2p_identity_persistance<S: Store>(
+        #[case]
+        #[future(awt)]
+        s: S,
+    ) {
+        let store = s;
+        let generated_keypair = store.init_identity(None).await.unwrap();
+        let persisted_keypair = store.init_identity(None).await.unwrap();
+
+        assert_eq!(generated_keypair.public(), persisted_keypair.public());
```
zvolin: I'm not sure I understand this test

node/src/store.rs
```diff
@@ -183,6 +190,10 @@ pub enum StoreError {
     /// Failed to open the store.
     #[error("Error opening store: {0}")]
     OpenFailed(String),
+
+    /// Attempting to set libp2p identity for a store that already has keypair generated
+    #[error("Attempting to set identity in store which has identity initialised already")]
+    IdentityAlreadySet,
```
zvolin: is it used anywhere? from the method description I have no idea when would it be returned

node/src/node/builder.rs
```diff
@@ -51,6 +51,18 @@ pub enum NodeBuilderError {
...
     #[error("Could not resolve any of the bootnode addresses")]
     FailedResolvingBootnodes,
+
+    /// Error propagated from keyring
+    #[error(transparent)]
+    KeyringError(#[from] keyring::Error),
+
+    /// Error decoding libp2p identity from the store
+    #[error(transparent)]
+    DecodingError(#[from] libp2p::identity::DecodingError),
```
zvolin: since this is specifically used for identity decoding, I think we should reflect that in name
```suggestion
    /// Error decoding libp2p identity from the store
    #[error(transparent)]
    IdentityDecodingError(#[from] libp2p::identity::DecodingError),
```

node/Cargo.toml
```diff
@@ -56,6 +56,7 @@ tokio-util = "0.7.11"
 tracing = "0.1.40"
 void = "1.0.2"
 uniffi = { version = "0.29.2", optional = true }
+keyring = "3.6.3"
```
zvolin: is it used anywhere?

node/src/node/builder.rs
```diff
@@ -51,6 +51,18 @@ pub enum NodeBuilderError {
     /// Builder failed to resolve dnsaddr multiaddresses for bootnodes
     #[error("Could not resolve any of the bootnode addresses")]
     FailedResolvingBootnodes,
+
+    /// Error propagated from keyring
+    #[error(transparent)]
+    KeyringError(#[from] keyring::Error),
```
zvolin: is it used anywhere?

node/src/store.rs
```diff
@@ -155,6 +156,12 @@ pub trait Store: Send + Sync + Debug {
 
     /// Close store.
     async fn close(self) -> Result<()>;
+
+    /// Handle the libp2p identity initialisation for the store. If `requested_keypair` is
+    /// provided it takes precedence over key in the store and permanently overwrites it.
+    /// If there's no requested keypair, store will either use currently persisted one, or
+    /// generate and store a new one.
+    async fn init_identity(&self, requested_keypair: Option<Keypair>) -> Result<Keypair>;
```
zvolin: this name and api feels a bit weird, it's kind of like `get_or_insert` but not quite :p but I cannot think of how to improve it :thinking:

node/src/store/indexed_db_store.rs
```diff
@@ -310,6 +311,14 @@ impl IndexedDbStore {
...
         .await
     }
+
+    async fn init_identity(&self, requested_keypair: Option<Keypair>) -> Result<Keypair> {
+        // IndexedDbStore cannot be responsible for managing the libp2p identity keys
+        // as we need to create separate indexeddb instance for each identity. As such
+        // we're expecting that the key management has already been done in the wasm glue
+        // code and the key is present here.
+        Ok(requested_keypair.expect("keypair should already be selected here"))
+    }
```
oblique: If I remember correct, some of our users are using Rust crate directly for wasm. Does this mean they always need to use `.keypair()` from the `NodeBuilder`? I think we should write clearly what they need to do.

node/src/store.rs
zvolin: what about we have two separate methods, `get_identity` and `set_identity`? if user provides us the identity, we just call `set_identity` before we call `get_identity`. if identity was empty and we call get, it initializes it with a random one

node/src/store/indexed_db_store.rs
zvolin: or maybe we should lift the solution from `node-wasm` here?

node/Cargo.toml
fl0rek: Removed 👍

node/src/node/builder.rs
fl0rek: Removed 👍

node/src/store.rs
fl0rek: Removed, no longer applicable with set/get

node/src/store.rs
fl0rek: Reworked into set/get

node/src/store/indexed_db_store.rs
fl0rek: Moving the key registry here won't solve the underlying issue - we need a key selection schema outside of the store for wasm but also want a sensible common interface for every other platform. Additionally as Yiannis mentioned, we also have rust on wasm vs js on wasm split.  As such I went with the following:
On all platforms, store sets/gets the key. For redb, this actually persist the key on disk. For indexeddb and in-memory this is in memory only, for wasm because there's a separate key election mechanism in node-wasm (this is triggered on node setup in `node-wasm/src/client.rs`
This makes web target behave more seamlessly from js (no setup), but still allow customising the store when used as rust crate. In that case users could use key registry for more complex identity mgmt.

zvolin: I think I'm fine with this solution, but could you expose the `KeyRegistry` then from `node-wasm` so that users don't need to re-implement it? I'm also wondering if we shouldn't mention it somewhere in the `lumina-node` docs, like for wasm be aware of its existance

zvolin: maybe on `IndexedDbStore`

zvolin: alternatively we could just move key manager module to node and make it pub there, so we don't need users to depend on node-wasm at all

zvolin: actually I have new idea, let's move the key manager to the node, in idb module, and then add methods
`IndexedDbStore::with_random_exclusive_p2p_identity(store_name_prefix: &str) -> (Self, KeyGuard)`
`IndexedDbStore::try_with_exclusive_p2p_identity(store_name_prefix: &str, keypair: Keypair) -> Result<(Self, KeyGuard)>`

We could only expose `KeyGuard` then (I think it's more descriptive name in such case) and document what that's about there

zvolin: those method names already get long, but maybe sneak in `persisted` here too. Also I feel when providing the key, maybe it doesn't need to be put in key manager then, just create a store with an address?

node/src/store.rs
```diff
@@ -153,6 +154,12 @@ pub trait Store: Send + Sync + Debug {
     /// Remove header with given height from the store.
     async fn remove_height(&self, height: u64) -> Result<()>;
 
+    /// Set libp2p identity keypair for the store
+    async fn set_identity(&self, requested_keypair: Keypair) -> Result<()>;
```
zvolin: I think this should be `&mut` so that it can only be set before passing store to lumina

fl0rek: folded setting the key into the store as discussed on daily.

node/src/store/indexed_db_store.rs
fl0rek: updated the entire thing after our discussions on daily

examples/ios/LuminaDemo/LuminaDemoTests/IdentityRetention.swift
```diff
@@ -0,0 +1,39 @@
...
+        let peerId0 = try await node0.localPeerId();
+        let _ = try await node0.stop();
+        
+        let node1 = try LuminaNode(config: config);
+        let _ = try await node1.start();
+        let peerId1 = try await node1.localPeerId();
+        let _ = try await node1.stop();
+        
+        assert(peerId0 == peerId1)
+
```
zvolin:
```suggestion
```

node/src/store/indexed_db_store.rs
```diff
@@ -55,12 +60,20 @@ pub struct IndexedDbStore {
...
+    _db_lock: NamedLock,
 }
 
 impl IndexedDbStore {
     /// Create or open a persistent store.
+    ///
+    /// IndexedDB is shared between all tabs from the same origin, but we enforce
+    /// that each store is opened exclusively. Subsequent open requests with the same name,
+    /// presumably coming from multiple tabs running Lumina, will each create extra stores
+    /// with `extra_tab_N` suffix.
```
oblique: I think we should have a `# Warning` above this. Wdyt?

examples/ios/LuminaDemo/LuminaDemoTests/GrpcClientTest.swift
```diff
@@ -13,34 +13,26 @@ import Testing
...
 
-        switch account {
-        case .base(let baseAccount):
-            let addressStr = try AddressObject.create(address: baseAccount.address).asString()
-            
-            assert(addressStr == NODE_0_ADDR)
-        default :
-            Issue.record("failed")
-        }
+        // TODO: test
```
oblique: Is this todo for this PR or another one?

node/src/store/indexed_db_store.rs
```diff
@@ -579,6 +616,59 @@ async fn detect_schema_version(db: &Rexie) -> Result<Option<u32>> {
...
+
+    let mut try_name = name.to_string();
+    let mut i = 0;
+    loop {
+        match NamedLock::try_lock(&try_name).await {
+            Ok(lock) => return Ok((lock, try_name)),
+            Err(NamedLockError::WouldBlock) => (),
+            Err(e) => return Err(e.into()),
+        }
+        try_name = format!("{name}-extra_tab_{i}");
```
zvolin: nit, could we start from `i = 1` and use `-` instead of `_`?

zvolin (approved):
nice

node/src/store/indexed_db_store.rs
```diff
@@ -55,12 +60,20 @@ pub struct IndexedDbStore {
     head: SendWrapper<RefCell<Option<ExtendedHeader>>>,
     db: SendWrapper<Rexie>,
     header_added_notifier: Notify,
+    //  locks the db instance using Web Locks API until `IndexedDbStore` instance is dropped
```
oblique:
```suggestion
    // locks the db instance using Web Locks API until `IndexedDbStore` instance is dropped
```

node/src/store.rs
```diff
@@ -1208,6 +1234,23 @@ mod tests {
...
+        #[future(awt)]
+        s: S,
+    ) {
+        let store = s;
+        let generated_keypair = store.get_identity().await.unwrap();
+        let persisted_keypair = store.get_identity().await.unwrap();
+
+        assert_eq!(generated_keypair.public(), persisted_keypair.public());
+    }
+
```
oblique, line 1253: I don't feel this checks the persistance. We should manually open/close/open-again the store. We can exclude in memory store.

node/src/store/in_memory_store.rs
```diff
@@ -130,18 +134,24 @@ impl InMemoryStore {
...
     }
 
-    /// Clone the store and all its contents. Async fn due to internal use of async mutex.
+    /// Clone the store and all its contents, except for libp2p identity, which is re-generated.
+    /// Async fn due to internal use of async mutex.
     pub async fn async_clone(&self) -> Self {
         InMemoryStore {
             inner: RwLock::new(self.inner.read().await.clone()),
             header_added_notifier: Notify::new(),
+            libp2p_identity: Keypair::generate_ed25519(),
```
oblique, line 143: Add a unit test that makes sure the key is different after async_clone. Just in case if we forget about it in a future refactor.

node/src/store/redb_store.rs
```diff
@@ -122,6 +124,15 @@ impl RedbStore {
                 let _headers_table = tx.open_table(HEADERS_TABLE)?;
                 let _ranges_table = tx.open_table(RANGES_TABLE)?;
                 let _sampling_table = tx.open_table(SAMPLING_METADATA_TABLE)?;
+                let mut identity_table = tx.open_table(LIBP2P_IDENTITY_TABLE)?;
+                if identity_table.is_empty()? {
```
oblique, line 129:
```suggestion
                let mut identity_table = tx.open_table(LIBP2P_IDENTITY_TABLE)?;

                if identity_table.is_empty()? {
```

node/src/store/redb_store.rs
```diff
@@ -35,6 +35,8 @@ const SCHEMA_VERSION_TABLE: TableDefinition<'static, (), u64> =
     TableDefinition::new("STORE.SCHEMA_VERSION");
 const RANGES_TABLE: TableDefinition<'static, &str, Vec<(u64, u64)>> =
     TableDefinition::new("STORE.RANGES");
+const LIBP2P_IDENTITY_TABLE: TableDefinition<'static, &str, &[u8]> =
```
oblique: I prefer something like `TableDefinition<'static, (), &[u8]>` and store the private part only

node/src/store/indexed_db_store.rs
zvolin:
```suggestion
    /// with `-extra-tab-N` suffix.
```

node/src/store.rs
fl0rek: Added tests for indexeddb and reddb 👍

examples/ios/LuminaDemo/LuminaDemoTests/GrpcClientTest.swift
fl0rek: Todo was about bringing swift tests to a working state, but settled on making them compile only for now.

node/src/store/in_memory_store.rs
fl0rek: Added 👍

node/src/store/redb_store.rs
fl0rek: Not feeling too stronly about it, so removed the peer id key 👍

node/src/store/indexed_db_store.rs
fl0rek: added a warning class, for reference it'll look like this in doc.rs
Screenshot 2025-10-06 at 7 22 51 PM

node/src/store/redb_store.rs
```diff
@@ -486,6 +498,19 @@ impl RedbStore {
         })
         .await
     }
+
+    async fn get_identity(&self) -> Result<Keypair> {
+        self.read_tx(move |tx| {
+            let identity_table = tx.open_table(LIBP2P_IDENTITY_TABLE)?;
+
+            let (_peer_id, key_bytes) = identity_table
```
oblique:
```suggestion
            let (_, key_bytes) = identity_table
```

oblique (approved):

PR #755, opened 2025-09-23
feat(grpc,client): Allow creating celestia-client with read-only grpc

### LINKED ISSUE #708: add support for balance queries in read only mode for celestia-client
I don't like that balance queries are not supported in read only mode, but let's skip that for now while we agree on solution

_Originally posted by @zvolin in https://github.com/eigerco/lumina/pull/682#discussion_r2231380465_

### LINKED ISSUE #745: Allow creating celestia-client without signer, but with gRPC
In that case, current error we throw at construction, should be deferred to the moment user calls function that requires signing.

fl0rek:
closes #745
closes #708

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -30,7 +30,7 @@ pub(crate) struct Context {
 /// A high-level client for interacting with a Celestia node.
 ///
 /// There are two modes: read-only mode and submit mode. Read-only mode requires
-/// RPC endpoint, while submit mode requires RPC and gRPC endpoints plus a signer.
+/// RPC and optionally gRPC endpoint, while submit mode requires both, plus a signer.
 ///
 /// # Examples
 ///
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1,5 +1,5 @@
 use std::error::Error as StdError;
-use std::fmt::Debug;
+use std::fmt::{self, Debug};
 use std::sync::Arc;
 
 use blockstore::cond_send::CondSend;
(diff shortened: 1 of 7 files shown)


client/src/state.rs
```diff
@@ -527,13 +537,13 @@ mod tests {
...
+        let balance = client_ro
             .state()
             .balance_for_address_unverified(&addr)
             .await
-            .unwrap_err();
-        assert!(matches!(e, Error::ReadOnlyMode));
+            .unwrap();
+        assert!(balance > 0);
 
         // Read only mode does not allow calling `balance`
```
fl0rek, line 550: Update error, docs for fn

client/src/client.rs
```diff
@@ -306,3 +294,29 @@ impl ClientBuilder {
         })
     }
 }
+
+impl Debug for Client {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("Client { .. }") // TODO: should we print url/pubkey here?
```
zvolin: not really, as user can just put custom transport

client/src/state.rs
```diff
@@ -39,7 +40,8 @@ impl StateApi {
         self.balance_for_address(&address).await
     }
 
-    /// Retrieves the Celestia coin balance for the signer.
+    /// Retrieves the Celestia coin balance for the signer.  To query balance without
```
zvolin, line 43:
```suggestion
    /// Retrieves the Celestia coin balance for the signer. To query balance without
```

grpc/src/client.rs
```diff
@@ -439,31 +451,27 @@ impl GrpcClient {
...
-            )
+            .calculate_transaction_gas_params(&pfb, &cfg, chain_id.clone(), &account)
             .await?;
 
         let fee = (gas_limit as f64 * gas_price).ceil() as u64;
         let tx = sign_tx(
             pfb,
-            account.chain_id.clone(),
-            &account.account,
+            chain_id.clone(),
```
zvolin, line 473:
```suggestion
            chain_id,
```

grpc/src/error.rs
```diff
@@ -80,13 +80,9 @@ pub enum Error {
...
 
-    /// No account
-    #[error("no account")]
-    NoAccount,
-
-    /// Client was not constructed with a signer
-    #[error("Client was not constructed with a signer")]
-    MissingSinger,
+    /// Client was constructed with without a signer
+    #[error("Client was constructed with without a signer")]
```
zvolin, line 84:
```suggestion
    /// Client was constructed without a signer
    #[error("Client was constructed without a signer")]
```

fl0rek: 🤦

oblique (approved):

zvolin (approved):

PR #761, opened 2025-09-29
feat(node)!: Implement peer discovery based on node type

oblique:


node/src/p2p/connection_control.rs
```diff
@@ -13,6 +16,7 @@ use void::Void;
 // TODO: Wrap ConnectionLimits in it and exclude limits from trusted peers
```
zvolin, line 16: I think this todo may no longer hold

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,708 @@
+use std::collections::HashMap;
+use std::fmt::Debug;
+use std::sync::LazyLock;
+use std::time::Duration;
+
+use futures::StreamExt;
+use libp2p::{
```
zvolin: split those please

node/src/p2p.rs
```diff
@@ -30,21 +30,11 @@ use celestia_types::sample::{Sample, SampleId};
 use celestia_types::{Blob, ExtendedHeader, FraudProof};
 use cid::Cid;
 use futures::stream::FuturesOrdered;
-use futures::{StreamExt, TryStreamExt};
-use libp2p::core::transport::ListenerId;
+use futures::TryStreamExt;
 use libp2p::{
```
zvolin: can you split the imports?

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,708 @@
...
+            addrs.dedup();
+            addrs.shrink_to_fit();
+
+            // Bootnodes are always trusted and protected
+            peer_tracker.set_trusted(peer_id, true);
+            peer_tracker.protect(peer_id, BOOTNODE_PROTECT_TAG);
+        }
+
+        let peer_tracker_info_watcher = peer_tracker.info_watcher();
+        let kademlia_interval = Interval::new(Duration::from_secs(30)).await;
```
zvolin: could we have const for it aswell?

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,708 @@
...
+    fn bootstrap(&mut self) {
+        self.connect_to_bootnodes();
+
+        // Check if there is bootstrap in progress
+        let bootstrap_query_exists = self
+            .swarm
+            .behaviour_mut()
+            .kademlia
+            .iter_queries()
+            .any(|query| matches!(dbg!(query.info()), QueryInfo::Bootstrap { .. }));
```
zvolin:
```suggestion
            .any(|query| matches!(query.info(), QueryInfo::Bootstrap { .. }));
```

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,708 @@
...
+        let protected_full_nodes = self.peer_tracker.protected_len(FULL_PROTECT_TAG);
+        let protected_archival_nodes = self.peer_tracker.protected_len(ARCHIVAL_PROTECT_TAG);
+
+        if info.num_connected_peers == 0 {
+            warn!("All peers disconnected");
+            self.bootstrap();
+        }
+
+        // Based on the spec, when a protected node gets disconnected, the we unprotect
+        // it (check `on_peer_disconnected`), and a another node must be choosen to be
```
zvolin:
```suggestion
        // it (check `on_peer_disconnected`), and another node must be choosen to be
```

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,708 @@
...
+    let local_peer_id = PeerId::from(keypair.public());
+    let store = kad::store::MemoryStore::new(local_peer_id);
+
+    let protocol_id = celestia_protocol_id(network_id, "/kad/1.0.0");
+    let config = kad::Config::new(protocol_id);
+
+    let mut kademlia = kad::Behaviour::with_config(local_peer_id, store, config);
+
+    if !listen_on.is_empty() {
+        kademlia.set_mode(Some(kad::Mode::Server));
```
zvolin, line 665: I think they said we should always use client mode, not sure about that

node/src/peer_tracker.rs
```diff
@@ -1,143 +1,223 @@
...
-            let addr = addr.borrow();
+    pub(crate) fn is_protected(&self, peer_id: &PeerId) -> bool {
+        self.peer(peer_id).is_some_and(|p| p.is_protected())
+    }
 
-            if !state.addrs.contains(addr) {
-                state.addrs.push(addr.to_owned());
-            }
-        }
+    #[allow(dead_code)]
```
zvolin, line 186: it's actually used
```suggestion
```

node/src/peer_tracker.rs
```diff
@@ -148,160 +228,226 @@ impl PeerTracker {
...
         }
 
-        peer_info.connections.push(connection_id);
+        !was_protected
+    }
 
-        // If peer was not already connected from before
-        if !peer_info.is_connected() {
-            peer_info.state = PeerState::Connected;
+    /// Add protect flag to the peer.
```
zvolin:
```suggestion
    /// Remove protect flag from the peer.
```

node/src/peer_tracker.rs
```diff
@@ -148,160 +228,226 @@ impl PeerTracker {
...
+        if peer.protected.remove(&tag) {
+            *self
+                .protect_counter
+                .get_mut(&tag)
+                .expect("protected flag was set but not counted") -= 1;
+
+            info!("Unprotect peer {peer_id} with {tag} tag");
+        }
+
+        // Return true is `protected` state changed
```
zvolin:
```suggestion
        // Return true if `protected` state changed
```

node/src/peer_tracker.rs
```diff
@@ -148,160 +228,226 @@ impl PeerTracker {
...
+            .entry(*peer_id)
+            .or_insert_with(|| Peer::new(*peer_id));
+
+        if !peer.archival {
+            peer.archival = true;
+
+            self.info_tx.send_modify(|tracker_info| {
+                tracker_info.num_connected_archival_nodes += 1;
+            });
+        }
```
zvolin: I think this should happen only if the peer is connected. Here we can be adding peers discovered from get providers, but not yet connected.
If peer is not connected yet, then we should wait for connection and modify tracker info in `increment_connected_peers`

node/src/peer_tracker.rs
```diff
@@ -148,160 +228,226 @@ impl PeerTracker {
...
+                    tracker_info.num_connected_full_nodes -= 1;
+                    true
+                }
+                (false, true) => {
+                    tracker_info.num_connected_full_nodes += 1;
+                    true
+                }
+                _ => false,
+            });
     }
```
zvolin: not sure if this can be called if peer is not connected, mayhaps if we receive identify event and it immediately disconnects? but I'd anyway put a `if peer.connected()` here, as above you do `or_insert_with`

node/src/peer_tracker.rs
```diff
@@ -148,160 +228,226 @@ impl PeerTracker {
...
 
-fn increment_connected_peers(info_tx: &watch::Sender<PeerTrackerInfo>, trusted: bool) {
+fn increment_connected_peers(info_tx: &watch::Sender<PeerTrackerInfo>, peer: &Peer) {
     info_tx.send_modify(|tracker_info| {
         tracker_info.num_connected_peers += 1;
 
-        if trusted {
+        if peer.trusted {
             tracker_info.num_connected_trusted_peers += 1;
         }
```
zvolin: is it possible to know the node kind at this point? wondering if we should also care about `full` there

node/src/peer_tracker.rs
```diff
@@ -333,19 +480,20 @@ mod tests {
     #[test]
     fn trust_after_connect() {
```
zvolin, line 450: I think it'd be nice to have this for archival too, not sure if we can have `full`ness info before connect, if yes, then for it too :p

node/src/peer_tracker.rs
```diff
@@ -1,143 +1,223 @@
...
+
+    pub(crate) fn best_ping(&self) -> Option<Duration> {
+        self.connections
+            .iter()
+            .flat_map(|(_, conn_info)| conn_info.ping)
+            .min()
     }
 }
 
 impl PeerTracker {
```
zvolin, line 149: I'm a bit afraid of the potential to make mistakes there, it seems quite easy to accidentally make a change that will report too many or too little peers of some kind. Maybe we should instead have something like `maybe_update_peer_tracker_info` where we just loop through peers, re-count them and call `notify_if_changed` instead of `+= 1` and `-= 1` everywhere?

zvolin (commented):
first pass, looks clean!

node/src/peer_tracker.rs
oblique: Because archival is discovered via kademlia, we may not have been connected yet, this is the reason I do it like this.

node/src/p2p/swarm_manager.rs
oblique: If we use client mode we would never publish the addresses we are listening

node/src/peer_tracker.rs
oblique: you may be able to mark a full node via kademlia response, but i decided to only use agent version. however i added the addition here for completeness

node/src/peer_tracker.rs
oblique: i was debating with myself about this, iterating all and recounting is not really efficient but it is less error prune.

node/src/peer_tracker.rs
oblique: restricted this method only for connected peers

node/src/peer_tracker.rs
oblique: (also i had a bug in peer tracker)

node/src/peer_tracker.rs
oblique: It is actually unused. Maybe you confused it with `Peer::is_protected_for`

node/src/p2p/connection_control.rs
oblique: Not really. We don't have a mechanism to restrict the number of connections. In celestia-node they have a connection pruner that runs every 1 minute. In our case we don't

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,703 @@
...
+            self.connect(peer_id, None);
+            return;
+        }
+
+        // When Kademlia doesn't know about peer_id, then we need to
+        // initiate a `get_closest_peers` query in order to find it and
+        // their addresses.
+        //
+        // We also avoid calling `get_closest_peers` for the same peer
+        // multiple times.
```
fl0rek:
```suggestion
        // When Kademlia doesn't know peer_id, we need to find it and
        // its addresses using `get_closest_peers` query.

        // First, make sure there isn't already a query in progress for this peer.
```
I'd separate the comment into two, since first part concerns the entire rest of the fn, while second concerns only the block immediately after.

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,703 @@
...
+
+        if !bootstrap_query_exists {
+            if let Err(e) = self.swarm.behaviour_mut().kademlia.bootstrap() {
+                warn!("Can't run kademlia bootstrap: {e}");
+            }
+        }
+    }
+
+    fn start_get_providers_kad_query(&mut self, topic: &RecordKey) {
+        // Avoid having multiple `GetProviders` queries for the same `topic`.
```
fl0rek:
```suggestion
        // Avoid running multiple `GetProviders` queries for the same `topic`.
```
nit, I felt it more to the point here, since it's specificially about concurrently running queries

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,703 @@
...
+    fn protect_best_peers<F>(&mut self, count: usize, tag: u32, condition: F)
+    where
+        F: Fn(&Peer) -> bool,
+    {
+        // Select peers that:
+        //
+        // * Satisfy `condition`
+        // * Are connected
+        // * Not already protected for `tag`
+        // * Have measured ping latency
```
fl0rek:
```suggestion
        // * Have known ping latency
```
above makes it sound as if the peers were doing the measuring ;p

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,703 @@
...
+        kademlia.set_mode(Some(kad::Mode::Server));
+    }
+
+    Ok(kademlia)
+}
+
+/// Converts a topic to `RecordKey`.
+///
+/// This is the equivalent of `nsToCid` that is used in [`RoutingDiscovery.FindPeers`][1]
+/// of go-libp2p which later on is unwraped to the internal hash in [`IpfsDHT.FindProvidersAsync`][2].
```
fl0rek:
```suggestion
/// of go-libp2p which later on is unwraped to internal hash in [`IpfsDHT.FindProvidersAsync`][2].
```
my article senses are tingling

node/src/peer_tracker.rs
```diff
@@ -1,308 +1,422 @@
...
 use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
 use tokio::sync::watch;
+use tracing::info;
 
 use crate::events::{EventPublisher, NodeEvent};
 
+/// How ofter garbage collector should be called.
+pub(crate) const GC_INTERVAL: Duration = Duration::from_secs(30);
+/// How much time a `Peer` needs to be disconnected to get expired.
```
fl0rek:
```suggestion
/// How much time a `Peer` needs to be disconnected to expire.
```

node/src/peer_tracker.rs
```diff
@@ -1,308 +1,422 @@
...
+use lumina_utils::time::Instant;
 use rand::seq::SliceRandom;
 use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
 use tokio::sync::watch;
+use tracing::info;
 
 use crate::events::{EventPublisher, NodeEvent};
 
+/// How ofter garbage collector should be called.
```
fl0rek:
```suggestion
/// How often garbage collector should be called.
```

node/src/peer_tracker.rs
```diff
@@ -1,308 +1,422 @@
...
+    pub(crate) fn is_connected(&self, peer_id: &PeerId) -> bool {
+        self.peer(peer_id).is_some_and(|p| p.is_connected())
+    }
+
+    pub(crate) fn is_protected(&self, peer_id: &PeerId) -> bool {
+        self.peer(peer_id).is_some_and(|p| p.is_protected())
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn is_protected_for(&self, peer_id: &PeerId, tag: u32) -> bool {
```
fl0rek: "Protected for tag" sounds weird to me somehow, but I'm not totally sure. Shouldn't it be "protected with tag"?

node/src/peer_tracker.rs
```diff
@@ -1,308 +1,422 @@
...
-        let mut state = self.get(peer);
+        peer.trusted = is_trusted;
+        self.recount_peer_tracker_info();
+    }
 
-        for addr in addrs {
-            let addr = addr.borrow();
+    /// Add protect flag to the peer.
+    ///
+    /// Tag allows different reasons of protection without interfering with one another.
```
fl0rek:
```suggestion
    /// Tag allows having different reasons for protection without interfering with one another.
```

node/src/peer_tracker.rs
```diff
@@ -1,308 +1,422 @@
...
-        }
+        !was_protected
     }
 
-    /// Sets peer as trusted.
-    pub fn set_trusted(&self, peer: PeerId, is_trusted: bool) {
-        let mut peer_info = self.get(peer);
+    /// Remove protect flag from the peer.
+    ///
+    /// Tag allows different reasons of protection without interfering with one another.
```
fl0rek:
```suggestion
    /// Tag allows having different reasons for protection without interfering with one another.
```

node/src/peer_tracker.rs
```diff
@@ -1,308 +1,422 @@
...
     }
 
-    /// Sets peer as disconnected if `connection_id` was the last connection.
-    ///
-    /// Returns `true` if was set to disconnected.
-    pub fn set_maybe_disconnected(&self, peer: PeerId, connection_id: ConnectionId) -> bool {
-        let mut peer_info = self.get(peer);
-
-        peer_info.connections.retain(|id| *id != connection_id);
+    /// Remove a connection from a peer.
```
fl0rek:
```suggestion
    /// Remove a connection from the peer.
```

node/src/pruner.rs
```diff
@@ -360,7 +360,8 @@ where
         // <-----------Non sampling area----------->|<--Sampling window->
         //
         // This case is straight forward: we just remove whatever is after
```
fl0rek:
```suggestion
        // This case is straightforward: we just remove whatever is after
```
If we're already here, then I guess ;p

node/src/p2p/swarm_manager.rs
```diff
@@ -0,0 +1,703 @@
...
+use tracing::{debug, error, instrument, trace, warn};
+
+use crate::events::{EventPublisher, NodeEvent};
+use crate::p2p::connection_control;
+use crate::p2p::swarm::new_swarm;
+use crate::p2p::Result;
+use crate::peer_tracker::{Peer, PeerTracker, PeerTrackerInfo, GC_INTERVAL};
+use crate::utils::{celestia_protocol_id, MultiaddrExt};
+
+const MAX_PROTECTED_FULL_NODES: usize = 5;
```
fl0rek: I have a slight dislike for the fact that when we say MAX here, we actually mean someting more like DESIRED_AMOUNT. I believe the wording comes from the spec, so it's still probably better to use same terminology as celestia, but could you add a comment here like 
```
// desired number of full node peers, below this threshold node will start searching for new ones
```

node/src/peer_tracker.rs
oblique: Like this: `is_protected_with_tag`?

fl0rek: For the fn name I was thinking just `is_protected`, yours works too 👍

node/src/p2p/swarm_manager.rs
oblique: renamed to `FULL_NODES_PROTECT_LIMIT`

node/src/peer_tracker.rs
oblique: renamed to `is_protected_with_tag`

node/src/p2p/swarm_manager.rs
oblique: renamed to `peer_health_check_interval` and added a const

PR #768, opened 2025-10-03
feat(grpc): resigning and resubmission of transactions

zvolin:


ci/docker-compose.yml
```diff
@@ -93,6 +99,34 @@ services:
   #     - credentials:/credentials
   #     - genesis:/genesis
 
+  # This validator is started as a separate network, only to be able to test
+  # transactions eviction cases. Thus it has low TTL and block size.
+  validator-eviction-testing:
```
fl0rek, line 104: I think validators should be grouped together, but not feeling too strongly about it.

ci/docker-compose.yml
```diff
@@ -93,6 +99,34 @@ services:
...
+    depends_on:
+      validator:
+        condition: service_healthy
+    build:
+      context: .
+      dockerfile: Dockerfile.validator
+    environment:
+      - NODE_NAME=validator-eviction-testing
+      # amount of DA nodes to provision (default: 3)
+      - NODE_COUNT=3
```
fl0rek, line 119: Does it mean that both validators will send the funds?

grpc/src/client.rs
```diff
@@ -58,6 +60,19 @@ pub(crate) struct SignerConfig {
...
+#[derive(Debug)]
+struct BroadcastedTx {
+    /// Broadcasted bytes
+    tx: Vec<u8>,
+
+    /// Transaction hash
+    hash: Hash,
+
+    /// Transaction sequence
+    sequence: u64,
```
fl0rek, line 90:
```suggestion
    /// Broadcasted bytes
    tx: Vec<u8>,
    /// Transaction hash
    hash: Hash,
    /// Transaction sequence
    sequence: u64,
```

grpc/src/error.rs
```diff
@@ -87,6 +93,56 @@ pub enum Error {
...
+            Error::TxBroadcastFailed(_, code, message)
+                if code == &ErrorCode::InvalidSequence || code == &ErrorCode::WrongSequence =>
+            {
+                message.as_str()
+            }
+            _ => return None,
+        };
+
+        Some(extract_sequence(msg))
+    }
```
fl0rek: In this PR they are always used together, essentially checking error type twice and then extracting the new sequence. Maybe they could be merged and then used as 
```rust
if let Some(new_sequence) = event.get_expected_sequence_on_mismatch()? {
...
}
```

grpc/src/client.rs
```diff
@@ -686,20 +759,48 @@ impl GrpcClient {
...
                     }
                 }
-                // node will treat this transaction like if it never happened, so
-                // we need to revert the account's sequence to the one of evicted tx.
-                // all transactions that were already submitted after this one will fail
-                // due to incorrect sequence number.
+                // evicted transaction should be retransmitted without ever trying to re-sign it,
+                // to prevent double spending, because some nodes may have it in a mempool and some not.
+                // if we would re-sign it and both old and new one are confirmed, it'd double spend.
+                // https://github.com/celestiaorg/celestia-app/pull/5783#discussion_r2360546232
```
```suggestion
                // Evicted transaction should be retransmitted without ever trying to re-sign it
                // to prevent double spending, because some nodes may have it in a mempool and some not.
                // If we would re-sign it and both old and new one are confirmed, it'd double spend.
                // https://github.com/celestiaorg/celestia-app/pull/5783#discussion_r2360546232
```

grpc/src/client.rs
```diff
@@ -686,20 +759,48 @@ impl GrpcClient {
...
-                    let mut acc = self.inner.account.lock().await;
-                    acc.as_mut().expect("account data present").sequence = sequence;
-                    return Err(Error::TxEvicted(hash));
+                    if self
+                        .broadcast_tx_with_cfg(tx.clone(), &cfg, context)
+                        .await
+                        .is_err()
+                    {
+                        // Go version of TxClient starts one minute timeout before it returns an eviction
+                        // error, to not make the user re-submitting failed transaction right away.
```
```suggestion
                        // Go version of TxClient starts a one minute timeout before it returns an eviction
                        // error, to avoid user re-submitting the failed transaction right away.
```

grpc/src/client.rs
```diff
@@ -686,20 +759,48 @@ impl GrpcClient {
...
+                        //
+                        // The reason for it is that Go client supports multi-endpoints, sending tx
```
```suggestion
                        // The reason for it is that Go client supports multi-endpoint, sending tx
```
```suggestion
                        // The reason for it is that Go client supports multiple endpoints, sending tx
```

grpc/src/client.rs
```diff
@@ -686,20 +759,48 @@ impl GrpcClient {
...
+                        // to multiple consensus nodes at once, each node having its own mempool.
+                        // One of them could evict the tx from a mempool, but the tx may still be
```
```suggestion
                        // One of them could evict the tx from the mempool, but the tx may still be
```

grpc/src/client.rs
```diff
@@ -686,20 +759,48 @@ impl GrpcClient {
...
+                        // proposed and confirmed by other nodes which didn't evict it.
+                        //
+                        // If client would return eviction error there right away, the user could
+                        // see the error, eagerly resubmit transaction, and have both old and new
+                        // one succeeding, resulting in double spending.
```
```suggestion
                        // In that case, if client would return eviction error right away, the user could
                        // see the error, eagerly resubmit the transaction, and have both old and new
                        // one succeed, resulting in double spending.
```

grpc/src/client.rs
```diff
@@ -686,20 +759,48 @@ impl GrpcClient {
...
+                        //
+                        // Since we don't support multi-endpoints in our client yet, we can just return
+                        // an error right away, as no other consensus node will try to include it.
```
```suggestion
                        // Since we don't support multi-endpoint in our client yet, we can just return
                        // an error right away, as no other consensus node will try to include the transaction.
```

grpc/src/client.rs
```diff
@@ -686,20 +759,48 @@ impl GrpcClient {
...
+                        return Err(Error::TxEvicted(hash));
+                    }
+                }
+                // we need to revert the account's sequence to the one of rejected tx.
```
```suggestion
                // we need to revert the account sequence to that of the rejected tx.
```

grpc/src/error.rs
```diff
@@ -87,6 +93,56 @@ pub enum Error {
     /// Error related to the metadata
     #[error(transparent)]
     Metadata(#[from] MetadataError),
+
+    /// Couldn't parse expected sequence from error message
```
fl0rek:
```suggestion
    /// Couldn't parse expected sequence from the error message
```

ci/docker-compose.yml
zvolin: I moved it to the bottom because it's completely separate network, but no strong feelings too

ci/docker-compose.yml
zvolin: yes, because the networks are separated, accounts need funding on both

grpc/src/error.rs
zvolin: yeah I was thinking that too, but, I couldn't find the nice name to express it

zvolin: switched [to a helper fn](https://github.com/eigerco/lumina/pull/768/commits/a243f999ef6b2242cd20c36f58890df2a0f2713e), wdyt?

grpc/src/test_utils.rs
```diff
@@ -10,6 +10,7 @@ pub use imp::*;
 pub struct TestAccount {
     /// Bech32 `AccountId` of this account
     pub address: Address,
+    #[allow(dead_code)]
```
oblique: Move this below the doc comment

grpc/src/client.rs
```diff
@@ -519,125 +554,177 @@ impl GrpcClient {
...
+            }
+            .encode_to_vec();
+
+            let res = self
+                .broadcast_tx_with_account(blob_tx, &cfg, &mut account, context)
+                .await;
+
+            if let Some(new_sequence) = res.as_ref().err().and_then(extract_sequence_on_mismatch) {
+                account.base.sequence = new_sequence?;
+                continue;
```
oblique, line 657: This goes back to `sign_tx`, shouldn't we recalculate gas?

grpc/src/client.rs
```diff
@@ -519,125 +554,177 @@ impl GrpcClient {
...
+            )
+            .await?;
+
+            let res = self
+                .broadcast_tx_with_account(tx.encode_to_vec(), &cfg, &mut account, context)
+                .await;
+
+            if let Some(new_sequence) = res.as_ref().err().and_then(extract_sequence_on_mismatch) {
+                account.base.sequence = new_sequence?;
+                continue;
```
oblique, line 716: Some as the above comment

grpc/src/client.rs
```diff
@@ -686,20 +776,48 @@ impl GrpcClient {
...
+                        //
+                        // Since we don't support multi-endpoint in our client yet, we can just return
+                        // an error right away, as no other consensus node will try to include the transaction.
+                        return Err(Error::TxEvicted(hash));
+                    }
+                }
+                // we need to revert the account sequence to that of the rejected tx.
+                TxStatus::Rejected => {
+                    let mut acc = self.lock_account(context).await?;
+                    acc.base.sequence = sequence;
```
oblique: Is the following possible?

1. User starts 3 submissions
2. They got signed and broadcasted sequencally, lets say their sequence numbers are: 101, 102, 103
3. On the confirm_tx, we get the replies with this order: 102, 101, 103
4. The `base.sequence` becames 103, but in reality it should be the lowest evicted.

ci/docker-compose.yml
fl0rek: thanks! shared secret keys and separate networks threw me off initially. Would you mind adding small comment like 
> This creates a separate private network, with the same accounts provisioned
or something to that effect

grpc/src/error.rs
fl0rek: lgtm!

grpc/src/client.rs
zvolin: So - in theory, we may hit this error and retry when new block is produced, meaning the gas prices could change. But the estimation service takes median from a few consecutive blocks, so the changes to gas price should be fairly amortized by that.

Celestia-app doesn't do it, it just [recursively calls the broadcast of transaction](https://github.com/celestiaorg/celestia-app/blob/main/pkg/user/tx_client.go#L550). Although in our code it would look quite nice to do this, we could remove the loop from `calculate_transaction_gas_params` and react to sequence errors only within the `broadcast_xyz` methods.
wdyt? imo it sounds good

grpc/src/client.rs
zvolin: yes, that would be the case here. I asked that question on the [client spec PR](https://github.com/celestiaorg/celestia-app/pull/5939#discussion_r2405551671) but it doesn't sound like too much of an issue for celestia. Maybe I should re-rise this point to them.

As for us, I switched to only roll back sequence on txs rejected for reasons other than sequence, that should ensure we always have the lowest

grpc/src/client.rs
zvolin: went that route

PR #772, opened 2025-10-15
feat(node,node-wasm,node-uniffi): Add header, blob, and namespace subscriptions

fl0rek:
Add a way to subscribe to newly incoming headers/blobs/shares from namespace. 
Includes necessary rework of the wasm worker client-server, to support establishing a subscription over a js port.

### DIFF
--- a/node-wasm/src/ports.rs
+++ b/node-wasm/src/ports.rs
@@ -50,7 +50,7 @@ impl TryFrom<MessageEvent> for MultiplexMessage<Command> {
         let MultiplexMessage::<Command> { id, mut payload } =
             from_value(ev.data()).context("could not deserialize message")?;
 
-        if let Some(port) = ev.get_ports().into_iter().take(1).last()
+        if let Some(port) = ev.get_ports().into_iter().next()
             && let Command::Management(WorkerCommand::ConnectPort(maybe_port)) = &mut payload
         {
             let _ = maybe_port.insert(port);


Cargo.toml
```diff
@@ -76,6 +74,7 @@ tendermint-proto = "0.40.3"
 time = { version = "0.3.35", default-features = false }
 tokio = "1.42"
 tokio-util = "0.7.1"
+tokio-stream = "^0.1.16"
```
zvolin:
```suggestion
tokio-stream = "0.1.16"
```

Cargo.toml
```diff
@@ -22,9 +22,7 @@ beetswap = "0.5"
 blockstore = "0.8"
 leopard-codec = "0.2"
 lumina-node = { version = "0.16.0", path = "node" }
-lumina-node-wasm = { version = "0.11.0", path = "node-wasm" }
 lumina-utils = { version = "0.4.0", path = "utils" }
-celestia-client = { version = "0.2.0", path = "client" }
```
zvolin, line 27: why were those removed?

node-uniffi/Cargo.toml
```diff
@@ -25,13 +25,15 @@ celestia-proto = { workspace = true, features = ["uniffi"] }
...
 async-trait.workspace = true
+futures.workspace = true
 k256.workspace = true
 libp2p.workspace = true
 redb.workspace = true
 serde_json.workspace = true
 tendermint.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true, features = ["macros", "sync"] }
+tokio-stream = { workspace = true }
```
zvolin:
```suggestion
tokio-stream.workspace = true
```

node-wasm/Cargo.toml
```diff
@@ -8,7 +8,7 @@ authors = ["Eiger <hello@eiger.co>"]
 homepage = "https://www.eiger.co"
 repository = "https://github.com/eigerco/lumina"
 readme = "README.md"
-rust-version = "1.87"
+rust-version = "1.89"
```
zvolin: this is breaking and would be good to be done separately

node-wasm/Cargo.toml
```diff
@@ -47,7 +47,7 @@ send_wrapper.workspace = true
 thiserror.workspace = true
 time = { workspace = true, features = ["wasm-bindgen"] }
 tokio = { workspace = true, features = ["sync"] }
-tokio-stream = { version = "0.1.16", features = ["sync"] }
+tokio-stream = { workspace = true }
```
zvolin:
```suggestion
tokio-stream.workspace = true
```

node/src/node.rs
```diff
@@ -498,6 +500,97 @@ where
...
+        let store = self
+            .store
+            .as_ref()
+            .cloned()
+            .expect("store should be present");
+        let p2p = self.p2p.as_ref().cloned().expect("p2p should be present");
+
+        let mut prev_head = store.head_height().await?;
+
+        let (tx, rx) = mpsc::channel(16);
```
zvolin, line 534: could you drop a comment with an explanation on why we want to store some amount of items rather than just returning them one by one as we go? Iirc the reason was to buffer some in case someone uses very low or even 0 pruning window, so that it's not returning only errors if it's not awaited quick enough

node/src/node.rs
```diff
@@ -498,6 +500,97 @@ where
...
+                            source: NodeError::P2p(e),
+                        });
+
+                    if tx.send(blobs_or_error).await.is_err() {
+                        return; // receiver dropped
+                    }
+                }
+                prev_head = new_head;
+            }
+        });
```
zvolin: do we want spawning? I'm wondering if we shouldn't just return a `try_stream!` there and avoid background tasks. We could mimic task's behavior to some point by having an `ArrayVec<EH, 16>`, and filling it as much as we can before returning next elem on each await

node/src/node.rs
```diff
@@ -498,6 +500,97 @@ where
...
+        Ok(ReceiverStream::new(rx))
+    }
+
+    /// Return a stream which will yield all the blobs from the namespace, as the new headers
+    /// are being received by the node starting from the first header received after the call.
+    /// Stream is guaranteed to return all blobs (possibly zero) or error for each height, in order.
+    pub async fn namespace_subscribe(
+        &self,
+        namespace: Namespace,
+    ) -> Result<ReceiverStream<Result<(u64, Vec<Blob>), SubscriptionError>>> {
```
zvolin: should we move `BlobsAtHeight` to `celestia-types` and reuse it everywhere?

node/src/node.rs
```diff
@@ -498,6 +500,97 @@ where
...
+            }
+        });
+
+        Ok(ReceiverStream::new(rx))
+    }
+
+    /// Return a stream which will yield all the blobs from the namespace, as the new headers
+    /// are being received by the node starting from the first header received after the call.
+    /// Stream is guaranteed to return all blobs (possibly zero) or error for each height, in order.
+    pub async fn namespace_subscribe(
```
zvolin, line 545: I think this should be `blob_subscribe`. We could also have `namespace_subscribe` but it should support also reserved namespaces, like `Tx` or `PFB` and return `NamespaceData` perhaps? or `Vec<Vec<Share>>` or `Vec<Share>` :thinking:

node/src/node.rs
```diff
@@ -498,6 +500,97 @@ where
...
+                let new_head = store.wait_new_head().await;
+                for height in (prev_head + 1)..=new_head {
+                    let header_or_error =
+                        store
+                            .get_by_height(height)
+                            .await
+                            .map_err(|e| SubscriptionError {
+                                height,
+                                source: NodeError::Store(e),
+                            });
```
zvolin: I'm wondering if it would be cool of us to not spam with `StoreError::NotFound` in case we hit pruned range. Maybe we could make `SubscriptionError` an enum with a `Lagged` variant, and then on not found we would check what is the next existing range in the store, set current height to it and return a lagged error, wdyt?

zvolin: Also, right now this would also be an issue with the blobs subscription, as we depend on having headers to retrieve them, so maybe we could have a generic stream function that we re-use, like:

```rs
fn create_subscription<I, F, Fut>(store, next: F) -> Pin<Box<dyn Stream<Item = I>> + Send + 'static>
where
    F: FnMut(u64) -> Fut + Sync + 'static,
    Fut: Future<Output = I> + Send + 'static,
```

after writing that tho I started having a feeling it's overkill :p

node-uniffi/src/types/subscriptions.rs
```diff
@@ -0,0 +1,100 @@
...
+use uniffi::Object;
+
+use crate::error::LuminaError;
+
+type HeaderSubscriptionItem = Result<ExtendedHeader, NodeSubscriptionError>;
+type BlobSubscriptionItem = Result<(u64, Vec<Blob>), NodeSubscriptionError>;
+
+#[derive(Object)]
+pub struct HeaderStream {
+    stream: RwLock<ReceiverStream<HeaderSubscriptionItem>>,
```
zvolin: I think we only use `.write()` on lock so those could be mutexes?

node-wasm/src/ports.rs
```diff
@@ -64,6 +53,16 @@ pub(crate) struct Port {
...
+impl MessageId {
+    /// i++
+    pub(crate) fn post_increment(&mut self) -> MessageId {
+        let ret = *self;
+        let (next, _carry) = self.0.overflowing_add(1);
+        self.0 = next;
+        ret
+    }
+}
+
```
zvolin: can we move this just below `struct MessageId`?

node-wasm/src/ports.rs
```diff
@@ -86,70 +85,89 @@ impl Port {
...
         Ok(())
     }
 }
 
+impl From<MessagePort> for MessagePortLike {
+    fn from(value: MessagePort) -> Self {
+        MessagePortLike { obj: value.into() }
+    }
+}
+
```
zvolin: please move this below the `MessagePortLike` def

node-wasm/src/ports.rs
```diff
@@ -86,70 +85,89 @@ impl Port {
...
 impl Drop for Port {
     fn drop(&mut self) {
         unregister_onmessage(&self.port, &self.onmessage)
     }
 }
 
+/// Prepare a port for receiving subscription items
+pub(crate) fn subscription_port<T: DeserializeOwned + 'static>(
```
zvolin: maybe we could have a
```rs
impl Port {
    /// Create a new port that is split into sender and the receiver
    fn new_split<T: DeserializeOwned + 'static>(obj: MessagePortLike) -> Result<(Port, mpsc::UnboundedReceiver<T>)>
}
```

node-wasm/src/ports.rs
```diff
@@ -86,70 +85,89 @@ impl Port {
...
+        tx.send(item)
+            .context("forwarding subscription item failed")?;
+        Ok(())
+    })?;
+
+    Ok((client_port, rx))
+}
+
+/// Prepare a Port that is ready to receive Commands from the client or WorkerResults from the worker
+pub(crate) fn prepare_port<T: PayloadWithTransferable + Serialize + DeserializeOwned + 'static>(
```
zvolin: prepare port sounds a bit vague, maybe something like `new_split_client_worker` :upside_down_face:

node-wasm/src/ports.rs
```diff
@@ -64,6 +53,16 @@ pub(crate) struct Port {
...
+        let ret = *self;
+        let (next, _carry) = self.0.overflowing_add(1);
+        self.0 = next;
+        ret
+    }
+}
+
 impl Port {
     /// Create a new Port out of JS object, registering on message callback.
     /// Minimal duck-type checking is performed using reflection when setting appropriate properties.
```
zvolin: I think it could also take `MessagePortLike` instead of `JsValue`

node-wasm/src/commands.rs
```diff
@@ -67,46 +87,93 @@ pub(crate) enum SingleHeaderQuery {
...
+    }
+}
+
+pub(crate) struct CommandWithResponder {
+    pub command: Command,
+    pub responder: oneshot::Sender<WorkerResult>,
+}
 
-    fn check_variant(self) -> Result<Self::Output, JsError>;
+pub(crate) trait PayloadWithTransferable {
```
zvolin: since we only care about ports there out of all transferables, I'd rename things to reflect that

node-wasm/src/commands.rs
```diff
@@ -16,16 +16,36 @@ use lumina_node::store::SamplingMetadata;
...
 #[derive(Debug, Serialize, Deserialize)]
-pub(crate) enum NodeCommand {
+pub(crate) enum Command {
+    Node(NodeCommand),
+    Management(ManagementCommand),
+    Subscribe(NodeSubscription),
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) enum ManagementCommand {
```
zvolin: how about `WorkerCommand`?

node-wasm/src/commands.rs
```diff
@@ -16,16 +16,36 @@ use lumina_node::store::SamplingMetadata;
...
+    GetEventsChannelName,
+    ConnectPort(#[serde(skip)] Option<MessagePortLike>),
     IsRunning,
     StartNode(WasmNodeConfig),
     StopNode,
-    GetEventsChannelName,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) enum NodeSubscription {
```
zvolin: How about `SubscribeCommand`?

zvolin (commented):
cool stuff there, did first pass :+1:

node/src/node.rs
```diff
@@ -498,6 +500,97 @@ where
...
+        let mut prev_head = store.head_height().await?;
+
+        let (tx, rx) = mpsc::channel(16);
+
+        spawn(async move {
+            loop {
+                let new_head = store.wait_new_head().await;
+                for height in (prev_head + 1)..=new_head {
+                    let blobs_or_error = p2p
+                        .get_all_blobs(namespace, height, None, store.as_ref())
```
zvolin: timeout must not be None, otherwise we will just hang here forever e.g. if header is missing in the store

node-wasm/src/ports.rs
fl0rek: I believe original idea was to differenciate between Objects and Objects that we've checked that they are ports (since types here offer little safety, you can just JsValue.into() MessagePortLike). Here we get JsValue, then check with reflection appropriate properties and declare it port-like by returning MessagePortLike in `register_on_message_callback`.
We weren't that meticulous with upholding this convention, so I guess we can treat MessagePortLike as Object that we heavily wish was a port.

Cargo.toml
fl0rek: No clue, they shouldn't be that's for sure - restored. Maybe runnig min dependencies cheker while editing some files?

node/src/node.rs
fl0rek: Added 👍

node/src/node.rs
fl0rek: per Friday's daily, left the spawn. Looking into having single subscription, so that we have at most one task per subscription regardless of subscribed clients.

node-wasm/Cargo.toml
fl0rek: Reverted 👍

node-uniffi/src/types/subscriptions.rs
fl0rek: Switched to mutex 👍

node-wasm/src/ports.rs
fl0rek: I've added two new methods: `with_multiplex_message_channel` and `with_subscription_channel`, which reflect the two usecases for Port we have - sending and responding to multiplexed messages, which may contain a port being transferred, with all the processing it entails; and sending subscription item from the worker to the client, where all we care is being able to serialize the message.

node-wasm/src/ports.rs
fl0rek: removed https://github.com/eigerco/lumina/pull/772#discussion_r2464846527

node-wasm/src/commands.rs
fl0rek: Got rid of the transferrable verbage, except for ports.rs, where we match the js port api.

node/src/node.rs
```diff
@@ -498,6 +500,99 @@ where
             Err(e) => Err(e.into()),
         }
     }
+
+    /// Return a stream which will yield all the headers, as they are being received by the node,
+    /// starting from the first header received after the call. Stream is guaranteed to return either
+    /// header or error for each height, in order.
```
oblique: This will look ugly on the doc. Split it to "title" and the rest

node-wasm/src/ports.rs
zvolin: Yeah I'm aware of the loose typing here. I meant to keep all the reflection but MessagePortLike is more of a docs gain here

node/src/node.rs
```diff
@@ -498,6 +500,99 @@ where
...
+        namespace: Namespace,
+    ) -> Result<ReceiverStream<Result<(u64, Vec<Blob>), SubscriptionError>>> {
+        let store = self
+            .store
+            .as_ref()
+            .cloned()
+            .expect("store should be present");
+        let p2p = self.p2p.as_ref().cloned().expect("p2p should be present");
+
+        let mut prev_head = store.head_height().await?;
```
oblique: This is wrong in the following edge case:

- Let's say I used lumina node 300 blocks ago and then terminated it.
- I start a new node and I immediately call `header_subscribe`
- I will start getting very old headers because of very old `head_height`

I believe the `prev_head` needs to be initialized based on `wait_new_head`.

For example something like this:

```rust
        spawn(async move {
            let mut prev_head = None;

            loop {
                let new_head = store.wait_new_head().await;

                if prev_head.is_none() {
                    prev_head = Some(new_head - 1);
                }

                for height in prev_head.unwrap() + 1..=new_head {
             // ....
```

node/Cargo.toml
```diff
@@ -56,6 +56,7 @@ smallvec = { version = "1.13.2", features = [
 thiserror.workspace = true
 tokio = { workspace = true, features = ["macros", "sync"] }
 tokio-util.workspace = true
+tokio-stream = { workspace = true }
```
oblique:
```suggestion
tokio-stream.workspace = true
```

node/src/node.rs
fl0rek: Added BlobsAtHeight for native and wasm. 
For uniffi, it's touch more hairy. Both `Share` and `Blob` cannot derive `uniff::Record` (we needs methods on them, Share has array in it, which isn't uniffi compatible), thus both `SharesAtHeight` and `BlobsAtHeight` need to be `uniffi::Object`s. To avoid cloning entire vec, I've used `Vec<Arc<T>>`, which is handled transparently by uniffi, but means we need to have a separate `*AtHeight` structs for uniffi.

node/src/node.rs
fl0rek: Added const `SHWAP_FETCH_TIMEOUT`, with 5 second timeout.

node-wasm/src/worker_client.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        Ok(WorkerClient {
+            request_tx,
+            _worker_join_handle,
+            _worker_drop_guard: cancellation_token.drop_guard(),
+        })
+    }
+
+    /// Send a `WorkerCommand`, awaiting for a `WorkerResponse`
+    pub async fn worker(&self, command: WorkerCommand) -> Result<WorkerResponse, WorkerError> {
```
oblique: The name doesn't indicate what this method does. What about `worker_exec` and `node_exec` for the `node`?

node-wasm/src/client.rs
```diff
@@ -460,12 +533,14 @@ mod tests {
 
         let client = spawn_connected_node(vec![bridge_ma.to_string()]).await;
 
-        let info = client.network_info().await.unwrap();
+        let info = client.network_info().await.expect("network info");
+        web_sys::console::log_1(&format!("{info:#?}").into());
```
oblique: remove leftover logs (check other lines too)

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,369 @@
...
+    ) -> Result<(), JsValue>;
+}
+
+impl From<MessagePort> for MessagePortLike {
+    fn from(value: MessagePort) -> Self {
+        MessagePortLike { obj: value.into() }
+    }
+}
+
+pub(crate) fn split_port<Tx>(port: MessagePortLike) -> Result<(PortSender<Tx>, RawPortReceier)> {
```
oblique: `Tx` confused me, I thought it was some sender abstraction. I prefer `T`. Change it also in other places.

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,369 @@
...
+impl Stream for RawPortReceier {
+    type Item = MessageEvent;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let event =
+            ready!(this.receiving_channel.poll_recv(cx)).expect("forward channel should not drop");
+
+        if event.data() == JsValue::NULL {
+            return Poll::Ready(None);
```
oblique, line 176: I think we should also call `receiving_channel.close()`.

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,369 @@
...
+        let mut m = MessageId(u32::MAX - 2);
+        assert_eq!(m.post_increment(), MessageId(u32::MAX - 2));
+        assert_eq!(m.post_increment(), MessageId(u32::MAX - 1));
+        assert_eq!(m.post_increment(), MessageId(u32::MAX));
+        assert_eq!(m.post_increment(), MessageId(0));
+        assert_eq!(m.post_increment(), MessageId(1));
+    }
+
+    #[wasm_bindgen_test]
+    async fn port_smoke() {
```
oblique, line 279: Write also a smoke test for closing the one end of the channel

node/src/node.rs
fl0rek: Added separate blob_subscribe and namespace_subscribe

node/src/node.rs
fl0rek: Moved the source of truth about subscription pacing to syncer. `HeightSequencer` is now responsible for publishing heights of new headers received by the node, guaranteeeing that they are in order (potentially waiting for syncer to fill in the gap in case we miss few a and need to fill back). It also waits for the head init, so this should avoid mentioned issue.

node/src/node.rs
fl0rek: done 👍

node-wasm/src/subscriptions.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// error message
+    pub error: String,
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+pub(crate) enum SubscriptionFeedback {
+    /// receiver has read the previous item and is ready for more
+    Ready,
+    /// receiver closed the channel
+    Close,
```
oblique: I didn't find how this is send to the other end. Maybe we should do it when `SubscriptionStreamInner` is dropped.

node-wasm/src/worker_client.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+impl WorkerClient {
+    /// Create a new `WorkerClient` and start a background thread which forwards responses from the js
+    /// callback.
+    pub fn new(port: JsValue) -> Result<WorkerClient> {
+        let cancellation_token = CancellationToken::new();
+        let mut worker = Worker::new(port.into(), cancellation_token.child_token())?;
+
+        let (request_tx, request_rx) = mpsc::unbounded_channel();
```
oblique: Correct me if i'm wrong. I think in this level we can have a bounded channel because sender and receiver are used from async fn

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    pub(crate) fn signal_inserted(&mut self, range: BlockRange) {
+        let Some(last_sent_height) = self.last_sent_height else {
+            // header insertions before head height is initialised
+            // this shouldn't happen, ignore it
+            return;
+        };
+
+        if &last_sent_height > range.end() {
```
oblique: It will be more correct to remove `1..=last_sent_height` and if the `range` is empty, then return

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    loop {
+        let height = match heights_receiver.recv().await {
+            Ok(height) => height,
+            Err(RecvError::Lagged(skipped)) => {
+                let _ = tx.send(Err(SubscriptionError::Lagged(skipped))).await;
+                continue;
+            }
+            Err(RecvError::Closed) => {
+                error!("unexpected heigth stream close");
```
oblique: Not really an unexpected error when `Syncer` is manually stopped. I prefer remove this log to be removed. Do it in other functions too

node/src/syncer.rs
```diff
@@ -1247,6 +1276,99 @@ mod tests {
...
+        // Syncer should request the gap [12, 14]
+        handle_session_batch(&mut p2p_mock, &headers, 12..=15, true).await;
+        assert_syncing(&syncer, &store, &[1..=15], 15).await;
+
+        // header sub receives header at height 19 (gap)
+        p2p_mock.announce_new_head(headers[18].clone());
+        assert_syncing(&syncer, &store, &[1..=15], 19).await;
+
+        println!("5");
+
```
oblique:
```suggestion
```

node/src/syncer.rs
```diff
@@ -1247,6 +1276,99 @@ mod tests {
...
+        // Syncer should request the gap [16, 18]
+        handle_session_batch(&mut p2p_mock, &headers, 16..=19, true).await;
+        assert_syncing(&syncer, &store, &[1..=19], 19).await;
+        println!("6");
```
oblique:
```suggestion
```

node/src/syncer.rs
```diff
@@ -1247,6 +1276,99 @@ mod tests {
...
+                tx.send(height).unwrap();
+            }
+        });
+
+        // header sub receiving header at height 11 (adjacent, should be added immediately)
+        p2p_mock.announce_new_head(headers[10].clone());
+        assert_syncing(&syncer, &store, &[10..=11], 11).await;
+
+        // header sub receiving header at height 15 (gap, should NOT be added immediately)
+        p2p_mock.announce_new_head(headers[14].clone());
```
oblique, line 1295: After creating a gap it would be nice to check if `try_recv` returns `None`

node/src/syncer.rs
```diff
@@ -1247,6 +1276,99 @@ mod tests {
...
+        assert_syncing(&syncer, &store, &[10..=11], 15).await;
+
+        // Syncer requests past header range, should be ignored by the sequencer
+        handle_session_batch(&mut p2p_mock, &headers, 2..=9, true).await;
+        handle_session_batch(&mut p2p_mock, &headers, 1..=1, true).await;
+
+        // Syncer should request the gap [12, 14]
+        handle_session_batch(&mut p2p_mock, &headers, 12..=15, true).await;
+        assert_syncing(&syncer, &store, &[1..=15], 15).await;
```
oblique, line 1306: Consume up to 15 here

node-wasm/src/ports.rs
fl0rek: Added 👍

node-wasm/src/ports.rs
fl0rek: `Tx` here is the transmitted type, as opposed to the receiving which is left raw, so that receiver can map it to their preferred type (mostly due to different handling of ports between command/response and subscriptions)

node-wasm/src/worker_client.rs
fl0rek: Indeed, changed 👍

node/src/node/subscriptions.rs
fl0rek: HeightSequencer got yoinked, resolving

node-wasm/src/subscriptions.rs
fl0rek: Port closing is now handled entirely on the port level, so this enum is unnecessary. Removed.

node-wasm/src/client.rs
fl0rek: grepped and removed all the console logs 👍

node-uniffi/src/types/subscriptions.rs
```diff
@@ -0,0 +1,172 @@
...
+
+#[uniffi::export]
+impl BlobsAtHeight {
+    fn height(&self) -> u64 {
+        self.height
+    }
+
+    fn blobs(self: Arc<Self>) -> Vec<Arc<Blob>> {
+        self.blobs.iter().map(Arc::clone).collect()
+    }
```
zvolin: here and in the `Shares`, couldn't we also have `&self`?

node/src/store.rs
```diff
@@ -338,6 +339,141 @@ fn to_headers_range(bounds: impl RangeBounds<u64>, last_index: u64) -> Result<Ra
...
+        // Clone headers before inserting since we need them for broadcasting
+        let headers_to_broadcast: Vec<ExtendedHeader> = verified_headers.as_ref().to_vec();
+
+        // Insert into the store
+        self.inner.insert(verified_headers).await?;
+
+        // Broadcast each header after successful insertion
+        for header in headers_to_broadcast {
+            // Ignore send errors - if there are no receivers, that's fine
+            let _ = self.broadcast_tx.send(header);
```
zvolin: add a `yield_now` here, because otherwise we can easily cause `Lagged` on the broadcast, even for quick receiveres

node/src/store.rs
```diff
@@ -338,6 +339,141 @@ fn to_headers_range(bounds: impl RangeBounds<u64>, last_index: u64) -> Result<Ra
     Ok(start..=end)
 }
 
+#[derive(Debug)]
+pub(crate) struct Wrapper<S: Store> {
```
zvolin: maybe `SubscribableStore`?

node/src/store.rs
```diff
@@ -338,6 +339,141 @@ fn to_headers_range(bounds: impl RangeBounds<u64>, last_index: u64) -> Result<Ra
...
+        // Convert headers to VerifiedExtendedHeaders to get access to the headers
+        let verified_headers: VerifiedExtendedHeaders = headers.try_into().map_err(|e| {
+            StoreError::InsertionFailed(StoreInsertionError::HeadersVerificationFailed(
+                e.to_string(),
+            ))
+        })?;
+
+        // Clone headers before inserting since we need them for broadcasting
+        let headers_to_broadcast: Vec<ExtendedHeader> = verified_headers.as_ref().to_vec();
+
```
zvolin: we need to keep track of the head height in wrapper, and only clone and broadcast header range that starts with head + 1

node/src/block_ranges.rs
```diff
@@ -436,6 +436,15 @@ impl BlockRanges {
...
 
+    /// Returns the tail (lowest) range and removes it from the ranges
+    pub fn pop_tail_range(&mut self) -> Option<BlockRange> {
+        if self.0.is_empty() {
+            return None;
+        }
+
+        Some(self.0.remove(0))
+    }
+
```
zvolin: this code is no longer needed, no?

node/src/node.rs
```diff
@@ -7,33 +7,38 @@ use std::ops::RangeBounds;
 use std::sync::Arc;
 use std::time::Duration;
 
+use celestia_types::blob::BlobsAtHeight;
```
zvolin: iiuc the convention you used there, this should go to the imports below

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,142 @@
...
+    #[error("Unable to receive subscription item at {height}: {source}")]
+    Node {
+        /// Height of the subscription item
+        height: u64,
+        /// Error that occured
+        #[source]
+        source: NodeError,
+    },
+    /// Receiver lagged too far behind and the subscription will restart from the current head
+    #[error("Subscription item height already pruned from the store, skipping {0} items")]
```
zvolin, line 36: can you reword the error? pruner isn't in the equation right now, no?

node/src/syncer.rs
```diff
@@ -20,7 +20,7 @@ use tracing::{debug, error, info, instrument, warn};
 use crate::block_ranges::{BlockRange, BlockRangeExt, BlockRanges};
 use crate::events::{EventPublisher, NodeEvent};
 use crate::p2p::{P2p, P2pError};
-use crate::store::{Store, StoreError};
+use crate::store::{Store, StoreError, Wrapper};
```
zvolin: you can revert all the changes here as `Wrapper` implements `Store` and syncer doesn't need to be aware of it

fl0rek: I thought of this as a way to enforce the wrapper usage, but we can set it in node, removed :+1:

node-uniffi/src/types/subscriptions.rs
fl0rek: Actually, yes. Fixed.

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,394 @@
...
+}
+
+impl TryFrom<MessageEvent> for MultiplexMessage<WorkerResult> {
+    type Error = Error;
+
+    fn try_from(ev: MessageEvent) -> Result<Self, Self::Error> {
+        let MultiplexMessage::<WorkerResult> { id, mut payload } =
+            from_value(ev.data()).context("could not deserialize message")?;
+
+        if let Some(port) = ev.get_ports().into_iter().take(1).last() {
```
zvolin:
```suggestion
        if let Some(port) = ev.get_ports().into_iter().next() {
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,394 @@
...
+    }
+}
+
+impl<Tx> Drop for PortSender<Tx> {
+    fn drop(&mut self) {
+        // send a close signal and 🤞
+        let close_signal: JsValue = CHANNEL_CLOSE_SYMBOL_NAME.to_string().into();
+        let _ = self
+            .port
+            .post_message_with_transferable(&close_signal, &JsValue::UNDEFINED);
```
zvolin:
```suggestion
            .post_message(&close_signal);
```

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let event =
+            ready!(this.receiving_channel.poll_recv(cx)).expect("forward channel should not drop");
+
+        if event
+            .data()
+            .dyn_ref::<JsString>()
+            .is_some_and(|s| s == CHANNEL_CLOSE_SYMBOL_NAME)
+        {
```
zvolin: I think you could compare this directly if you want, although the current explicit version is also nice, for being explicit :p
```suggestion
        if event.data() == CHANNEL_CLOSE_SYMBOL_NAME {
```

node-wasm/src/subscriptions.rs
```diff
@@ -0,0 +1,250 @@
...
+            JsCast::unchecked_ref(self.next.as_ref().as_ref()),
+            &JsValue::UNDEFINED,
+            &js_sys::Array::new(),
+        )
+        .expect("apply on closure should succeed")
+        .unchecked_into()
+    }
+
+    pub fn return_self(self) -> JsValue {
+        JsValue::from(self.clone())
```
zvolin, line 90: should this method have `&self` receiver? otherwise cloning is not needed, but consuming isn't very javascriptish either

node/src/store.rs
fl0rek: This got moved to syncer, with proper range tracking

node/src/store.rs
fl0rek: Went with BroadcastingStore,wdyt?

node-wasm/src/client.rs
```diff
@@ -3,27 +3,32 @@
 use std::time::Duration;
 
 use blockstore::EitherBlockstore;
-use celestia_types::nmt::Namespace;
-use celestia_types::{Blob, ExtendedHeader};
-use js_sys::Array;
+use celestia_types::blob::BlobsAtHeight;
```
zvolin, line 6: if we want to separate our stuff from other stuff then those 2 belong below

zvolin: although I'm fine with merging all the imports to not need to put such comments on reviews :p

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,394 @@
...
+    let receiver = RawPortReceier {
+        port,
+        receiving_channel,
+        onmessage,
+    };
+    Ok((sender, receiver))
+}
+
+// Sending part of a wrapped MessagePort.
+// WARN: do not use unit structs as a sent type, as they are interpreted as a
```
zvolin: as a what? :p

node-wasm/src/ports.rs
zvolin: let's get rid of the generic here, and have it only on `send`. The transmitted type is anyway `JsValue`

zvolin: we could then have `PortSender` and `PortReceiver`, as they would both be kinda `Raw`

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,394 @@
...
+            .context("could not set onmessage callback")?;
+    } else {
+        return Err(Error::new("Don't know how to register onmessage callback"));
+    }
+
+    Ok(())
+}
+
+/// unregister onmessage callback in any of the various forms it can take
+pub(crate) fn unregister_onmessage(port: &JsValue, callback: &Closure<dyn Fn(MessageEvent)>) {
```
zvolin, line 227: super nit but can we either add `_callback` here or remove it from `register`?

node-wasm/src/subscriptions.rs
```diff
@@ -0,0 +1,250 @@
...
+            value: JsValue::UNDEFINED,
+        }
+    }
+
+    pub fn ready(value: JsValue) -> Self {
+        Self { done: false, value }
+    }
+}
+
+#[wasm_bindgen]
```
zvolin:
```suggestion
#[wasm_bindgen(skip_typescript)]
```

node-wasm/src/subscriptions.rs
```diff
@@ -0,0 +1,250 @@
...
+                    }
+                })?
+            })
+            .boxed_local(),
+    ));
+
+    let async_iterator: AsyncIteratorImpl = Closure::<dyn FnMut() -> Promise>::new(move || {
+        let (receiver, feedback) = (receiver.clone(), feedback.clone());
+        future_to_promise(async move {
+            if let Err(e) = feedback.borrow().send(&SubscriptionReceiverReady, &[]) {
```
zvolin, line 127: there's a comment on the `PortSender` to not use unit structs with it :detective:

node-wasm/src/subscriptions.rs
```diff
@@ -0,0 +1,250 @@
...
+// Wrap provided port into SubscriptionStream and prepare it to be used as
+// js AsyncIterator. Assumes provided port was prepared with [`forward_stream_to_message_port`]
+pub(crate) fn into_async_iterator<S>(port: MessagePortLike) -> Result<AsyncIterator>
+where
+    S: DeserializeOwned + Into<JsValue> + 'static,
+{
+    let (feedback, receiver) = split_port(port)?;
+
+    let feedback = Rc::new(RefCell::new(feedback));
+    let receiver = Rc::new(RwLock::new(
```
zvolin: since we never call `.read()` this could be `Mutex`

node-wasm/src/client.rs
```diff
@@ -335,27 +377,57 @@ impl NodeClient {
...
+        into_async_iterator::<BlobsAtHeight>(port)
+    }
+
+    #[wasm_bindgen(js_name = namespaceSubscribe)]
+    pub async fn namespace_subscribe(&self, namespace: Namespace) -> Result<AsyncIterator> {
+        let command = SubscriptionCommand::Shares(namespace);
+        let port = self.worker.subscribe(command).await?;
+
+        into_async_iterator::<SharesAtHeight>(port)
+    }
```
zvolin: is there a way to type it more than just `AsyncIterator<any>`?

node-wasm/src/ports.rs
fl0rek: I stumbled upon a fix for handling unit structs as I was writing this comment, so I've abandoned it and then didn't go back to remove it :see_no_evil: Removed, as unit structs are not handled correctly

node-wasm/src/subscriptions.rs
fl0rek: Removed the comment as the handling was fixed

node-wasm/src/subscriptions.rs
fl0rek: Tested it with &self and it behaves correctly, previous issues I've had must have been with the previous version of the Iterator here

node-wasm/src/client.rs
fl0rek: merged them into one actually, since this seems to be a re-occurring problem ;p

node-wasm/src/ports.rs
fl0rek: I've added them to bring some type safety to the js land, since in our case it'd be incorrect to send any other message than the one type receiver is expected to receive. Removing since it seems to be bringing more confusion.

node-wasm/src/client.rs
fl0rek: Added unchecked_return_type, which seems to have done the trick. Once I managed typescript support in nvim, it correctly identified as an async iterator that would yield either ExtendedHeader/.. or SubscriptionError error.

node-wasm/src/ports.rs
fl0rek: now both are without the suffix.

node/src/node.rs
```diff
@@ -510,56 +508,37 @@ where
...
-
-        // We're keeping a small buffer of headers, so that new headers are cached eagerly
-        // as they come in, giving consumer more time before they are no longer available.
-        // This is mostly relevant for cases where pruning window is set to zero.
-        let (tx, rx) = mpsc::channel(16);
-
-        spawn(async move { forward_new_headers(syncer, store, tx).await });
-
-        Ok(ReceiverStream::new(rx))
+            .expect("store should be present")
```
oblique:
```suggestion
            .expect("syncer should be present")
```

node/src/store.rs
```diff
@@ -206,7 +206,7 @@ pub enum StoreInsertionError {
     #[error("Provided headers failed to be verified with existing neighbors: {0}")]
     NeighborsVerificationFailed(String),
 
-    /// Store containts are not met.
+    /// Store constants are not met.
```
oblique:
```suggestion
    /// Store contraints are not met.
```

node/src/syncer.rs
```diff
@@ -1330,11 +1322,23 @@ mod tests {
...
+                select! {
+                    header = header_receiver.recv() => {
+                        let header = match header {
+                            Err(RecvError::Closed) => break,
+                            header_or_error => header_or_error.unwrap()
+                        };
+                        tx.send(Some(header)).unwrap();
+                    },
+                    _ = &mut disconnect_signal, if !disconnect_signal.is_terminated() => {
+                        tx.send(None).unwrap();
```
oblique, line 1351: I don't understand what you are trying to achieve

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,266 @@
...
+            .last_sent_height
+            .expect("syncer should have initialised the height by now");
+
+        let Some(lowest_range_height) = range.first().map(|h| h.height().value()) else {
+            // Ignore empty range
+            return Ok(());
+        };
+
+        if lowest_range_height < last_sent_height {
+            // We know range cannot cross last_sent_height, so either both ends are before or after.
```
oblique, line 117: Add a `debug_assert` that checks if `last_sent_height` is within `range`. Add it above this if statement

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,266 @@
...
+        while i < self.pending.len() {
+            if self.pending[i]
+                .first()
+                .expect("header range shouldn't be empty")
+                .height()
+                .value()
+                == self
+                    .last_sent_height
+                    .expect("last_sent_height should be initialised here")
+                    + 1
```
oblique: add `first_pending_height` and `last_sent_height` variables to make this if statement more readable

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,266 @@
...
+            {
+                let range = self.pending.swap_remove(i);
```
oblique, line 142: I prefer a `VecDeque` instead of this. but up to you

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,266 @@
...
+        self.last_sent_height = Some(
+            headers
+                .last()
+                .expect("range shouldn't be empty here")
+                .height()
+                .value(),
+        );
+        for header in headers {
+            let _ = self.sender.send(header); // ok to ignore no receivers
+            yield_now().await;
```
oblique, line 166: Add a comment on why we need yielding here

node-wasm/src/subscriptions.rs
```diff
@@ -0,0 +1,248 @@
...
+        let (tx, rx) = split_port(p0.into()).unwrap();
+        let async_iterator = into_async_iterator::<String>(p1.into());
+        drop(async_iterator);
+
+        tx.send(&"foo".to_string(), &[]).unwrap();
+
+        let mut rx = rx.map(|ev| from_value::<SubscriptionReceiverReady>(ev.data()).unwrap());
+        // At this point receiver would have already signaled its readiness,
+        // so we need to clear the signal from the channel first.
+        //assert_eq!(rx.next().await.unwrap(), SubscriptionReceiverReady);
```
oblique: is this a leftover?

node/src/store.rs
fl0rek: :see_no_evil:

node/src/syncer.rs
fl0rek: This inserts a None at the place where we disconnected, so that we can later see whether headers are properly ordered in relation to it.

node-wasm/src/subscriptions.rs
fl0rek: yup, removed

node/src/node/subscriptions.rs
fl0rek: Added :+1:

node/src/node/subscriptions.rs
fl0rek: Changed :+1:

node/src/node/subscriptions.rs
fl0rek: We assert that the last_sent_height is not in the range being inserted. Added relevant assert.

node-wasm/src/ports.rs
oblique: I don't mind type safety, I just wanted it as `T` instead of `Tx`. But anyway since you remove it, let's rename ` RawPortReceier` to ` PortReceiver`

node/src/syncer.rs
```diff
@@ -1247,6 +1271,119 @@ mod tests {
         assert_syncing(&syncer, &store, &[1..=20], 20).await;
     }
 
+    #[async_test]
+    async fn height_sequencer_smoke() {
```
oblique:
```suggestion
    async fn height_sequence_smoke() {
```

node/src/syncer.rs
```diff
@@ -1247,6 +1271,119 @@ mod tests {
...
+        assert_syncing(&syncer, &store, &[1..=15], 19).await;
+
+        // Syncer should request the gap [16, 18]
+        handle_session_batch(&mut p2p_mock, &headers, 16..=19, true).await;
+        assert_syncing(&syncer, &store, &[1..=19], 19).await;
+
+        for i in 11..=19 {
+            let height = received_heights.recv().await.unwrap();
+            assert_eq!(height, i);
+        }
```
oblique, line 1325: Added more assertions

```suggestion
        // header sub receiving header at height 11 (adjacent, should be added immediately)
        p2p_mock.announce_new_head(headers[10].clone());
        assert_syncing(&syncer, &store, &[10..=11], 11).await;
        assert_eq!(received_heights.recv().await.unwrap(), 11);

        // header sub receiving header at height 15 (gap, should NOT be added immediately)
        p2p_mock.announce_new_head(headers[14].clone());
        assert_syncing(&syncer, &store, &[10..=11], 15).await;
        received_heights.try_recv().unwrap_err();

        // Syncer requests past header range, should be ignored
        handle_session_batch(&mut p2p_mock, &headers, 2..=9, true).await;
        handle_session_batch(&mut p2p_mock, &headers, 1..=1, true).await;
        received_heights.try_recv().unwrap_err();

        // Syncer should request the gap [12, 15]
        handle_session_batch(&mut p2p_mock, &headers, 12..=15, true).await;
        assert_syncing(&syncer, &store, &[1..=15], 15).await;

        for i in 12..=15 {
            let height = received_heights.recv().await.unwrap();
            assert_eq!(height, i);
        }

        // header sub receives header at height 19 (gap)
        p2p_mock.announce_new_head(headers[18].clone());
        assert_syncing(&syncer, &store, &[1..=15], 19).await;
        received_heights.try_recv().unwrap_err();

        // Syncer should request the gap [16, 18]
        handle_session_batch(&mut p2p_mock, &headers, 16..=19, true).await;
        assert_syncing(&syncer, &store, &[1..=19], 19).await;

        for i in 16..=19 {
            let height = received_heights.recv().await.unwrap();
            assert_eq!(height, i);
        }
```

node/src/syncer.rs
```diff
@@ -1247,6 +1271,119 @@ mod tests {
...
+    }
+
+    #[async_test]
+    async fn height_sequencer_handles_disconnect() {
```
oblique:
```suggestion
    async fn height_sequence_handles_disconnect() {
```

node/src/syncer.rs
oblique: Add a comment for it

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,385 @@
...
+    onmessage: Closure<dyn Fn(MessageEvent)>,
+}
+
+impl Stream for PortReceiver {
+    type Item = MessageEvent;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        let event =
+            ready!(this.receiving_channel.poll_recv(cx)).expect("forward channel should not drop");
```
oblique: After the channel is closed `poll_recv` will always return `Poll::Ready(None)`, which will cause panic. Return the `Poll::Ready(None)` in this case.

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,385 @@
...
+        let (p0, p1) = MessageChannel::new_ports().unwrap();
+        let (tx0, _rx0) = split_port(p0.into()).unwrap();
+        let (_tx1, mut rx1) = split_port(p1.into()).unwrap();
+
+        tx0.send(&1u32, &[]).unwrap();
+        drop(tx0);
+
+        let v: u32 = from_value(rx1.next().await.unwrap().data()).unwrap();
+        assert_eq!(v, 1);
+        assert!(rx1.next().await.is_none());
```
oblique, line 307: Add this line one more time in order to check that `None` is still returned

node-wasm/src/subscriptions.rs
```diff
@@ -0,0 +1,245 @@
...
+        let (p0, p1) = MessageChannel::new_ports().unwrap();
+
+        let (tx, rx) = split_port(p0.into()).unwrap();
+        let async_iterator = into_async_iterator::<String>(p1.into());
+        drop(async_iterator);
+
+        tx.send(&"foo".to_string(), &[]).unwrap();
+
+        let mut rx = rx.map(|ev| from_value::<SubscriptionReceiverReady>(ev.data()).unwrap());
+        assert!(rx.next().await.is_none());
```
oblique, line 243: Run it one more time to make sure `next` is still returning `None`

node-wasm/src/worker_server.rs
```diff
@@ -0,0 +1,303 @@
...
+                }
+                res = self.pending_responses_map.next(), if !self.pending_responses_map.is_empty() => {
+                    let response =
+                            &mut res
+                            .ok_or(Error::new("Responses channel closed, should not happen"))?;
+                    let ports: Vec<_> = response.payload.take_port().into_iter().collect();
+                    self.response_sender
+                        .send(response, ports.as_ref())
+                        .with_context(|| format!("failed to send outgoing response for {:?}", response.id))?;
+
```
oblique:
```suggestion
```

node/src/node/subscriptions.rs
```diff
@@ -0,0 +1,273 @@
...
+    last_sent_height: Option<u64>,
+    pending: Vec<Vec<ExtendedHeader>>,
+}
+
+impl<S> BroadcastingStore<S>
+where
+    S: Store,
+{
+    pub fn new(store: Arc<S>) -> Self {
+        let (sender, _) = broadcast::channel(HEIGHT_SEQUENCER_BROADCAST_CAPACITY);
```
oblique: Remove or rename the `sequencer`. Check also other places

node/src/node.rs
```diff
@@ -498,6 +503,71 @@ where
...
+
+    /// Subscribe to new headers received by the node from the network.
+    ///
+    /// Return a stream which will yield all the headers, as they are being received by the
+    /// node, starting from the first header received after the call.
+    pub async fn header_subscribe(&self) -> Result<broadcast::Receiver<ExtendedHeader>> {
+        Ok(self
+            .syncer
+            .as_ref()
+            .expect("syncer should be present")
```
oblique: Actually you can use `self.syncer()`

node/src/store.rs
```diff
@@ -206,7 +206,7 @@ pub enum StoreInsertionError {
     #[error("Provided headers failed to be verified with existing neighbors: {0}")]
     NeighborsVerificationFailed(String),
 
-    /// Store containts are not met.
+    /// Store contraints are not met.
```
oblique: 🙈
```suggestion
    /// Store constraints are not met.
```

Also there is typo in `error` and in variant name

node/tests/subscriptions.rs
```diff
@@ -0,0 +1,72 @@
...
+            break blobs;
+        }
+        assert!(height < submitted_at);
+        assert!(blobs.is_empty());
+    };
+
+    assert_eq!(&blobs, from_ref(&blob));
+
+    let BlobsAtHeight { blobs, .. } = blob_stream.next().await.unwrap().unwrap();
+    assert_eq!(blobs, vec![]);
```
oblique, line 71: So in case of no blobs we always receive an empty vec? Why not receiving nothing?

zvolin: because it indicates that there's new height, that didn't have blobs. We could have `Option<Vec<>>` instead, but we have this api for quite a while already, not sure if worth it

node-wasm/src/ports.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl TryFrom<MessageEvent> for MultiplexMessage<Command> {
+    type Error = Error;
+
+    fn try_from(ev: MessageEvent) -> Result<Self, Self::Error> {
+        let MultiplexMessage::<Command> { id, mut payload } =
+            from_value(ev.data()).context("could not deserialize message")?;
+
+        if let Some(port) = ev.get_ports().into_iter().take(1).last()
```
zvolin:
```suggestion
        if let Some(port) = ev.get_ports().into_iter().next()
```

node/src/node/subscriptions.rs
fl0rek: Renamed to `HEADER_BROADCAST_CHANNEL_CAPACITY`

node/src/node.rs
fl0rek: true, thanks!

node-wasm/src/ports.rs
fl0rek: true, fixed :+1:

node-wasm/src/ports.rs
```diff
@@ -302,6 +305,7 @@ mod tests {
         let v: u32 = from_value(rx1.next().await.unwrap().data()).unwrap();
         assert_eq!(v, 1);
         assert!(rx1.next().await.is_none());
+        assert!(rx1.next().await.is_none());
```
oblique, line 309: Add also a comment on why we do it two times

node/src/store.rs
```diff
@@ -206,7 +206,7 @@ pub enum StoreInsertionError {
     #[error("Provided headers failed to be verified with existing neighbors: {0}")]
     NeighborsVerificationFailed(String),
 
-    /// Store containts are not met.
+    /// Store constraints are not met.
     #[error("Contraints not met: {0}")]
     ContraintsNotMet(BlockRangesError),
```
oblique, line 211: Fix also these two

fl0rek: I've left this on purpose as all the other changes are additive, thus non-breaking. I'll create a separate breaking PR.

node/src/node/subscriptions.rs
zvolin: I don't think `VecDeque` provides any gains here, as the ranges can be out of order. We may have to pick an arbitrary range from the middle of pending to be sent first

types/src/blob.rs
```diff
@@ -67,6 +67,19 @@ pub struct BlobParams {
     pub gov_max_square_size: u64,
 }
 
+/// List of blobs together with height they were published at
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+#[cfg_attr(
+    all(feature = "wasm-bindgen", target_arch = "wasm32"),
+    wasm_bindgen(getter_with_clone, inspectable)
+)]
+pub struct BlobsAtHeight {
```
zvolin, line 76: I think we have a second definition in `rpc` that we could get rid of, and just reexport it from rpc/blob for it to not be a breaking change if you want

node/src/node/subscriptions.rs
zvolin: Actually I changed my mind, this is gonna be nicer with deque

zvolin: Actually I changed my mind, this is gonna be nicer with deque

zvolin: nah, I was thinking we could order the ranges at insertion time with vecdeque push_front / push_back, but it doesn't work in case we loose connection, reconnect, don't have a time to sync the gap and loose connection again. Unless we insert and sort, vecdeque has no gains

types/src/blob.rs
fl0rek: It has additional Option in there I didn't want here. I think it's needed for rpc, nul vs empty vec?

node-wasm/src/ports.rs
fl0rek: Leaving it as is, unless there are objections - PortReceiver and generic send.

node/src/syncer.rs
fl0rek: added comment at the beginning https://github.com/eigerco/lumina/pull/772/files#diff-4d0e726dde3f3ba66c0cd97c8f61c7aa1fc2104ca66e3d04d2bda1e488909930R1340

node/src/node/subscriptions.rs
fl0rek: Unless there are objections, leaving as is.  This will have single digit elements anyway.

node-wasm/src/ports.rs
fl0rek: added https://github.com/eigerco/lumina/pull/772/files#diff-bd5f72d52a37d4192ab2e805db3fc07599386766ef34937a88c4dbb000c12074R308

PR #773, opened 2025-10-15
chore!: Migrate to Rust 2024

oblique:


node/src/daser.rs
```diff
@@ -620,8 +620,8 @@ fn random_indexes(square_width: u16, max_samples_needed: usize) -> HashSet<(u16,
     let mut rng = rand::thread_rng();
 
     while indexes.len() < max_samples_needed {
-        let row = rng.gen::<u16>() % square_width;
-        let col = rng.gen::<u16>() % square_width;
+        let row = rng.r#gen::<u16>() % square_width;
+        let col = rng.r#gen::<u16>() % square_width;
```
oblique, line 624: rand v0.9 deprecates `gen()` and introduces `random()`. However tendermint is using rand v0.8, so I couldn't upgrade rand for the whole workspace.

Should I upgrade to rand v0.9 for the workspace and leave rand v0.8 in celestia-types?

node-uniffi/src/lib.rs
```diff
@@ -70,11 +70,12 @@ impl LuminaNode {
...
-            node.stop().await;
-            Ok(())
-        } else {
-            Err(LuminaError::NodeNotRunning)
+        match node.take() {
+            Some(node) => {
+                node.stop().await;
+                Ok(())
+            }
+            _ => Err(LuminaError::NodeNotRunning),
```
zvolin, line 78: I think `if let` was more elegant, is this change needed?

Cargo.toml
```diff
@@ -79,7 +82,7 @@ tonic-web-wasm-client = "0.7"
 thiserror = "2.0.3"
 tracing = "0.1.41"
 tracing-subscriber = "0.3.15"
-uniffi = "0.29"
+uniffi = "0.29.2"
```
zvolin, line 85: do we need to enforce this patch version?

node/src/daser.rs
zvolin: we kept using the older `rand` to not enforce `RUST_FLAGS=getrandom_backend` on all wasm users, I'm fine with bumping it if `getrandom` released fix already

node/src/network.rs
```diff
@@ -92,12 +92,12 @@ impl Network {
...
             Network::Arabica => "arabica-11",
             Network::Mocha => "mocha-4",
-            Network::Custom(ref s) => &s.id,
+            Network::Custom(s) => &s.id,
         }
     }
 
     /// Get official Celestia and Lumina bootnodes for the given network.
-    pub fn canonical_bootnodes(&self) -> impl Iterator<Item = Multiaddr> {
+    pub fn canonical_bootnodes(&self) -> impl Iterator<Item = Multiaddr> + use<> {
```
zvolin, line 100: does an empty `+ use<>` mean to not capture `&self` lifetime?

node-uniffi/src/lib.rs
oblique: This change was made from `cargo fix` in order to make the drops of older if-let the same for 2024 edition. The behaviour not really changes tho, so I can revert it.

Cargo.toml
oblique: Yes. We have a compilation error on `0.29.0` and `0.29.1`. Minimal versions CI job was not passing.

node/src/network.rs
oblique: Yes and it is equivalent to the 2021 (or older) behaviour. In 2024 all lifetimes are captured unless if `use<..>` is used.

Without this the `cargo check --tests -p lumina-node` gives me the following error:

```
error[E0716]: temporary value dropped while borrowed
   --> node/src/network.rs:218:23
    |
218 |         let private = Network::Custom(id).canonical_bootnodes();
    |                       ^^^^^^^^^^^^^^^^^^^                      - temporary value is freed at the end of this statement
    |                       |
    |                       creates a temporary value which is freed while still in use
219 |         assert_eq!(private.count(), 0);
    |                    ------- borrow later used here
    |
note: this call may capture more lifetimes than intended, because Rust 2024 has adjusted the `impl Trait` lifetime capture rules
   --> node/src/network.rs:218:23
    |
218 |         let private = Network::Custom(id).canonical_bootnodes();
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider using a `let` binding to create a longer lived value
    |
218 ~         let binding = Network::Custom(id);
219 ~         let private = binding.canonical_bootnodes();
    |
help: use the precise capturing `use<...>` syntax to make the captures explicit
    |
100 |     pub fn canonical_bootnodes(&self) -> impl Iterator<Item = Multiaddr> + use<> {
    |                                                                          +++++++

For more information about this error, try `rustc --explain E0716`.
```

PR #777, opened 2025-10-17
fix(ci): Fix iOS hanging

oblique:


.github/workflows/ci.yml
```diff
@@ -206,16 +207,59 @@ jobs:
...
+        xcrun simctl bootstatus "$UDID" -b
+
+        # Set env for the next steps
+        echo "SIM_UDID=$UDID" >> "$GITHUB_ENV"
+
     - name: xcode clean build
       working-directory: examples/ios/LuminaDemo
-      run: xcodebuild -scheme LuminaDemo clean build CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO
+      env:
+        NSUnbufferedIO: "YES"
```
zvolin, line 240: what is this for?

zvolin (approved):

.github/workflows/ci.yml
oblique: It enables unbuffered IO. It causes the output to be displayed as soon as `printf` is used. However the only references I find to it, are from 2017 or older.

oblique: I can remove it if you want, it will not change the result of the fix

zvolin: no hard feelings, was just wondering what's that to see if we want to comment on it why we want it, but I'm good after your explanation

PR #780, opened 2025-10-20
feat(client,rpc)!: support auth in wasm rpc client

zvolin:


client/src/blob.rs
```diff
@@ -159,26 +159,24 @@ impl BlobApi {
...
         let inner = self.inner.clone();
 
+        // we need to re-stream it to map error and satisfy 'static
         try_stream! {
-            let mut subscription = inner.rpc.blob_subscribe(namespace).await?;
-
+            let mut subscription = inner.rpc.blob_subscribe(namespace);
             while let Some(item) = subscription.next().await {
-                let blobs = item?;
-                // TODO: Should we validate blobs?
```
fl0rek, line 180: guessing the answer is "no"? ;p

rpc/src/client.rs
```diff
@@ -163,41 +163,86 @@ mod native {
...
 
-            Ok(Client { client })
+            Ok(Client {
+                id: AtomicU64::new(0),
+                url: conn_str.into(),
+                auth_token: auth_token.map(ToOwned::to_owned),
+            })
+        }
+
+        // TODO: Add request timeouts
```
fl0rek, line 211: Is this left to be addressed in the future or this PR?

rpc/tests/utils/client.rs
```diff
@@ -27,31 +28,72 @@ pub enum AuthLevel {
...
     }
 }
 
+#[cfg(target_arch = "wasm32")]
+fn token_from_env(auth_level: AuthLevel) -> Result<Option<String>> {
+    #[derive(rust_embed::Embed)]
+    #[folder = "$CARGO_MANIFEST_DIR/../"]
+    #[allow_missing = true]
+    #[include = ".env*"]
+    struct Env;
```
fl0rek, line 47: Cool idea with embedding the env file for wasm 👌

fl0rek (commented):
nice workaround! few clarifying questions, otherwise lgtm

client/src/blob.rs
zvolin: re-added this todo but in rpc

rpc/src/client.rs
zvolin: yeah I was thinking in a future, together with exposing this config on native, within https://github.com/eigerco/lumina/issues/791

rpc/tests/utils/client.rs
zvolin: ty :hugs:

rpc/src/client.rs
```diff
@@ -163,41 +163,86 @@ mod native {
...
-        /// be specified in the provided `conn_str`. For more flexibility
-        /// consider creating the client using [`jsonrpsee`] directly.
-        ///
-        /// Since headers are not supported in the current version of
-        /// `jsonrpsee-wasm-client`, celestia-node requires disabling
-        /// authentication (--rpc.skip-auth) to use wasm.
-        ///
-        /// For a secure connection you have to hide it behind a proxy.
-        pub async fn new(conn_str: &str) -> Result<Self, Error> {
+        /// Only the 'http\[s\]' protocols are supported because javascirpt
```
fl0rek:
```suggestion
        /// Only the 'http\[s\]' protocols are supported because JavaScript
```

rpc/src/fraud.rs
```diff
@@ -1,20 +1,108 @@
...
+                        }
+
+                        // after we got some proofs from the node, it would
+                        // keep giving us the same proofs again and again,
+                        // so we just end the stream here
+                        break;
+                    }
+                };
+            }
+        }
```
fl0rek, line 97:
```suggestion
        try_stream! {
            match rpc::FraudSubscriptionClient::fraud_subscribe(self, proof_type).await {
                Ok(mut fraud_sub) => loop {
                    yield fraud_sub.next().await
                        .ok_or_else(|| custom_client_error("unexpected end of stream"))??;
                },
                Err(Error::HttpNotImplemented) => {
                    let mut header_sub = HeaderClient::header_subscribe(self);
                    loop {
                        // tick; we don't care about the header
                        header_sub
                            .next()
                            .await
                            .ok_or_else(|| custom_client_error("unexpected end of stream"))??;

                        let proofs = rpc::FraudClient::fraud_get(self, proof_type).await?;
                        if !proofs.is_empty() {
                            for proof in proofs {
                                yield proof;
                            }

                            // after we got some proofs from the node, it would
                            // keep giving us the same proofs again and again,
                            // so we just end the stream here
                            break;
                        }
                    }
                }
                other => {other?;}
            };
        }
```
I wonder if this could be simplified a bit, by having two separate loops instead of one and using Options.. Above Suggestion shaved of 10 out of 40 lines, but could probably be improved further.

rpc/src/header.rs
```diff
@@ -1,50 +1,197 @@
...
+                    rpc::HeaderClient::header_wait_for_height(self, head.height().value() + 1).await?
+                };
+
+                header.validate().map_err(custom_client_error)?;
+                head.verify_adjacent(&header).map_err(custom_client_error)?;
+
+                head = header.clone();
+                yield header;
+            }
+        }
```
fl0rek, line 154:
```suggestion
        try_stream! {
            let mut head = rpc::HeaderClient::header_local_head(self).await?;
            head.validate().map_err(custom_client_error)?;

            let mut real_subscription = match rpc::HeaderSubscriptionClient::header_subscribe(self).await {
                Ok(subscription) => Ok(Some(subscription)),
                Err(Error::HttpNotImplemented) => Ok(None),
                Err(e) => Err(e)
            }?;

            loop {
                let header = match &mut real_subscription {
                    Some(subscription) => subscription.next().await
                        .ok_or_else(|| custom_client_error("unexpected end of stream"))??,
                    None => rpc::HeaderClient::header_wait_for_height(self, head.height().value() + 1).await?,
                };

                header.validate().map_err(custom_client_error)?;
                head.verify_adjacent(&header).map_err(custom_client_error)?;

                head = header.clone();
                yield header;
            }
        }
```
similar point to above

fl0rek (approved):
The emulated subscription loop part was rubbing me a wrong way, touch annoying we cannot just `return Err(..)` there. Left best proposition I could come up with, leaving it to your judgement
Other than that, LGTM!

rpc/src/fraud.rs
```diff
@@ -1,20 +1,108 @@
...
+
+                    let proofs = rpc::FraudClient::fraud_get(self, proof_type).await?;
+                    if !proofs.is_empty() {
+                        for proof in proofs {
+                            yield proof;
+                        }
+
+                        // after we got some proofs from the node, it would
+                        // keep giving us the same proofs again and again,
+                        // so we just end the stream here
```
fl0rek:
```suggestion
                        // After we get any proofs from the node, it will
                        // keep giving us the same proofs on each request,
                        // so we just end the stream here
```

PR #801, opened 2025-11-11
fix(ci): Workaround `discover_network_peers` test of `lumina-node-wasm`

oblique:


node-wasm/tests/discovery.rs
```diff
@@ -0,0 +1,53 @@
...
+
+mod utils;
+
+use crate::utils::{
+    fetch_bridge_webtransport_multiaddr, new_rpc_client, remove_database, spawn_connected_node,
+};
+
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+// NOTE: This test is splitted from the rest because when it runs in the same
```
fl0rek:
```suggestion
// NOTE: This test is split from the rest because when it runs in the same
```

zvolin (approved):

PR #807, opened 2025-11-13
feat(node)!: Make HeaderEx aware of node types and connected peers

oblique:


node/src/node.rs
```diff
@@ -331,6 +331,11 @@ where
         Ok(self.p2p().set_peer_trust(peer_id, is_trusted).await?)
     }
 
+    /// Mark the peer as archival.
+    pub async fn mark_as_archival(&self, peer_id: PeerId) -> Result<()> {
```
oblique, line 336: This is mostly needed for our test cases. Should I hide from the docs?

zvolin: up to you

zvolin: but I guess we can hide it

zvolin: or have it under `test_utils` feature, but I don't remember if we have one in node

node/src/p2p/header_ex.rs
```diff
@@ -130,13 +140,13 @@ where
...
         &mut self,
         ev: ToSwarm<ReqRespEvent, THandlerInEvent<ReqRespBehaviour>>,
-    ) -> Option<ToSwarm<(), THandlerInEvent<Self>>> {
+    ) -> Option<ToSwarm<Event, THandlerInEvent<Self>>> {
         match ev {
             ToSwarm::GenerateEvent(ev) => {
                 self.on_req_resp_event(ev);
                 None
             }
-            _ => Some(ev.map_out(|_| ())),
```
zvolin, line 139: I'm a bit afraid of this unreachable. Cannot ReqRespBehaviour generate other events like dial / notify handler? Or even if it cannot now, maybe it gets added in future and we will start panicking there? We could instead `map_out` it to any of our variant, as we are sure it is a noop here

node/src/p2p/header_ex/client.rs
```diff
@@ -116,122 +163,198 @@ where
...
+    }
 
-        trace!("Request initiated");
+    fn has_pending_head_requests(&self) -> bool {
+        !self.head_reqs.is_empty() && !self.head_req_scheduled
     }
 
-    fn send_request(
+    fn has_pending_requests(&self) -> bool {
+        self.has_pending_head_requests() || self.pending_reqs.values().any(|reqs| !reqs.is_empty())
```
zvolin: myself todo: ask q

node/src/p2p/header_ex/client.rs
```diff
@@ -116,122 +163,198 @@ where
...
+                PeerKind::Any => peer.is_connected(),
+                PeerKind::Archival => peer.is_connected() && peer.is_archival(),
+                PeerKind::Trusted => peer.is_connected() && peer.is_trusted(),
+                PeerKind::TrustedArchival => {
+                    peer.is_connected() && peer.is_trusted() && peer.is_archival()
+                }
+            })
+            .collect::<SmallVec<[_; MAX_PEERS]>>();
+
+        if !peers.is_empty() {
```
zvolin, line 274: shouldn't we have an `else` here that schedules events to get archival / trusted peers, instead of always scheduling them if there are any requests, even tho we may have such peers already?

node/src/p2p.rs
```diff
@@ -962,6 +968,25 @@ where
...
+        match ev {
+            header_ex::Event::SchedulePendingRequests => {
+                let ctx = self.swarm.context();
+
+                ctx.behaviour
+                    .header_ex
+                    .schedule_pending_requests(ctx.peer_tracker);
+            }
+            header_ex::Event::NeedTrustedPeers => {
+                self.swarm.connect_to_bootnodes();
```
zvolin, line 985: this will actually skip peers which were manually set as trusted, as we have such api

node/src/p2p/header_ex/client.rs
```diff
@@ -85,26 +119,39 @@ impl Drop for OneshotSender {
...
+
 impl<S> HeaderExClientHandler<S>
 where
     S: RequestSender,
 {
     pub(super) fn new() -> Self {
         HeaderExClientHandler {
             reqs: HashMap::new(),
+            head_reqs: VecDeque::new(),
+            head_req_scheduled: false,
```
zvolin, line 136: I have a feeling this is completely unneeded, wouldn't everything work the same if this field hasn't existed?

node/src/p2p/header_ex/client.rs
```diff
@@ -1184,25 +1428,38 @@ mod tests {
...
 
         let mut generator = ExtendedHeaderGenerator::new_from_height(5);
-        let mut invalid_header5 = generator.next();
+        let header5 = generator.next();
+        let mut invalid_header5 = header5.clone();
         invalidate(&mut invalid_header5);
 
-        mock_req.send_n_responses(&mut handler, 10, vec![invalid_header5.to_header_response()]);
+        // Head request retry indefinitely
+        for _ in 0..5 {
```
zvolin: maybe something like `0..MAX_RETRIES+2` to indicate it goes beyond the usual const?

node/src/p2p/header_ex/client.rs
```diff
@@ -1211,22 +1468,37 @@ mod tests {
...
-            HeaderRequest::with_origin(0, 1),
-            tx,
-            &peer_tracker,
-        );
+        handler.on_send_request(HeaderRequest::with_origin(0, 1), tx);
 
-        mock_req.send_n_failures(&mut handler, 5, OutboundFailure::Timeout);
-        mock_req.send_n_failures(&mut handler, 5, OutboundFailure::ConnectionClosed);
+        // Head request retry indefinitely
+        for _ in 0..5 {
```
zvolin: likewise

node/src/p2p/header_ex/client.rs
```diff
@@ -1313,6 +1600,185 @@ mod tests {
...
+        assert!(handler.has_pending_requests());
+
+        // Try poll without peers
+        let ev = poll_client(&mut handler, &mut mock_req, &empty_peer_tracker).await;
+        // `SchedulePendingRequests` is generated when there are pending requests.
+        // Since we have no peers, the requests are still pending.
+        assert!(matches!(ev, Event::SchedulePendingRequests));
+        assert!(handler.has_pending_requests());
+
+        // Poll again a peer
```
zvolin:
```suggestion
        // Poll again with a connected peer
```

node/src/p2p/header_ex/client.rs
```diff
@@ -1313,6 +1600,185 @@ mod tests {
...
+            &mut handler,
+            &mut mock_req,
+            &peer_tracker,
+            Duration::from_millis(10),
+        )
+        .await;
+        // Request is now scheduled (i.e. not pending)
+        assert!(!handler.has_pending_requests());
+
+        // Responde with failure
```
zvolin:
```suggestion
        // Respond with failure
```

node/src/p2p/header_ex/client.rs
```diff
@@ -1313,6 +1600,185 @@ mod tests {
...
```
zvolin:
```suggestion
        // Respond with failure
```

node/src/p2p/swarm_manager.rs
```diff
@@ -34,7 +34,8 @@ const BOOTNODE_PROTECT_TAG: u32 = 0;
 const FULL_PROTECT_TAG: u32 = 1;
 const ARCHIVAL_PROTECT_TAG: u32 = 2;
 
-const PEER_HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(60);
+const PEER_HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(10 * 60);
```
zvolin: 10 mins are quite a lot, but it is needed only if there are no events from peer tracker, no?

node/src/p2p/header_ex.rs
oblique: The `map_out` is not very descriptive from its name, but it only maps the type of `ToSwarm::GenerateEvent` variant. And we actually need it because we consume the generated event. You can check the code [here](https://github.com/libp2p/rust-libp2p/blob/0125fa96b8de44ce3c3836a26339bfd9ba6db6fe/swarm/src/behaviour.rs#L381-L414).

node/src/p2p/header_ex/client.rs
oblique: No. We clear the `head_reqs` only after we figure out the head, not when we schedule it.

node/src/p2p/header_ex.rs
zvolin: yeah I was aware of what it does, I guess I just got confused

node/src/p2p/header_ex/client.rs
zvolin: yeah, now it clicks :p

node/src/node.rs
oblique: exposed it with `test_utils`

node/src/p2p/header_ex/client.rs
oblique: We could, but I prefer them in the caller because it checks also `head_reqs` and `head_reqs_scheduled` fields.

node/src/p2p/header_ex/client.rs
oblique: added

node/src/p2p/swarm_manager.rs
oblique: accidentally commited it. i was manually testing something. reverted

node/src/p2p.rs
oblique: We need a small refactor if we want to have `connect_to_trusted_nodes`. I opened an issue: https://github.com/eigerco/lumina/issues/824

node/src/p2p/header_ex/client.rs
zvolin: I was confused here, https://github.com/eigerco/lumina/pull/807/commits/76a066462738d8baab56d7ab86d74c43db0555d6 clarifies that

node/src/p2p/header_ex/client.rs
```diff
@@ -116,122 +163,198 @@ where
...
-        let state = State {
-            request,
-            respond_to: OneshotSender::new(respond_to),
-        };
+        }
 
-        self.reqs.insert(req_id, state);
+        let mut peers = peer_tracker
+            .peers()
+            .filter(|peer| match peer_kind {
```
zvolin, line 264: would it be nice of us to prioritize full nodes here over light ones? not saying to do it right now, but might be a nice addition in a future :thinking:

node/src/p2p/header_ex/client.rs
```diff
@@ -116,122 +163,198 @@ where
...
+                PeerKind::Any => peer.is_connected(),
+                PeerKind::Archival => peer.is_connected() && peer.is_archival(),
+                PeerKind::Trusted => peer.is_connected() && peer.is_trusted(),
+                PeerKind::TrustedArchival => {
+                    peer.is_connected() && peer.is_trusted() && peer.is_archival()
+                }
+            })
```
fl0rek, line 271:
```suggestion
            .map(|peer| peer.id())
            })
```
Just Id is used later, so let's just use that since there's shuffling of the vector.

node/src/p2p/header_ex/client.rs
```diff
@@ -290,29 +411,137 @@ where
...
+                        self.pending_reqs
+                            .entry(peer_kind)
+                            .or_default()
+                            .push_back(State { peer_kind, ..state });
+
+                        continue;
+                    }
+
+                    let res = res.map_err(P2pError::from);
+                    state.respond_to.maybe_send(res);
```
fl0rek, line 485:
```suggestion
                    } else {
                        let res = res.map_err(P2pError::from);
                        state.respond_to.maybe_send(res);
                    }
```
maybe just else? ;p

node/src/p2p/header_ex/client.rs
oblique: Yeah, in general we can replace shuffling with some kind of sorting, which we can use peer scoring and node type etc.

node/src/p2p/header_ex/client.rs
oblique: i prefer the early return style (early continue in this case) because it looks a bit less nested

node/src/p2p/header_ex/client.rs
oblique: The vector collects of `&Peer`, so using the id will not change anything

PR #818, opened 2025-11-20
refactor(ci): use compose healthchecks to ensure docker setup is ready

zvolin:


ci/docker-compose.yml
```diff
@@ -49,9 +53,13 @@ services:
       - SKIP_AUTH=true
     ports:
       - 26658:26658
-    volumes:
-      - credentials:/credentials
-      - genesis:/genesis
+    volumes: *vols
+    healthcheck: &healthcheck-node
+      # wait for the node to start json-rpc server
+      test: celestia header local-head
```
oblique, line 59: Shouldn't we also wait for a specific height?

oblique (approved):

ci/docker-compose.yml
zvolin: validators wait for height 5, so this is ensured

PR #819, opened 2025-11-21
feat(rpc,grpc,client)!: Add request, connection timeouts for gRPC, jRPC, celestia-client

fl0rek:


client/src/state.rs
```diff
@@ -725,4 +732,40 @@ mod tests {
...
+            Error::Rpc(RpcError::JsonRpc(JrpcError::RequestTimeout)),
+            _client_build_error
+        ));
+    }
+
+    #[async_test]
+    async fn grpc_timeout() {
+        let client = new_client().await;
+
+        let _balance_timeout = client
```
oblique:
```suggestion
        let balance_timeout = client
```

client/src/state.rs
```diff
@@ -725,4 +732,40 @@ mod tests {
             api.query_redelegations(&val_addr, &val_addr).await.unwrap(),
         );
     }
+
+    #[async_test]
+    async fn rpc_timeout() {
+        let _client_build_error = Client::builder()
```
oblique:
```suggestion
        let client_build_error = Client::builder()
```

grpc/src/client.rs
```diff
@@ -1346,6 +1349,40 @@ mod tests {
...
+            panic!("invalid error type, expected TonicError");
+        };
+        assert_eq!(timeout.code(), Code::DeadlineExceeded);
+
+        // too short timeout set in builder
+        let client = GrpcClient::builder()
+            .url(CELESTIA_GRPC_URL)
+            .timeout(Duration::from_nanos(1))
+            .build()
+            .unwrap();
```
oblique, line 1557: Call `get_all_balances` after this to check if error is received

rpc/src/client.rs
```diff
@@ -177,52 +201,89 @@ mod wasm {
...
             let fut = {
                 let mut req = JsRequest::post(&self.url);
 
+                if let Some(timeout) = self.timeout_ms {
+                    let abort_controller = AbortController::new().unwrap(); // FIXME: unwrap
+                    let abort_signal = abort_controller.signal();
+                    Timeout::new(timeout, move || {
+                        abort_controller.abort();
+                    })
+                    .forget();
```
oblique: Will this leak or garbage collected when timeout is triggered?

rpc/src/client.rs
```diff
@@ -177,52 +201,89 @@ mod wasm {
...
             })
         }
 
-        // TODO: Add request timeouts
         async fn send<T: Serialize>(&self, request: T) -> Result<JsResponse, ClientError> {
             let fut = {
                 let mut req = JsRequest::post(&self.url);
 
+                if let Some(timeout) = self.timeout_ms {
+                    let abort_controller = AbortController::new().unwrap(); // FIXME: unwrap
```
zvolin: do you plan on doing the `FIXME` in this scope?

zvolin (commented):
looks nice, although it seems builder is hidden in docs
image

rpc/src/client.rs
fl0rek: It internally calls [forget](https://docs.rs/gloo-timers/latest/src/gloo_timers/callback.rs.html#95) on the closure,  [only reference](https://docs.rs/gloo-timers/latest/src/gloo_timers/callback.rs.html#65) to it remaining is the one passed to the js timeout function. As I understand it, this two things will make it so that js does the cleanup, once gc kicks in. 

For what it's worth, this exits cleanly:
```Rust
    #[async_test]
    async fn wasm_cleanup() {
        let (tx, mut rx) = mpsc::unbounded_channel();
        gloo_timers::callback::Timeout::new(1000, move || {
            tx.send(1).unwrap();
        })
        .forget();
        assert_eq!(rx.recv().await, Some(1));
        assert_eq!(rx.recv().await, None);
+    }
```

rpc/src/client.rs
fl0rek: AbortController is marked as baseline widely available in mdn, so switched to expect here.

grpc/src/client.rs
fl0rek: added

rpc/src/client.rs
oblique: So the `Timeout` will stay in memory until it expires and it will eventually garbage collected.

I think it's more proper to drop it when we exit the function.

fl0rek: Good point actually, moved it to a local variable so that it can be dropped automatically and thus the dangling timeout will be cleared.

zvolin (approved):

PR #822, opened 2025-11-21
chore(grpc): fix flaky submit_and_get_tx test

zvolin:


fl0rek (approved):

grpc/src/client.rs
```diff
@@ -1084,6 +1085,8 @@ mod tests {
             .unwrap();
         let tx2 = tx_client.get_tx(tx.hash).await.unwrap();
 
+        lumina_utils::time::sleep(Duration::from_millis(100)).await;
```
oblique: Import `sleep` instead

PR #830, opened 2025-11-28
chore: workaround the floodsub incompatibility

zvolin:


node/src/p2p/shr_ex.rs
```diff
@@ -44,12 +53,34 @@ impl<S> Behaviour<S>
...
+            // and floodsub publish behaviour
+            .flood_publish(true)
+            .validation_mode(gossipsub::ValidationMode::Strict)
+            .validate_messages()
+            .build()
+            .map_err(|e| P2pError::GossipsubInit(e.to_string()))?;
+
+        // build a gossipsub network behaviour
+        let mut shr_ex_sub: gossipsub::Behaviour =
+            gossipsub::Behaviour::new(message_authenticity, shr_ex_sub_config)
```
fl0rek:
```suggestion
        let mut shrex_sub: gossipsub::Behaviour =
            gossipsub::Behaviour::new(message_authenticity, shrex_sub_config)
```

node/src/p2p/shr_ex.rs
```diff
@@ -32,7 +33,15 @@ where
     S: Store + 'static,
 {
     req_resp: request_response::Behaviour<TodoCodec>,
-    shr_ex_sub: floodsub::Behaviour,
+    // TODO: this is a workaround, should be replaced with a real floodsub
+    // rust-libp2p implementation of the floodsub isn't compliant with a spec,
+    // so we work that around by using a gossipsub configured with floodsub support.
+    // Gossipsub will always receive from and forward messages to all the floodsub peers.
+    // Since we always maintain a connection with a few bridge (and possibly archival)
+    // nodes, then if we assume those nodes correctly use only floodsub protocol there,
```
fl0rek:
```suggestion
    // nodes, then if we assume those nodes correctly use only floodsub protocol,
```
could just be dropped I think

fl0rek (approved):
Nice idea. Left minor notes, otherwise looks good

oblique (approved):

Issue #833, opened 2025-12-01
GRPC endpoint fallthrough to backups


evan-forbes:
users need a way to switch to different rpcs if one fails automatically.

currently the go client uses a parallel submission approach. that approach has an insane number of edge cases that tbh I don't think are covered. instead of submitting to multiple endpoints in parallel, we can always go through a list of providers and fallthrough to the next provider after the current one fails for a non-broadcast related error

for example, we get no response or are unable to establish a connection because a node is down. Then we fallthrough to the next endpoint to try there. We don't fallthrough for things such as a sequence mismatch, that is still handled as we already handle things here

PR #836, opened 2025-12-02
feat(grpc): use multiple endpoints, fallback in case of errors

mcrakhman:
## Overview

This implements fallback as indicated in this issue: https://github.com/eigerco/lumina/issues/833.

**Retry with different url:** 
- network errors (e.g. connection etc)

**Not retry (require action from a client):**
- grpc errors (e.g. endpoint not found,..)
- application errors (e.g. sequence mismatch,..)

### DIFF
--- a/node-wasm/js/package.json
+++ b/node-wasm/js/package.json
@@ -35,7 +35,7 @@
         "typescript": "^5.6.3"
     },
     "scripts": {
-        "tsc": "tsc index.js --declaration --allowJs --emitDeclarationOnly",
+        "tsc": "tsc index.js --lib dom,dom.iterable,esnext --declaration --allowJs --emitDeclarationOnly",
         "update-readme": "typedoc --plugin typedoc-plugin-markdown --out ./docs --readme ../README.md ../pkg/lumina_node_wasm.d.ts && concat-md  --decrease-title-levels --dir-name-as-title docs > README.md"
     }
 }


client/src/client.rs
```diff
@@ -226,6 +226,23 @@ impl ClientBuilder {
...
 
+    /// Add multiple gRPC endpoints at once for fallback support.
+    ///
+    /// When multiple endpoints are configured, the client will automatically
+    /// fall back to the next endpoint if a network-related error occurs.
+    ///
+    /// # Note
+    ///
+    /// In WASM the endpoints need to support gRPC-Web.
+    pub fn grpc_urls(mut self, urls: impl IntoIterator<Item = impl AsRef<str>>) -> ClientBuilder {
```
zvolin, line 237: this helper to set multiple urls at once looks nice, could you implement it in `celestia_grpc::GrpcClientBuilder` and just re-use it here and in other places?

grpc/src/builder.rs
```diff
@@ -50,15 +47,22 @@ impl GrpcClientBuilder {
         GrpcClientBuilder::default()
     }
 
-    /// Set the `url` to connect to using [`Channel`] transport.
+    /// Add a URL endpoint. Multiple calls add multiple fallback endpoints.
+    ///
+    /// When multiple endpoints are configured, the client will automatically
+    /// fall back to the next endpoint if a network-related error occurs.
     ///
     /// [`Channel`]: tonic::transport::Channel
```
zvolin: since we no longer mention the `Channel` in docs, this link could be removed

grpc/src/builder.rs
```diff
@@ -131,12 +136,23 @@ impl GrpcClientBuilder {
...
+        let transports: Vec<BoxedTransport> = self
+            .transports
+            .into_iter()
+            .map(|entry| match entry {
+                TransportEntry::EndpointUrl(url) => build_transport(url),
+                TransportEntry::BoxedTransport(t) => Ok(t),
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let transports: std::sync::Arc<[BoxedTransport]> = transports.into();
```
zvolin: nit, I'd move this call to  `GrpcClient::new(transports.into(), ...)`, this way we don't need to specify the type

grpc/src/error.rs
```diff
@@ -166,3 +187,81 @@ impl From<GrpcClientBuilderError> for wasm_bindgen::JsValue {
...
+        assert!(!Error::FailedToParseResponse.is_network_error());
+        assert!(!Error::TxEmptyBlobList.is_network_error());
+        assert!(!Error::MissingSigner.is_network_error());
+        assert!(!Error::PublicKeyMismatch.is_network_error());
+        assert!(!Error::UnexpectedResponseType("test".into()).is_network_error());
+        assert!(!Error::SequenceParsingFailed("test".into()).is_network_error());
+    }
+
+    #[cfg(not(target_arch = "wasm32"))]
+    #[tokio::test]
```
fl0rek, line 256: Please use `#[async_test]` from `lumina_utils::test_utils`, it automatically routes tests to wasm_bindgen_test on wasm.

grpc/src/builder.rs
zvolin: or actually we can collect directly to the `Arc`, then I'd just like it to be imported rather than used with full path starting from `std`

grpc/src/error.rs
mcrakhman: In this case we don't want this test to run on wasm, because it doesn't have this specific error

grpc/grpc-macros/src/lib.rs
```diff
@@ -53,30 +53,66 @@ impl GrpcMethod {
...
+                        let mut client = #grpc_client_struct::new(transport)
+                            .max_decoding_message_size(MAX_MSG_SIZE)
+                            .max_encoding_message_size(MAX_MSG_SIZE);
+
+                        let mut request = ::tonic::Request::from_parts(
+                            context.metadata.clone(),
+                            ::tonic::Extensions::new(),
+                            ::std::clone::Clone::clone(&param),
+                        );
+                        request.set_timeout(::std::time::Duration::from_secs(30));
```
evan-forbes, line 78: to confirm, this timeout is just a one time timeout correct? we're not always returning after 30s from now on are we?

zvolin: wdym by "one time timeout"? this just sets the `grpc-timeout` header on the request. Actually I think it may not be the best approach here, because in case of network errors, server may not even see that

grpc/grpc-macros/src/lib.rs
```diff
@@ -53,30 +53,66 @@ impl GrpcMethod {
...
+
+                        match fut.await {
+                            Ok(resp) => {
+                                // If we succeeded on a fallback endpoint (idx > 0),
+                                // move it to the front for future requests
+                                if idx > 0 {
+                                    let mut guard = transports.lock().await;
+                                    if guard.len() > idx {
+                                        guard.swap(0, idx);
+                                    }
```
zvolin: swap here doesn't really work. We should instead make a swap on a local vector (which is a bit problematic because we iter it, maybe we need to switch to only `idx`) and just `mem::replace`. If you run 2 parallel grpc calls at the same time, and they both reach the `.lock()` here, one will just undo the work of the other.
Although since we clone here anyway, and the `Vec` only has `Box`es, maybe that's actually a good place to remove any locking and have an [`ArcSwap`](https://docs.rs/arc-swap/latest/arc_swap/type.ArcSwap.html)? @oblique wdyt

grpc/grpc-macros/src/lib.rs
zvolin: > we're not always returning after 30s from now on are we?

but I believe that was the intention, is that an issue?

grpc/grpc-macros/src/lib.rs
zvolin: I think I like the idea of arcswap here, as we only want to have it eventually corrected, it's not an issue if some new request is made using a faulty endpoint and going into the fallback

mcrakhman: maybe we can just set the whole local sequence after swap then the calls would be idempotent

mcrakhman: ah ok thats exactly what you suggested

client/src/client.rs
```diff
@@ -226,6 +226,23 @@ impl ClientBuilder {
...
+        let mut grpc_builder = self.grpc_builder.unwrap_or_default();
+        for url in urls {
+            grpc_builder = grpc_builder.url(url.as_ref());
+        }
+        self.grpc_builder = Some(grpc_builder);
```
zvolin:
```suggestion
        self.grpc_builder = Some(grpc_builder.urls(urls));
```

grpc/src/builder.rs
```diff
@@ -50,15 +49,31 @@ impl GrpcClientBuilder {
         GrpcClientBuilder::default()
     }
 
-    /// Set the `url` to connect to using [`Channel`] transport.
+    /// Add multiple URL endpoints.
+    ///
+    /// When multiple endpoints are configured, the client will automatically
+    /// fall back to the next endpoint if a network-related error occurs.
+    pub fn grpc_urls(mut self, urls: impl IntoIterator<Item = impl AsRef<str>>) -> Self {
```
zvolin: The `grpc_` prefix is used only in `celestia-client` as we need to differentiate between json-rpc and grpc, here we can drop it

```suggestion
    pub fn urls(mut self, urls: impl IntoIterator<Item = impl AsRef<str>>) -> Self {
```

grpc/src/js_client.rs
```diff
@@ -75,6 +75,20 @@ impl GrpcClientBuilder {
...
+    ///
+    /// When multiple endpoints are configured, the client will automatically
+    /// fall back to the next endpoint if a network-related error occurs.
+    ///
+    /// Note that this method **consumes** builder and returns updated instance of it.
+    /// Make sure to re-assign it if you keep builder in a variable.
+    #[wasm_bindgen(js_name = "withUrls")]
+    pub fn with_urls(self, urls: Vec<String>) -> Self {
+        Self {
+            inner: self.inner.grpc_urls(urls),
```
zvolin:
```suggestion
            inner: self.inner.urls(urls),
```

grpc/src/js_client/grpc_client.rs
```diff
@@ -31,6 +31,15 @@ impl GrpcClient {
         crate::GrpcClientBuilder::new().url(url).into()
     }
 
+    /// Create a builder for [`GrpcClient`] with multiple URL endpoints for fallback support.
+    ///
+    /// When multiple endpoints are configured, the client will automatically
+    /// fall back to the next endpoint if a network-related error occurs.
+    #[wasm_bindgen(js_name = withUrls)]
+    pub fn with_urls(urls: Vec<String>) -> GrpcClientBuilder {
+        crate::GrpcClientBuilder::new().grpc_urls(urls).into()
```
zvolin:
```suggestion
        crate::GrpcClientBuilder::new().urls(urls).into()
```

grpc/src/uniffi_client.rs
```diff
@@ -69,6 +69,16 @@ impl GrpcClientBuilder {
         GrpcClientBuilder(Mutex::new(Some(builder)))
     }
 
+    /// Create a new builder with multiple fallback URLs.
+    ///
+    /// When multiple endpoints are configured, the client will automatically
+    /// fall back to the next endpoint if a network-related error occurs.
+    #[uniffi::constructor(name = "withUrls")]
+    pub fn with_urls(urls: Vec<String>) -> Self {
+        let builder = RustBuilder::new().grpc_urls(urls);
```
zvolin:
```suggestion
        let builder = RustBuilder::new().urls(urls);
```

grpc/grpc-macros/Cargo.toml
```diff
@@ -17,3 +17,4 @@ proc-macro = true
 quote = "1.0.35"
 syn = { version = "2.0.87", features = ["full", "extra-traits"] }
 proc-macro2 = "1.0.83"
+
```
zvolin:
```suggestion
```

grpc/grpc-macros/src/lib.rs
```diff
@@ -53,30 +53,60 @@ impl GrpcMethod {
...
+                                        error,
+                                    );
+                                    last_error = Some(error);
+                                    continue;
+                                }
+                                return Err(error);
+                            }
+                        }
+                    }
+                    Err(last_error.expect("at least one transport should be tried"))
```
oblique, line 110:
```suggestion

                    Err(last_error.expect("at least one transport should be tried"))
```

grpc/grpc-macros/src/lib.rs
fl0rek: As written this changes the timeout to be 30sec by default.

grpc/src/boxed.rs
```diff
@@ -12,6 +13,25 @@ use tonic::codegen::Service;
 
 use crate::utils::CondSend;
 
+/// Metadata associated with a transport endpoint
+#[derive(Debug, Clone, Default)]
+pub(crate) struct TransportMetadata {
```
fl0rek, line 18: Where is this used? I'm having trouble figuring out why it's needed.

mcrakhman: Because we want to say which transport failed in logs

PR #837, opened 2025-12-04
chore(node-wasm): fix flaky request_network_head_header

zvolin:


node-wasm/tests/client.rs
```diff
@@ -31,9 +32,19 @@ async fn request_network_head_header() {
...
+        rpc_client.header_network_head().map(Result::unwrap),
+        client.request_head_header().map(Result::unwrap),
+    );
+
+    // due to head selection algorithm, we may get the network head or a previous header
+    // if not enough nodes got a new head already
+    if head_header.height() == bridge_head_header.height() {
+        assert_eq!(head_header, bridge_head_header);
+    } else {
+        assert!(head_header.verify_adjacent(&bridge_head_header).is_ok());
```
oblique, line 45: The adjacent will work only in case of `head_header` been smaller than `bridge_head_header`.

zvolin: yes

zvolin: I think this one will always be latest, and that only from our node we may eventually get older one since we do a head selection algorithm when requesting it

oblique (approved):

fl0rek (approved):

PR #842, opened 2025-12-17
feat(node)!: Implement ShrEx client

oblique:


types/src/eds.rs
```diff
@@ -479,16 +478,15 @@ impl EdsId {
...
-            // TODO: change error
-            return Err(CidError::InvalidMultihashLength(buffer.len()));
+            return Err(Error::InvalidLength(buffer.len(), EDS_ID_SIZE));
         }
 
         let height = buffer.get_u64();
 
         if height == 0 {
-            return Err(CidError::InvalidCid("Zero block height".to_string()));
+            return Err(Error::ZeroBlockHeight);
```
oblique: We can remove this check and just use `EdsId::new(height)`

PR #843, opened 2025-12-17
feat(grpc): Split submission and confirmation

fl0rek:


grpc/src/client.rs
```diff
@@ -372,7 +418,62 @@ impl GrpcClient {
...
+
+        AsyncGrpcCall::new(move |context| async move {
+            let broadcasted_tx = this
+                .clone()
+                .submit_blobs_impl(&blobs, cfg.clone(), &context)
+                .await?;
+
+            Ok(SubmittedTx::new(
+                broadcasted_tx.clone(),
+                AsyncGrpcCall::new(move |context| async move {
```
mcrakhman, line 471: afaiu the problem with this approach is that the `AsyncGrpcCall` is not `Send`, thus it can't be moved to a separate thread where we want to await on confirmation (this is how we would like to use it).

Is it possible to make it `Send`? 

If not, would it make sense to either:
- just return `BroadcastedTx` and expose a method `confirm_tx` which creates a new `AsyncGrpcCall`.
- or return the same thing, but leave the option for the caller to decide if it wants to call the provided `confirm()` in the same thread or to use another new methods to call in a different thread with `BroadcastedTx`.

I assume you used this architecture to allow the caller to implicitly reuse the context of the first call for the second call.

fl0rek: ~~I don't think we can make AsyncGrpcCall Send easily - call there holds a reference to GrpcClient, we'd need to add synchronisation for the entire client - this could be added together with tx client v2 implementation, since we'll be adding queues there anyway.~~
With the current code though, one could take the broadcasted tx data and build a confirm_tx requests in another thread with another client. Would that work as a sub for now?~

mcrakhman: Yes, seems fine!

grpc/src/client.rs
```diff
@@ -372,7 +418,76 @@ impl GrpcClient {
...
+                    this.confirm_tx_impl(broadcasted_tx, cfg, &context).await?;
+                    todo!()
```
```suggestion
```

grpc/src/client.rs
zvolin: we can actually make it send easily, it's drafted there: https://github.com/eigerco/lumina/pull/844

an immediate workaround would be to call `.into_future()` on the `AsyncGrpcCall` before moving it to another thread too

zvolin (approved):

PR #851, opened 2025-12-30
feat(node): Shrex sub blacklist

fl0rek:


node/src/p2p.rs
```diff
@@ -707,6 +708,7 @@ where
     B: Blockstore + 'static,
     S: Store + 'static,
 {
+    blacklist: allow_block_list::Behaviour<BlockedPeers>,
```
zvolin: I think better place for this behavior would be `connection_control`, and a block method exposed on swarm manager

zvolin: or maybe just inside swarm manager, as connection_control is a singular behavior rn

zvolin (approved):

PR #863, opened 2026-01-06
feat: reconnect on RestartNeeded

mcrakhman:
## Overview
Add reconnecting WS client with retry logic + unit tests. Resolves this: https://github.com/eigerco/lumina/issues/864

### Problem
We encountered presumably some disconnect scenarios which lead the client to be in a state that needs Restart. When the connection drops, retries at higher levels would just hit the same closed WS background task, so progress could stall. Previously this required re-creating the client and starting again.

### Changes summary

Add retry logic that will reconnect once and pause all the current requests using the same client until the reconnection completes.


rpc/src/client.rs
```diff
@@ -74,26 +72,231 @@ mod native {
...
+
+    type BuildFn<C> =
+        Arc<dyn Fn() -> futures_util::future::BoxFuture<'static, Result<C, Error>> + Send + Sync>;
+
+    impl WsReconnectClient<WsClient> {
+        async fn new(
+            url: &str,
+            auth_token: Option<&str>,
+            connect_timeout: Option<Duration>,
+            request_timeout: Option<Duration>,
```
zvolin, line 103: I think if we just stored all those options as fields inside `WsReconnectClient` then we wouldn't need the `BulidFn` and family, and the code would be simpler

rpc/src/client.rs
```diff
@@ -129,7 +332,15 @@ mod native {
...
-                Client::Ws(client) => client.batch_request(batch).await,
+                Client::Ws(client) => {
+                    let inner = client.get_inner().await?;
+                    let err = match inner.batch_request(batch).await {
+                        Err(err @ ClientError::RestartNeeded(_)) => err,
+                        res => return res,
+                    };
+                    client.mark_poisoned().await;
+                    Err(err)
+                }
```
zvolin, line 311: could you move this impl next to all the others?

rpc/src/client.rs
zvolin: we agreed async on keeping the approach as is

rpc/src/client.rs
```diff
@@ -74,26 +72,231 @@ mod native {
...
+                Err(err @ ClientError::RestartNeeded(_)) => err,
+                res => return res,
+            };
+            self.mark_poisoned().await;
+            Err(err)
+        }
+    }
+
+    #[derive(Clone)]
+    struct ReusableParams(Option<String>);
```
zvolin: seems like a leftover after the implementation with loops

rpc/src/client.rs
mcrakhman: As discussed we can't do this easily due to lifetime limitations, so you approved this behaviour

Issue #858, opened 2026-01-05
Allow celestia-client and grpc-client to configure txStatus polling interval


citizen-stig:
# Story

There's a simple rollup, which just submits one blob per Celestia Block and reads data from 2 namespaces.

But its API credits usage from RPC provider is pretty costly.

Screenshot from QuickNode dashboard, tx status endpoint dominates the stat (that's per 15 minutes, with some downtime):

<img width="1501" height="572" alt="Image" src="https://github.com/user-attachments/assets/60714a61-8bae-40c4-b080-2e8085a49173" />

https://github.com/eigerco/lumina/blob/7f3509263465d3b1f9600a84a0972f98c2f9341a/grpc/src/client.rs#L903

# Suggestion

Allow client to configure polling interval. The downside is going to be that blob submission call will return with some delay. 
But considering that blob inclusion varies between 6 and 18 seconds, there's room for saving some API credits and reducing load on the RPC provider.

Ideal solution would be to have a subscription, and node notifying the client, but that will require more changes

zvolin:
cc @mcrakhman

PR #865, opened 2026-01-06
feat: implement custom confirm interval

mcrakhman:
Implement custom confirmation interval for confirm tx method: https://github.com/eigerco/lumina/issues/858

### DIFF
--- a/grpc/src/tx.rs
+++ b/grpc/src/tx.rs
@@ -111,7 +111,6 @@ impl TxConfig {
         self.confirmation_interval_ms = confirmation_interval_ms;
         self
     }
-
 }
 
 impl Default for TxConfig {


grpc/src/tx.rs
```diff
@@ -95,6 +105,29 @@ impl TxConfig {
...
+    /// Specify the confirmation polling interval in milliseconds.
+    pub fn with_confirmation_interval_ms(mut self, confirmation_interval_ms: u64) -> Self {
+        self.confirmation_interval_ms = confirmation_interval_ms;
+        self
+    }
+
+    /// Return the confirmation polling interval in milliseconds.
+    pub fn confirmation_interval_ms(&self) -> u64 {
+        self.confirmation_interval_ms
+    }
```
zvolin: nit, getter is not really needed as the field is pub and Copy

PR #873, opened 2026-01-14
feat: tx client v2

mcrakhman:
# Overview

This note summarizes the logic implemented in `grpc/src/tx_client_v2.rs` and
`grpc/src/tx_client_impl.rs`.

This is very roughly made according to the spec (with main ideas):
https://github.com/celestiaorg/celestia-app/blob/main/specs/src/tx_client_v2.md

## tx_client_v2.rs

- Provides the generic transaction pipeline split into a front-end submitter
  (`TransactionManager`) and a background worker (`TransactionWorker`).
- `TransactionManager::add_tx` serializes sequence reservation, signs the
  request via a `SignFn`, attaches one-shot callbacks, and enqueues the signed
  transaction.
- `TransactionWorker` maintains an ordered queue, tracks per-node submission
  state, and drives the lifecycle with a single event loop:
  - Submits signed transactions to nodes, respecting sequence order.
  - Polls confirmation in batches (`status_batch`) on a timer.
  - Emits `TransactionEvent` results to resolve submit/confirm callbacks.
  - Handles recoverable failures by moving to a recovery state and re-syncing
    expected sequence.
- Confirmation is in-order: confirmed info is buffered by sequence and drained
  when all prior sequences are confirmed.

## tx_client_impl.rs

- Implements the generic traits for a real gRPC client:
  - `GrpcSignContext` fetches account/chain state and gas estimates.
  - `SignFnBuilder` builds a `SignFn<Hash, TxInfo>` that creates signed tx bytes
    from `TxRequest` and `TxConfig`.
  - `TxServer for GrpcClient` wires `submit` and `status_batch` to gRPC calls and
    maps failures into v2 error types.
- Adds `TransactionService` which hides the worker plumbing:
  - Builds a `TransactionWorker` using the client signer and current sequence.
  - Exposes `submit(TxRequest, TxConfig) -> TxHandle<Hash, TxInfo>`.
  - Supports `recreate_worker()` to rebuild the worker and manager if needed.

### DIFF
--- a/grpc/src/tx_client_v2/mod.rs
+++ b/grpc/src/tx_client_v2/mod.rs
@@ -824,7 +824,6 @@ pub struct TransactionWorker<S: TxServer> {
     confirmations: FuturesUnordered<ConfirmationFuture<S>>,
     signing: Option<SigningFuture<S>>,
     confirm_ticker: Interval,
-    confirm_interval: Duration,
     max_status_batch: usize,
     should_confirm: bool,
 }
@@ -873,7 +872,6 @@ impl<S: TxServer + 'static> TransactionWorker<S> {
                 confirmations: FuturesUnordered::new(),
                 signing: None,
                 confirm_ticker: Interval::new(confirm_interval),
-                confirm_interval,
                 max_status_batch,
                 should_confirm: false,
             },
--- a/grpc/src/tx_client_v2/node_manager.rs
+++ b/grpc/src/tx_client_v2/node_manager.rs
@@ -1274,7 +1274,7 @@ mod tests {
         let mut manager = NodeManager::<DummyServer>::new(vec![node_id.clone()], Some(0));
         let txs = TestBuffer::new(Some(0));
 
-        let _outcome = manager.apply_event(
+        let outcome = manager.apply_event(
             NodeEvent::NodeResponse {
                 node_id: node_id.clone(),
                 response: NodeResponse::Submission {
@@ -1283,13 +1283,15 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         assert!(outcome.iter().any(|event| {
             matches!(
                 event,
-                WorkerMutation::MarkSubmitted { sequence, id } if *sequence == 1 && *id == 10
+                &WorkerMutation::MarkSubmitted {
+                    sequence: 1,
+                    id: 10
+                }
             )
         }));
     }
@@ -1313,13 +1315,12 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         assert!(
             !outcome
                 .iter()
-                .any(|event| matches!(event, WorkerMutation::Confirm { .. }))
+                .any(|event| matches!(event, &WorkerMutation::Confirm { .. }))
         );
     }
 
@@ -1342,7 +1343,6 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         let mut confirmed = Vec::new();
@@ -1362,7 +1362,7 @@ mod tests {
         txs.add_transaction(make_tx(1, None)).unwrap();
         txs.set_submitted_id(1, 10);
 
-        let outcome = manager.apply_event(
+        let _outcome = manager.apply_event(
             NodeEvent::NodeResponse {
                 node_id: node_id.clone(),
                 response: NodeResponse::Submission {
@@ -1373,7 +1373,6 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         let wake = manager.plan(&txs, 16, true);
@@ -1388,7 +1387,7 @@ mod tests {
     }
 
     #[test]
-    fn sequence_mismatch_missing_id_stops_on_submit() {
+    fn sequence_mismatch_missing_id_resubmits_with_delay() {
         let node_id: NodeId = Arc::from("node-1");
         let mut manager = NodeManager::<DummyServer>::new(vec![node_id.clone()], Some(0));
         let mut txs = TestBuffer::new(Some(0));
@@ -1405,15 +1404,12 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
-        assert!(matches!(
-            manager.nodes.get(&node_id).unwrap(),
-            NodeState::Stopped(_)
-        ));
-        let error = manager.tail_stop_error().expect("missing stop error");
-        assert!(matches!(error, StopError::SubmitError(_)));
+        let NodeState::Active(state) = manager.nodes.get(&node_id).unwrap() else {
+            panic!("node should be active");
+        };
+        assert!(state.shared.submit_delay.is_some());
     }
 
     #[test]
@@ -1443,7 +1439,6 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         assert!(matches!(
@@ -1475,12 +1470,12 @@ mod tests {
         let mut manager = NodeManager::<DummyServer>::new(vec![node_id], Some(0));
         let txs = TestBuffer::new(Some(0));
 
-        let outcome = manager.apply_event(NodeEvent::NodeStop, &txs, Duration::from_secs(1));
+        let outcome = manager.apply_event(NodeEvent::NodeStop, &txs);
 
         assert!(
             outcome
                 .iter()
-                .any(|event| matches!(event, WorkerMutation::WorkerStop))
+                .any(|event| matches!(event, &WorkerMutation::WorkerStop))
         );
     }
 
@@ -1494,7 +1489,7 @@ mod tests {
             state.shared.signing_inflight = true;
         }
 
-        let _outcome = manager.apply_event(
+        let outcome = manager.apply_event(
             NodeEvent::NodeResponse {
                 node_id: node_id.clone(),
                 response: NodeResponse::Signing {
@@ -1503,7 +1498,6 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         assert!(outcome.iter().any(
@@ -1529,7 +1523,7 @@ mod tests {
             state.shared.signing_inflight = true;
         }
 
-        let outcome = manager.apply_event(
+        let _outcome = manager.apply_event(
             NodeEvent::NodeResponse {
                 node_id: node_id.clone(),
                 response: NodeResponse::Signing {
@@ -1541,10 +1535,12 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(3),
         );
 
-        assert!(manager.signing_delay.is_some());
+        let NodeState::Active(state) = manager.nodes.get(&node_id).unwrap() else {
+            panic!("node should be active");
+        };
+        assert!(state.shared.signing_delay.is_some());
         assert!(matches!(
             manager.nodes.get(&node_id).unwrap(),
             NodeState::Active(_)
@@ -1561,7 +1557,7 @@ mod tests {
             state.shared.confirm_inflight = true;
         }
 
-        let _outcome = manager.apply_event(
+        let outcome = manager.apply_event(
             NodeEvent::NodeResponse {
                 node_id: node_id.clone(),
                 response: NodeResponse::Confirmation {
@@ -1572,7 +1568,6 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         assert!(outcome.is_empty());
@@ -1601,7 +1596,7 @@ mod tests {
             state.shared.last_submitted = Some(7);
         }
 
-        let outcome = manager.apply_event(
+        let _outcome = manager.apply_event(
             NodeEvent::NodeResponse {
                 node_id: node_a.clone(),
                 response: NodeResponse::Submission {
@@ -1610,7 +1605,6 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(2),
         );
 
         let NodeState::Active(state) = manager.nodes.get(&node_a).unwrap() else {
@@ -1649,7 +1643,6 @@ mod tests {
                 },
             },
             &txs,
-            Duration::from_secs(1),
         );
 
         let NodeState::Active(state) = manager.nodes.get(&node_id).unwrap() else {
@@ -1664,7 +1657,7 @@ mod tests {
         let mut manager = NodeManager::<DummyServer>::new(vec![node_id.clone()], None);
         let txs = TestBuffer::new(None);
 
-        let _ = manager.apply_event(NodeEvent::NodeStop, &txs, Duration::from_secs(1));
+        let _ = manager.apply_event(NodeEvent::NodeStop, &txs);
 
         let NodeState::Stopped(state) = manager.nodes.get(&node_id).unwrap() else {
             panic!("node should be stopped");
@@ -1678,7 +1671,7 @@ mod tests {
         let mut manager = NodeManager::<DummyServer>::new(vec![node_id.clone()], Some(5));
         let txs = TestBuffer::new(Some(5));
 
-        let _ = manager.apply_event(NodeEvent::NodeStop, &txs, Duration::from_secs(1));
+        let _ = manager.apply_event(NodeEvent::NodeStop, &txs);
 
         let NodeState::Stopped(state) = manager.nodes.get(&node_id).unwrap() else {
             panic!("node should be stopped");
@@ -1696,7 +1689,7 @@ mod tests {
             state.shared.last_submitted = Some(7);
         }
 
-        let _ = manager.apply_event(NodeEvent::NodeStop, &txs, Duration::from_secs(1));
+        let _ = manager.apply_event(NodeEvent::NodeStop, &txs);
 
         let NodeState::Stopped(state) = manager.nodes.get(&node_id).unwrap() else {
             panic!("node should be stopped");


grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+use async_trait::async_trait;
+use celestia_types::state::ErrorCode;
+use tendermint_proto::google::protobuf::Any;
+use tokio::sync::{Mutex, Notify, mpsc, oneshot};
+use tokio::task::JoinSet;
+use tokio::time;
+
+use crate::{Error, Result, TxConfig};
+/// Identifier for a submission/confirmation node.
+pub type NodeId = String;
```
zvolin: since we do a lot of cloning on this, let's make it `Arc<str>`

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+}
+
+#[derive(Debug)]
+pub struct TxHandle<TxId: TxIdT, ConfirmInfo> {
+    /// Sequence reserved for this transaction.
+    pub sequence: u64,
+    /// Receives submit result.
+    pub on_submit: oneshot::Receiver<Result<TxId>>,
+    /// Receives confirm result.
+    pub on_confirm: oneshot::Receiver<Result<ConfirmInfo>>,
```
zvolin: nit, `on_submit` etc sounds better on `TxCallbacks` as it is called when manager submits and confirms txs. Wondering if here `submitted` and `confirmed` wouldn't be better names used like `tx.submitted.await`

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+    },
+    UnknownTransaction {
+        error_code: ErrorCode,
+        message: String,
+    },
+}
+
+#[async_trait]
+pub trait TxServer: Send + Sync {
+    type TxId: TxIdT + Clone + Eq + StdHash + Send + Sync + 'static;
```
zvolin: `TxIdT` already enforces `Clone`

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+    Submitting,
+    Stopped(StopReason),
+}
+
+impl ProcessorState {
+    fn update(&mut self, other: ProcessorState) {
+        match (&self, other) {
+            (ProcessorState::Stopped(_stop_reason), _) => {}
+            (ProcessorState::Recovering { .. }, ProcessorState::Stopped(stop_reason)) => {
+                *self = ProcessorState::Stopped(stop_reason);
```
zvolin:
```suggestion
            (ProcessorState::Recovering { .. }, stopped @ ProcessorState::Stopped(_)) => {
                *self = stopped;
```

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+        self.new_submit.notify_one();
+        Ok(())
+    }
+
+    /// Run the worker loop until shutdown or a terminal error.
+    ///
+    /// # Notes
+    /// - The worker is single-owner; it should be run in a dedicated task.
+    /// - On terminal failures it returns `Ok(())` after notifying callbacks.
+    pub async fn process(&mut self, shutdown: Arc<Notify>) -> Result<()> {
```
zvolin: I think for shutdown I'd prefer a cancelation token. In `lumina_utils` there's a cancelation token and cross-platform `spawn_cancellable` which could be used to also cancel all the pending futures.

Also I'm wondering why is shutdown notifier accepted here and not a property on self?

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+    events: VecDeque<TransactionEvent<S::TxId, S::ConfirmInfo>>,
+    confirmed_info: HashMap<u64, S::ConfirmInfo>,
+    confirmed_sequence: u64,
+    next_enqueue_sequence: u64,
+    node_state: HashMap<NodeId, NodeSubmissionState>,
+    add_tx_rx: mpsc::Receiver<Transaction<S::TxId, S::ConfirmInfo>>,
+    new_event: Arc<Notify>,
+    new_submit: Arc<Notify>,
+    submissions: JoinSet<SubmissionResult<S::TxId>>,
+    confirmations: JoinSet<ConfirmationResult<S::TxId, S::ConfirmInfo>>,
```
zvolin: as we discussed, this will need to be `FuturesUnordered` to work in wasm

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+                            },
+                        )));
+                }
+                TxStatus::Rejected { reason: _reason } => {
+                    self.state
+                        .update(ProcessorState::Stopped(StopReason::ConfirmFailure(
+                            ConfirmFailure {
+                                reason: recovery_reason.clone(),
+                            },
+                        )));
```
zvolin: I'm not sure I understand, why do we stop on eviction and rejection (regardless of reason) during recovery?

grpc/src/tx_client_v2/mod.rs
```diff
@@ -0,0 +1,1645 @@
...
+            });
+        }
+    }
+
+    fn spawn_confirmations(&mut self, max_seq: Option<u64>, node_filter: Option<&NodeId>) {
+        if self.tx_index.is_empty() {
+            return;
+        }
+        // gathering all the sequences for each node
+        for (node_id, state) in self.node_state.iter_mut() {
```
zvolin: why do we iter mut there? I havent' noticed any mutation

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+        let start = self.confirmed_sequence.saturating_add(1);
+        if sequence < start {
+            return None;
+        }
+        let offset = sequence - start;
+        if offset >= self.txs.len() as u64 {
+            return None;
+        }
+        Some(offset as usize)
+    }
```
zvolin: this `self.txs` and `self.tx_index` is a bit confusing to me, what's the advantage of it rather than just having `HashMap<sequence, tx>`? If you're looking for the hashmap which allows iteration in insertion order, removing from front and back, then `indexmap` is a great crate, kinda if `HashMap` and `VecDeque` had a child

grpc/src/tx_client_v2.rs
```diff
@@ -0,0 +1,1645 @@
...
+                }
+            }
+        }
+    }
+
+    fn spawn_submissions(&mut self) {
+        if self.nodes.is_empty() {
+            return;
+        }
+        for (node_id, node) in self.nodes.clone() {
```
zvolin: do we need this clone? if node_id and node are both in Arc's, it'd be cheaper to call `.cloned()` on the iterator itself rather than cloning the hashmap

grpc/src/tx_client_v2.rs
mcrakhman: So we recover if the server advanced past the client's state. This can happen if:
- we are using some parallel client which submits different transactions, or
- some other node included our transactions, even though the current node evicted them for whatever reason

So to be able to distinguish between these two options, we actually need to confirm only our transaction with this sequence. Because if the node said that it has sequence number, then our transaction with same sequence either should be confirmed or should be rejected.

mcrakhman: I will simplify this

grpc/src/tx_client_v2/mod.rs
mcrakhman:
```rust
for (_, state) in self.node_state.iter_mut() {
   state.submitted_seq = state.submitted_seq.max(sequence);
}
```

grpc/src/tx_client_v2.rs
mcrakhman: the problem is that we also need to know `tx -> sequence` because we don't get this info from the confirmation response. And I don't want to rely on the fact that the server returns hashes in the sequence order, unless sequence is the part of the response

grpc/src/tx_client_impl.rs
```diff
@@ -0,0 +1,678 @@
...
+            context: self.context,
+            pubkey: self.pubkey,
+            signer: self.signer,
+            cached_account: OnceCell::new(),
+            cached_chain_id: OnceCell::new(),
+        })
+    }
+}
+
+pub struct TransactionService {
```
zvolin, line 85: derive `Clone` and implement `Debug` on it

grpc/src/tx_client_impl.rs
```diff
@@ -0,0 +1,678 @@
...
+        .await?;
+
+        let mut worker_guard = self.inner.worker.lock().await;
+        if let Some(old_worker) = worker_guard.replace(worker_handle) {
+            old_worker.shutdown.cancel();
+            // Wait for worker to finish
+            let _ = old_worker.done_rx.await;
+        }
+
+        let mut manager_guard = self.inner.manager.write().await;
```
zvolin: I think we should lock manager before worker

grpc/src/tx_client_v2/mod.rs
```diff
@@ -0,0 +1,1788 @@
...
+            let tx = push_oneshot(&mut self.submissions);
+            spawn_cancellable(token.clone(), async move {
+                time::sleep(delay).await;
+                let result = node.submit(bytes.clone(), target_sequence).await;
+                let _ = tx.send(SubmissionResult {
+                    node_id,
+                    sequence: target_sequence,
+                    result,
+                });
+            });
```
zvolin: wdyt about expanding on this helper method, we could maybe do sth like
```rs
fn spawn_task<T, F, Fut>(
    unordered: &mut FuturesUnordered<BoxFuture<'static, Option<T>>>,
    token: CancellationToken,
    fut_fn: F,
) where
    T: Send + 'static,
    F: FnOnce(oneshot::Sender<T>) -> Fut,
    Fut: Future<Output = ()> + Send + 'static,
{
    let (tx, rx) = oneshot::channel();

    spawn_cancellable(token, fut_fn(tx));

    unordered.push(async move { rx.await.ok() }.boxed());
}
```

and use it like
```rs
spawn_task(&mut self.submissions, token.clone(), |res_tx| async move {
    time::sleep(delay).await;
    let result = node.submit(bytes.clone(), target_sequence).await;
    let _ = res_tx.send(SubmissionResult {
        node_id,
        sequence: target_sequence,
        result,
    });
});
```

grpc/src/tx_client_v2/mod.rs
```diff
@@ -0,0 +1,1788 @@
...
+                    Ok::<_, Error>(())
+                } => res?,
+            }
+            if self.state.is_stopped() {
+                break;
+            }
+        }
+
+        // Wait for all in-flight tasks to complete before returning
+        self.drain_pending().await;
```
zvolin: I believe this is no longer needed as we have cancellation now

grpc/src/tx_client_impl.rs
```diff
@@ -0,0 +1,678 @@
...
+            return match parsed {
+                Ok(expected) => SubmitFailure::SequenceMismatch { expected },
+                Err(err) => SubmitFailure::NetworkError { err: Arc::new(err) },
+            };
+        }
+    }
+
+    match code {
+        ErrorCode::InsufficientFunds => SubmitFailure::InsufficientFunds,
+        ErrorCode::InsufficientFee => SubmitFailure::InsufficientFee { expected_fee: 0 },
```
zvolin: should we parse expected fee out of the message?

grpc/src/tx_client_v2/mod.rs
```diff
@@ -301,12 +275,11 @@ pub enum TxStatus<ConfirmInfo> {
...
-    InvalidTx { error_code: ErrorCode },
-    /// Account has insufficient funds.
-    InsufficientFunds,
-    /// Fee too low for the computed gas price.
-    InsufficientFee { expected_fee: u64 },
+    /// Submission failed with a specific error code and message.
+    Other {
+        error_code: ErrorCode,
+        message: String,
+    },
```
zvolin: nit, I'd like `Other` to be last variant

grpc/src/grpc/celestia_tx.rs
```diff
@@ -120,8 +140,28 @@ impl IntoGrpcParam<RawTxStatusRequest> for Hash {
     }
 }
 
+impl IntoGrpcParam<RawTxStatusBatchRequest> for Vec<Hash> {
```
zvolin: can we implement it for `IntoIterator<Item = Hash>`?

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+use tracing::debug;
+
+use crate::tx_client_v2::tx_buffer::TxBuffer;
+use crate::tx_client_v2::{
+    NodeEvent, NodeId, NodeResponse, RejectionReason, SubmitFailure, TxIdT, TxServer, TxStatus,
+    WakeUpKind,
+};
+
+#[derive(Debug)]
+pub(crate) enum WorkerEvent<TxId: TxIdT, ConfirmInfo> {
```
zvolin: I think it'd be better to have WorkerEvent defined near worker in `mod.rs` and `NodeEvent` here

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+                format!("ConfirmError({})", tx_status_label(status))
+            }
+            StopError::SubmitError(status) => format!("SubmitError({})", tx_status_label(status)),
+            StopError::Other => "Other".to_string(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct NodeShared {
```
zvolin: maybe `CommonState`?

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+
+    fn is_stopped(&self) -> bool {
+        matches!(self, NodeState::Stopped(_))
+    }
+
+    fn epoch(&self) -> u64 {
+        self.shared().epoch
+    }
+
+    fn confirm_inflight(&self) -> bool {
```
zvolin: since it's present in each state, shouldn't it go to shared?

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+            StopError::ConfirmError(status) => {
+                format!("ConfirmError({})", tx_status_label(status))
+            }
+            StopError::SubmitError(status) => format!("SubmitError({})", tx_status_label(status)),
+            StopError::Other => "Other".to_string(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
```
zvolin: this can be `Copy` and then the `.shared()` doesn't need to return the ref

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+        let confirm_inflight = self.confirm_inflight();
+        let stop_seq = last_submitted
+            .unwrap_or(shared.last_confirmed)
+            .max(shared.last_confirmed);
+        *self = NodeState::Stopped(StoppedState {
+            shared,
+            confirm_inflight,
+            stop_seq,
+            error: StopError::Other,
+        });
```
zvolin:
```suggestion
        let last_submitted = self.shared().last_submitted;
        let stop_seq = last_submitted
            .unwrap_or(self.shared().last_confirmed)
            .max(self.shared().last_confirmed);
        self.stop_with_error(stop_seq, StopError::Other);
```

grpc/src/tx_client_v2/mod.rs
```diff
@@ -0,0 +1,712 @@
...
+            }
+            .boxed(),
+        );
+        self.confirmations.push(
+            async move {
+                confirm_shutdown.cancelled().await;
+                None
+            }
+            .boxed(),
+        );
```
zvolin: > pushing pending futures to avoid busy loop

A more elegant way to avoid busy loop would be to just match the pattern in select, like
```rust
select! {
    Some(result) = self.submissions.next() => { ... }
}
```

Iirc, if the pattern doesn't match (in this case, no tasks would give back `None`) then all remaining futures will continue to be polled until some of them is ready

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+        let mut min_confirmed: Option<u64> = None;
+        for node in self.nodes.values() {
+            if node.is_stopped() {
+                continue;
+            }
+            let confirmed = node.shared().last_confirmed;
+            min_confirmed = Some(match min_confirmed {
+                Some(current) => current.min(confirmed),
+                None => confirmed,
+            });
```
zvolin:
```suggestion
            min_confirmed = min_confirmed
                .map(|current| current.min(confirmed))
                .unwrap_or(confirmed);
```

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+        min_confirmed
+    }
+
+    pub(crate) fn tail_error_status(&self) -> Option<TxStatus<S::ConfirmInfo>> {
+        let mut best: Option<(u64, TxStatus<S::ConfirmInfo>)> = None;
+        for node in self.nodes.values() {
+            if let NodeState::Stopped(state) = node {
+                let Some(status) = state.error.status() else {
+                    continue;
+                };
```
zvolin:
```suggestion
            if let NodeState::Stopped(state) = node &&
                let Some(status) = state.error.status() {
```

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+        event: NodeEvent<S::TxId, S::ConfirmInfo>,
+        txs: &TxBuffer<S::TxId, S::ConfirmInfo>,
+        confirm_interval: Duration,
+        max_status_batch: usize,
+    ) -> NodeOutcome<S::TxId, S::ConfirmInfo> {
+        let mut worker_events = Vec::new();
+        let mut wake_up = None;
+
+        debug!(event = %event.summary(), "node event");
+        match event {
```
zvolin, line 361: this match is touch too complex and nested, maybe we could extract some of the logic to helper methods, like `on_wakeup` `on_submission_response` `on_confirmation_response`

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+                            match node {
+                                NodeState::Active(active) => active.confirm_inflight = false,
+                                NodeState::Recovering(recovering) => {
+                                    recovering.confirm_inflight = false;
+                                }
+                                NodeState::Stopped(stopped) => {
+                                    stopped.confirm_inflight = false;
+                                }
+                            }
+                            if wake_up.is_none() {
```
zvolin: here and below I believe wake_up must always be none

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+        },
+    }
+}
+
+fn on_submit_success(state: &mut ActiveState, sequence: u64) {
+    state.pending = None;
+    match state.shared.last_submitted {
+        None => state.shared.last_submitted = Some(sequence),
+        Some(last) => {
+            if last + 1 == sequence {
```
zvolin, line 690: I think this should be debug_assert, no?

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+            let next_seq = match state.shared.last_submitted {
+                Some(last) => last.max(confirmed) + 1,
+                None => confirmed + 1,
+            };
+            let tx = txs.get(next_seq)?;
+            let delay = state
+                .shared
+                .submit_delay
+                .take()
+                .unwrap_or_else(|| Duration::from_nanos(0));
```
zvolin:
```suggestion
                .unwrap_or_default();
```

grpc/src/tx_client_v2/node_manager.rs
```diff
@@ -0,0 +1,1209 @@
...
+        NodeState::Active(state) => &mut state.shared,
+        NodeState::Stopped(state) => &mut state.shared,
+        NodeState::Recovering(_) => {
+            return false;
+        }
+    };
+    if txs.confirmed_seq() >= expected {
+        return false;
+    }
+    if sequence < expected {
```
zvolin, line 926: some comments what those ifs mean would be useful

zvolin (approved):

PR #874, opened 2026-01-17
feat: use plaintext key for validator

mcrakhman:
There exists a bug that prevents `celestia-app` correctly decrypting the encrypted test key (`validator-0.key`) using pre-defined password when running on MacOS in Docker (docker runs amd64 binaries using Rosetta2 translation layer which presumably has some problems when decrypting the key using AVX2 instructions, the problem goes away when we compile using `purego`). 

Easiest fix is to use the `.plaintext-key` which is already in the repo.

### DIFF
--- a/ci/run-validator.sh
+++ b/ci/run-validator.sh
@@ -93,7 +93,7 @@ create_or_import_key() {
   fi
 
   # otherwise, just import it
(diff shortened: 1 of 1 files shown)


zvolin (commented):
looks good, but I think the same fix is needed in `ci/run-node.sh` for the `cel-key` key import, no?

PR #875, opened 2026-01-17
feat(grpc)!: add per-endpoint configuration for metadata and timeouts

mcrakhman:
**Related issue:** Closes https://github.com/eigerco/lumina/issues/833 (follow-up)

## Summary

This PR extends the multi-endpoint fallback mechanism (introduced in #836) to support **per-endpoint configuration**. Each gRPC endpoint can now have its own metadata (e.g., auth tokens) and timeout settings, enabling scenarios where different endpoints require different credentials.

The API is **backward compatible** - the simple `.url()` API remains unchanged, while new `*_with_config()` variants are added for per-endpoint configuration.

### `EndpointConfig` Type (New)

Configuration for individual endpoints supporting:
- **ASCII metadata** - Auth tokens, request IDs, text headers
- **Binary metadata** - Signatures, serialized protos (keys must have `-bin` suffix)
- **Timeout** - Per-endpoint request timeout

### Rust Core (`celestia-grpc`)

**GrpcClientBuilder methods:**

| Method | Description |
|--------|-------------|
| `.url(url)` | Add URL with default config |
| `.url_with_config(url, config)` | Add URL with EndpointConfig |
| `.urls([urls])` | Add multiple URLs with default configs |
| `.urls_with_config([(url, config), ...])` | Add multiple URLs with configs |

**EndpointConfig methods:**

| Method | Description |
|--------|-------------|
| `EndpointConfig::new()` | Create empty config |
| `.metadata(key, value)` | Add ASCII metadata |
| `.metadata_bin(key, value)` | Add binary metadata |
| `.metadata_map(map)` | Add pre-built MetadataMap |
| `.timeout(duration)` | Set request timeout |

## Migration Guide

**No breaking changes for basic usage.** The simple API remains the same:

```rust
// This still works exactly as before
let client = GrpcClient::builder()
    .url("http://localhost:9090")
    .build()?;
```

For users who were using global metadata/timeout on the builder (if any), these should now be moved to `EndpointConfig`:

```rust
// Before (if global methods existed)
// let client = GrpcClient::builder()
//     .url("http://localhost:9090")
//     .metadata("auth", "token")  // global
//     .build()?;

// After
let client = GrpcClient::builder()
    .url_with_config("http://localhost:9090", EndpointConfig::new()
        .metadata("auth", "token"))  // per-endpoint
    .build()?;
```

## Implementation Note: `Context::extend()` timeout semantics

The `extend()` method's timeout behavior was changed from `min(t0, t1)` to "per-request wins":

```rust
// Before
(Some(t0), Some(t1)) => Some(t0.min(t1))

// After
(Some(_), Some(t1)) => Some(t1)
```

**Rationale:** This preserves the existing behavior where per-request `.timeout()` overrides endpoint timeouts.

Previously, context merging via `.context(&self.inner.context)` happened *before* the user could call `.timeout()`. Since `.timeout()` does direct assignment (not `extend()`), it would overwrite the merged value. The `min()` logic was effectively dead code for this path.

With per-endpoint contexts, merging now happens *inside* the async closure, *after* `.timeout()` has already set the call context. Both the transport context and call context have values when `extend()` is called. To preserve the same effective behavior (per-request timeout wins), we changed `extend()` to prefer the "other" (per-request) timeout.

### DIFF
--- a/grpc/src/uniffi_client.rs
+++ b/grpc/src/uniffi_client.rs
@@ -180,29 +180,6 @@ impl GrpcClientBuilder {
         GrpcClientBuilder(Mutex::new(Some(builder)))
     }
 
-    /// Add another URL endpoint.
-    #[uniffi::method(name = "addUrl")]
-    pub fn add_url(self: Arc<Self>, url: String) -> Arc<Self> {
-        self.map_builder(move |builder| builder.url(url));
-        self
-    }
-
-    /// Add another endpoint.
-    #[uniffi::method(name = "endpoint")]
-    pub fn endpoint(self: Arc<Self>, endpoint: Arc<Endpoint>) -> Arc<Self> {
-        let endpoint = endpoint.take();
-        self.map_builder(move |builder| builder.endpoint(endpoint));
-        self
-    }
-
-    /// Add multiple endpoints.
-    #[uniffi::method(name = "endpoints")]
-    pub fn endpoints(self: Arc<Self>, endpoints: Vec<Arc<Endpoint>>) -> Arc<Self> {
-        let endpoints = endpoints.into_iter().map(|e| e.take()).collect::<Vec<_>>();
-        self.map_builder(move |builder| builder.endpoints(endpoints));
-        self
-    }
-
     /// Add public key and signer to the client being built
     #[uniffi::method(name = "withPubkeyAndSigner")]
     pub fn pubkey_and_signer(


grpc/src/js_client.rs
```diff
@@ -78,11 +191,30 @@ impl GrpcClientBuilder {
...
+    ///
+    /// Note that this method **consumes** builder and returns updated instance of it.
+    /// Make sure to re-assign it if you keep builder in a variable.
+    #[wasm_bindgen(js_name = "withUrlWithConfig")]
+    pub fn with_url_with_config(self, url: String, config: EndpointConfig) -> Self {
+        Self {
+            inner: self.inner.url_with_config(url, config.inner),
+        }
+    }
+
```
zvolin: for more idiomatic javascript, I think you could just keep only `.withUrl` and on rust side take an `Option`, like

```rs
pub fn with_url(self, url: String, config: Option<EndpointConfig>) -> Self {
```

This way, js users will be able to call it with optional arguments

```js
await new GrpcClientBuilder().withUrl("https://foo").build()
await new GrpcClientBuilder().withUrl("https://foo", new EndpointConfig()).build()
```

grpc/src/builder.rs
```diff
@@ -21,8 +21,107 @@ use crate::{DocSigner, GrpcClient, GrpcClientBuilderError};
...
+    fn from(url: String) -> Self {
+        Endpoint::new(url)
+    }
+}
+
+impl From<&str> for Endpoint {
+    fn from(url: &str) -> Self {
+        Endpoint::new(url)
+    }
+}
```
zvolin: can we do `impl<S> From<S> for Endpoint where S: AsRef<str>`? or does it somehow conflict?

zvolin: or maybe where `S: Into<String>` :thinking:

grpc/src/js_client.rs
```diff
@@ -11,16 +12,94 @@ mod grpc_client;
...
 /// Builder for [`GrpcClient`] and [`TxClient`].
 ///
 /// Url must point to a [grpc-web proxy](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md).
 ///
 /// # Keyless client example
 ///
 /// ```js
-/// const client = await GrpcClient
+/// const client = await new GrpcClientBuilder()
```
zvolin: are you sure this example works? `GrpcClientBuilder()` doesn't have constructor

grpc/src/js_client/grpc_client.rs
```diff
@@ -383,6 +405,18 @@ impl GrpcClient {
     }
 }
 
+fn endpoints_from_js_or_throw(value: JsValue) -> Vec<crate::Endpoint> {
```
zvolin: the `or_throw` is not really needed, any function that returns `Result` will throw in javascript if it return `Err`, so you can just write regular rust code with errors and don't worry about explicit throwing

grpc/src/uniffi_client.rs
```diff
@@ -63,23 +123,86 @@ impl GrpcClientBuilder {
...
+    #[uniffi::constructor(name = "withEndpoints")]
+    pub fn with_endpoints(endpoints: Vec<Arc<Endpoint>>) -> Self {
+        let endpoints = endpoints.into_iter().map(|e| e.take()).collect::<Vec<_>>();
+        let builder = RustBuilder::new().endpoints(endpoints);
+        GrpcClientBuilder(Mutex::new(Some(builder)))
+    }
+
+    /// Add another URL endpoint.
+    #[uniffi::method(name = "addUrl")]
+    pub fn add_url(self: Arc<Self>, url: String) -> Arc<Self> {
```
zvolin: I'd remove the `add` prefix to be consistent with other methods

grpc/src/js_client.rs
```diff
@@ -67,29 +146,43 @@ pub struct GrpcClientBuilder {
...
+    /// Accepts a string, an `Endpoint`, or an array of strings/endpoints.
     ///
     /// Note that this method **consumes** builder and returns updated instance of it.
     /// Make sure to re-assign it if you keep builder in a variable.
     #[wasm_bindgen(js_name = "withUrl")]
-    pub fn with_url(self, url: String) -> Self {
-        Self {
-            inner: self.inner.url(url),
-        }
+    pub fn with_url(
```
zvolin, line 156: if we want consistent javascript <-> rust api we should also add `withEndpoint` variants here and on the javascript `GrpcClient`

grpc/src/js_client.rs
```diff
@@ -67,29 +146,43 @@ pub struct GrpcClientBuilder {
...
+        self,
+        #[wasm_bindgen(unchecked_param_type = "UrlsOrEndpoints")] url: JsValue,
```
zvolin, line 158: actually this doesn't seem really needed, the only difference between this and `withUrls` is the name, but both accept either one or many urls

grpc/src/js_client.rs
mcrakhman: will fix in next pr

grpc/src/js_client.rs
mcrakhman: will fix in next pr

PR #877, opened 2026-01-24
chore(types)!: Support celestia-app V7

jonas089:
This PR adds types to support V7 of celestia-app.

### DIFF
--- a/types/src/consts.rs
+++ b/types/src/consts.rs
@@ -118,7 +118,7 @@ pub mod appconsts {
         pub const MAX_TX_SIZE: u64 = 8_388_608; // 8MB
     }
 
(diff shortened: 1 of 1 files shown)


types/src/consts.rs
```diff
@@ -118,9 +118,26 @@ pub mod appconsts {
         pub const MAX_TX_SIZE: u64 = 8_388_608; // 8MB
     }
 
+    // https://github.com/celestiaorg/celestia-app/blob/main/pkg/appconsts/app_consts.go
```
zvolin: it'd be nice to point to the tag instead

jonas089: @zvolin I pointed it to the RC tag for now.

PR #884, opened 2026-02-06
feat(node)!: switch sample and row requests to shrex

zvolin:
Switches the DA protocol from bitswap to shrex.

Currently there's no celestia-node tag to test it with, I was running it with local builds


mcrakhman:
Closing this PR in favor of the other one containing these commits and updates: https://github.com/celestiaorg/lumina/pull/976

PR #890, opened 2026-02-16
feat(rpc)!: add server trait to rpc crate

Ferret-san:
## Overview

Certain Rust based tools like [localestia](https://github.com/celestiaorg/localestia) (a celestia mock for integration testing and experimentation built with redis) need to define their own server methods and traits, which then complicates the process of compatibility testing by requiring extra test suites to ensure compatibility (i.e how can we ensure that the localestia rpc methods match the canonical ones).

By adding the `server` attribute to the different traits in the `rpc` crate and implementing some generics for the different namespaces we can ensure that people that want to customize or extend lumina can do so while having compliant RPC methods, ensuring libraries and requests are semantically consistent across the board.


rpc/src/state.rs
```diff
@@ -3,106 +3,103 @@ use celestia_types::state::{
...
-    async fn state_account_address(&self) -> Result<Address, Error>;
+    async fn state_account_address(&self) -> RpcResult<Address>;
 
-    /// Balance retrieves the Celestia coin balance for the node's account/signer and verifies it against the corresponding block's AppHash.
+    /// See [`crate::StateClient::state_balance`].
     #[method(name = "Balance")]
-    async fn state_balance(&self) -> Result<Coin, Error>;
+    async fn state_balance(&self) -> RpcResult<Coin>;
 
-    /// BalanceForAddress retrieves the Celestia coin balance for the given address and verifies the returned balance against the corresponding block's AppHash.
```
mcrakhman, line 20: Please revive the removed comments

rpc/src/state.rs
```diff
@@ -3,106 +3,103 @@ use celestia_types::state::{
...
-    ///
-    /// # NOTE
-    ///
-    /// The balance returned is the balance reported by the block right before the node's current head (head-1). This is due to the fact that for block N, the block's `AppHash` is the result of applying the previous block's transaction list.
+    /// See [`crate::StateClient::state_balance_for_address`].
     #[method(name = "BalanceForAddress")]
-    async fn state_balance_for_address(&self, addr: &Address) -> Result<Coin, Error>;
+    async fn state_balance_for_address(&self, addr: Address) -> RpcResult<Coin>;
+    //
```
mcrakhman: Remove empty comments

rpc/src/state.rs
```diff
@@ -3,106 +3,103 @@ use celestia_types::state::{
...
     #[method(name = "Balance")]
-    async fn state_balance(&self) -> Result<Coin, Error>;
+    async fn state_balance(&self) -> RpcResult<Coin>;
 
-    /// BalanceForAddress retrieves the Celestia coin balance for the given address and verifies the returned balance against the corresponding block's AppHash.
-    ///
-    /// # NOTE
-    ///
-    /// The balance returned is the balance reported by the block right before the node's current head (head-1). This is due to the fact that for block N, the block's `AppHash` is the result of applying the previous block's transaction list.
+    /// See [`crate::StateClient::state_balance_for_address`].
```
mcrakhman: Also this refs won't work because afaiu StateClient is also generated and without docs, so here we need to have normal comments

PR #933, opened 2026-03-03
feat: implement rsema1d

mcrakhman:
## Summary

Rust implementation of the `rsema1d` codec:

- Reed-Solomon row extension (`k` original rows + `n` parity rows)
- Merkle commitment over rows
- RLC-based row verification
- Reconstruction of original rows from any `k` available rows

## Docs/benchmarks

- End-to-end usage documentation in README with runnable API example.
- Compatibility tooling:
  - `scripts/run_go_compat.sh` (Go fuzzy vectors + Rust compat verification)
- Benchmark tooling:
  - `scripts/run_benchmarks.sh` (Rust/Go bench runner)

### Rust Criterion (`encode`)

- `128KB_k1024_n3072`: ~`25.6 MiB/s`
- `1MB_k1024_n3072`: ~`67.0 MiB/s`
- `1MB_k4096_n12288`: ~`47.5 MiB/s`
- `8MB_k4096_n12288`: ~`70.8 MiB/s`
- `128MB_k4096_n12288`: ~`74.1 MiB/s`
- `128MB_k8192_n24576`: ~`71.9 MiB/s`

### Rust Criterion (`encode_in_place`)

- `128KB_k1024_n3072`: ~`29.5 MiB/s`
- `1MB_k1024_n3072`: ~`98.4 MiB/s`
- `1MB_k4096_n12288`: ~`61.1 MiB/s`
- `8MB_k4096_n12288`: ~`81.8 MiB/s`
- `128MB_k4096_n12288`: ~`90.1 MiB/s`
- `128MB_k8192_n24576`: ~`86.9 MiB/s`

### Go benchmark (`Encode_*`, GOMAXPROCS=16)

- `128KB_k1024_n3072`: `16.39 MiB/s`
- `1MB_k1024_n3072`: `78.94 MiB/s`
- `1MB_k4096_n12288`: `35.00 MiB/s`
- `8MB_k4096_n12288`: `90.53 MiB/s`
- `128MB_k4096_n12288`: `107.26 MiB/s`
- `128MB_k8192_n24576`: `103.91 MiB/s`

## Performance Note

`encode` is currently a bit slower on larger sizes.  
The likely reason is that the Leopard encoding library path is slower at large shard sizes (at least on the Intel Xeon hardware used on validator machines).

`encode_in_place` improves throughput compared to `encode`, but large-size encode throughput is still below Go in the provided run.

### DIFF
--- a/rsema1d/go/cmd/generate_fuzzy_vectors/main.go
+++ b/rsema1d/go/cmd/generate_fuzzy_vectors/main.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/celestiaorg/rsema1d"
+	"github.com/celestiaorg/rsema1d/field"
 )
 
 type TestVector struct {
@@ -18,12 +19,14 @@ type TestVector struct {
 	Seed   int64      `json:"seed"`
 
 	OriginalData []string `json:"original_data"` // Hex-encoded original rows
+	RLCOrig      []string `json:"rlc_orig"`      // Hex-encoded original RLC values (16 bytes each)
 
 	// Public outputs
 	Commitment string `json:"commitment"` // The final commitment
 
 	// Proof tests
-	ProofTests []ProofTest `json:"proof_tests"`
+	ProofTests        []ProofTest        `json:"proof_tests"`
+	RowInclusionTests []RowInclusionTest `json:"row_inclusion_tests"`
 }
 
 type TestParams struct {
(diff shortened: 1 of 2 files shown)


rach-id (commented):
Asked claude to check the fuzzing to ensure that we have equivalence between the two implementation and got this feedback:
```
  1. Deserialized fields are incomplete — The Rust ProofTest struct is missing several Go-emitted fields:
  - is_original — not deserialized (serde will silently ignore it with #[serde(deny_unknown_fields)] absent, which is fine, but it's a missed validation opportunity)
  - row_hash, row_proof_hash, rlc_proof_hash — completely ignored. These hash digests could be used to verify that Rust produces identical proof nodes as Go, not just that Go's proofs verify under Rust. Currently, the test only
   checks Go→Rust verification, not byte-level proof equivalence.

  2. No Rust→Go proof comparison — The test verifies Go's proofs using Rust's verifier, but never checks that Rust's generate_row_proof(idx) or generate_standalone_proof(idx) produce the same proof nodes as Go. If the Merkle
  tree construction differs in padding/ordering but both are internally consistent, this test would pass while the implementations are incompatible. Adding explicit proof-node comparison (or checking against the hash digests)
  would catch this.

  3. Reconstruction only tests trivial case — The Rust compat test reconstructs from indices 0..k (the original rows). This is a no-op for RS decoding since no erasure recovery is needed. The Go fuzzer doesn't emit
  reconstruction vectors at all. Cross-compat of actual erasure recovery (from mixed original+parity rows) is entirely untested.

  4. verify_with_context for parity rows depends on Rust-computed RLC — For non-standalone proofs, the Rust test creates VerificationContext from Rust's own rlc_orig. This means if Go's RLC computation differed from Rust's, the
   test wouldn't catch it for parity rows (it would fail for standalone proofs though, since those embed the RLC proof from Go).
```

Looks legit to me, but I defer to you if they're useful

PR #954, opened 2026-04-15
feat: add fibre and valaddr proto definitions

mcrakhman:
## Summary (part of the fibre pr split)
- Add protobuf definitions for the `celestia.fibre.v1` and `celestia.valaddr.v1` modules (message types, query/tx services, genesis, params, events)
- Update `proto/build.rs` to compile the new packages and add the `fix_grpc_service_paths` post-processing step for tonic
- Rename `go-square/blob/v2` to `v4` and update `types/src/blob.rs` imports accordingly
- Vendor the CometBFT `SignRawBytesRequest` proto (`tendermint-celestia-mods/privval/types.proto`) needed for validator signature domain separation


Cargo.toml
```diff
@@ -97,6 +97,10 @@ wasm-bindgen-test = "0.3"
 web-sys = "0.3.70"
 
 [patch.crates-io]
+# All versions of core2 have been yanked from crates.io, which breaks
+# `cargo minimal-versions`. Patch it from git until cid drops the dependency.
+core2 = { git = "https://github.com/bbqsrc/core2", rev = "545e84bc" }
```
rootulp, line 102: @claude maybe core2 was yanked from crates.io for a good reason? Can you investigate

rootulp (approved):
LGTM

[not blocking] instead of vendoring the Fibre protos, is it possible to import them from buf after https://github.com/celestiaorg/celestia-app/pull/7007

Cargo.toml
mcrakhman: I already checked it, it is just not supported by author, because repo is archived, we can remove it after the repos that use it will upgrade:

The situation:                                                                                                                                                           
  - cid 0.11.1 and multihash 0.19.3 both have core2 ^0.4 as a non-optional dependency                                                                                      
  - No newer versions of either crate exist
  - Active PRs upstream replace core2 with no_std_io2 (API-compatible fork): https://github.com/multiformats/rust-multihash/pull/407,
  https://github.com/multiformats/rust-cid/pull/185 — not merged yet

PR #960, opened 2026-04-15
ci!: remove iOS, Android, and uniffi binding CI jobs

mcrakhman:
## Summary

Remove iOS, Android, and UniFFI binding CI and release jobs. The only production consumer of the mobile binaries was **Leap wallet**, which is [sunsetting on May 28, 2026](https://www.leapwallet.io/sunset). Investigation confirmed no other production consumers exist (71 total binary downloads across all releases, 0 crates.io reverse dependencies, and only 2 external GitHub repos — both abandoned PoCs by the same author).

### What's removed
- `build-ios`, `build-android`, and `test-bindings` CI jobs
- `lumina-node-uniffi` from `clippy`, `test`, and `msrv` matrices
- UniFFI artifact builds from the release workflow
- Corresponding entries in the `required` job `needs` list

### What's kept
The `node-uniffi` crate, example apps, and build scripts remain in the repo — only CI pipeline jobs and release automation are removed. Deprecation notices have been added to relevant docs.

## Test plan
- [ ] Verify CI passes without the removed jobs
- [ ] Confirm `required` check still gates on all remaining jobs


rootulp (approved):
Do we need a follow up PR to add to docs or release notes that iOS and Android are no longer supported?

Context that is missing from the PR description is that the Leap wallet is being sunset and AFAICT they are the only consumer of the iOS and Android builds in this repo. Worth double checking that there are no other consumers of the iOS and Android artifacts in this repo.

Also PR description could use `!` to signify this is breaking

Ref: https://www.leapwallet.io/

mcrakhman:
> Do we need a follow up PR to add to docs or release notes that iOS and Android are no longer supported?
> 
> Context that is missing from the PR description is that the Leap wallet is being sunset and AFAICT they are the only consumer of the iOS and Android builds in this repo. Worth double checking that there are no other consumers of the iOS and Android artifacts in this repo.
> 
> Also PR description could use `!` to signify this is breaking
> 
> Ref: https://www.leapwallet.io/

Thanks, that makes sense. Will update this PR

rootulp (approved):

PR #963, opened 2026-04-16
feat: add multi-account TX service and refactor tx client

mcrakhman:
## Summary
- Enables PFF submission in an asynchronous thread
- Introduces `MultiAccountTxService` for managing transactions across multiple signer accounts
- Refactors the existing TX client to remove delayed resubmission logic and decouple from the single-signer model
- Updates `NodeManager` to start uploads immediately

## Test plan
- [ ] Existing tx_client_v2 tests pass
- [ ] Integration test `submit_with_worker_and_confirm` passes with new signer flow

### DIFF
--- a/grpc/src/tx_client_v2/mod.rs
+++ b/grpc/src/tx_client_v2/mod.rs
@@ -129,8 +129,8 @@ impl<T> TxIdT for T where T: Clone + std::fmt::Debug {}
 pub enum TxPayload {
     /// Pay-for-blobs transaction.
     Blobs(Vec<celestia_types::Blob>),
-    /// Raw Cosmos transaction body with arbitrary messages.
-    Tx(celestia_types::state::RawTxBody),
+    /// A single cosmos message (e.g. MsgPayForFibre, MsgSend).
+    Message(tendermint_proto::google::protobuf::Any),
 }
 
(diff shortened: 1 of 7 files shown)


grpc/src/tx_client_v2/mod.rs
```diff
@@ -129,8 +129,8 @@ impl<T> TxIdT for T where T: Clone + std::fmt::Debug {}
 pub enum TxPayload {
     /// Pay-for-blobs transaction.
     Blobs(Vec<celestia_types::Blob>),
-    /// Raw Cosmos transaction body with arbitrary messages.
-    Tx(celestia_types::state::RawTxBody),
+    /// A single cosmos message (e.g. MsgPayForFibre, MsgSend).
+    Message(tendermint_proto::google::protobuf::Any),
```
rootulp, line 133: A tx can contain more than one message 🤔

rootulp (approved):

grpc/src/tx_client_v2/mod.rs
mcrakhman: Yes, it can actually contain many messages, this is not prohibited by the cosmos sdk afaiu.

PR #976, opened 2026-06-12
feat(node)!: switch sample and row requests to shrex

mcrakhman:
## Switch sampling from Bitswap to ShrEx

Migrates data-availability sampling off **Bitswap/Shwap** (`beetswap`) onto **ShrEx**, Celestia's dedicated share-exchange protocol.

**Why it matters:** Bitswap *pushed* received blocks into the blockstore automatically (coupling it to the header store). ShrEx is *pull-based* — the `Daser` now fetches samples and persists them itself, which lets us delete a pile of workarounds.

**What changed:**
- Sampling now runs over ShrEx; `Daser` owns the blockstore and writes fetched samples directly.
- Removed `beetswap`, `p2p/shwap.rs`, `SampleBlockstore`, and the `get_sample` cleanup hack.
- Renamed the sampling-result flag `timed_out` → `failed` (ShrEx fails in more ways than timeouts).
- CI bumps: celestia-node `v0.30.2`, celestia-app `v8.0.8`.

**Breaking:**
- `request_row_namespace_data` removed (ShrEx has no per-row namespace-data request; use `request_all_blobs` for whole-block namespace data). **`celestia-client` users are not affected.**
- Event field `timed_out` → `failed`; `P2pArgs` no longer takes a `blockstore` (`DaserArgs` does).


node/src/daser.rs
```diff
@@ -522,40 +540,49 @@ where
...
+                        sample.encode(&mut bytes);
+                        samples.push((cid, bytes.freeze()));
+
+                        false
+                    }
+                    // Validation is done at DA protocol level.
+                    // If we hit the timeout or maximum retries of the request
+                    // we can mark it as failure
+                    Err(P2pError::RequestTimedOut)
+                    | Err(P2pError::ShrEx(ShrExError::MaxTriesReached)) => true,
```
vgonkivs: Switching the sample source from Bitswap to ShrEx widens the error set hitting this match, but the catch-all is unchanged — recoverable errors now crash the whole Daser. 

With Bitswap, get_sample effectively only ever surfaced RequestTimedOut. With ShrEx it now goes through get_header and can return HeaderPruned, HeaderNotSynced, and ShrEx(RequestCancelled). All of  these fall into Err(e) => return Err(e.into()), which propagates through res?  up to the run() loop and stops the Daser entirely ("Daser  stopped because of a fatal error").

node/src/daser.rs
```diff
@@ -352,12 +366,16 @@ where
...
                 Some(cmd) = self.cmd_rx.recv() => self.on_cmd(cmd).await,
                 Some(res) = self.sampling_futs.next() => {
-                    // Beetswap only returns fatal errors that are not related
+                    // The future only returns fatal errors that are not related
                     // to P2P nor networking.
-                    let (height, timed_out) = res?;
+                    let (height, samples, failed) = res?;
+
+                    for (cid, sample) in samples {
+                        self.blockstore.put_keyed(&cid, &sample).await?;
```
vgonkivs: propagates any BlockstoreError up through run() and stops sampling for the whole node. IMO transient write failure shouldn't take down DAS

vgonkivs (commented):
Left a few comments but other than that LGTM. 

Also, jFYI GetRangeNamespaceData request is missed. I think we should add it to be consistent with the Go implementation(not here).

vgonkivs (approved):

PR #977, opened 2026-06-22
feat: share one failover engine across RPC and gRPC clients

mcrakhman:
## What & why

RPC and gRPC each had their own multi-endpoint failover with different
semantics and duplicated logic. This extracts a single transport-agnostic
engine and reuses it from both, with principled, non-flapping switch-back to
the most-preferred endpoint once it recovers.

## The shared engine — `lumina-utils::failover` (new `failover` feature)

`Failover<C, E>` is generic over the transport client (`C`) and its error
type (`E`). Routing is a **pure function of per-endpoint health flags** — there
is deliberately no stored "active" index:

- a priority-ordered endpoint list (`[0]` = most preferred) with lazy or
  pre-built clients, each carrying a `healthy` flag,
- `active()` is **derived** as the lowest-index healthy endpoint, so an
  in-flight fallback request can never clobber a concurrent recovery of a
  more-preferred endpoint,
- network-error classification is injected as `fn(&E) -> bool`,
- `run()` — try healthy endpoints in priority order (unhealthy ones only as a
  last resort), fail over **only on network errors** (application / JSON-RPC
  errors are surfaced as-is), and mark only the touched endpoint healthy /
  unhealthy,
- `recover_unhealthy_with()` / `spawn_health_check()` — a background task that
  probes every currently-unhealthy endpoint and revives the ones that have
  recovered; reviving a higher-priority endpoint automatically routes back to
  it.

### Why health flags instead of an `active` pointer

A single mutable `active` index meant any successful request re-pinned it. An
in-flight fallback request finishing *after* the health-check switched back to
the preferred endpoint would silently re-pin the fallback (no warn — it
succeeded), causing perpetual ping-pong at the health-check cadence and a flood
of "switched back to preferred endpoint" debug logs. Deriving routing from
per-endpoint flags removes the shared winner pointer, so that bug class is gone
structurally. Logs now fire only on genuine `unhealthy -> healthy` transitions.

## RPC (`celestia-rpc`)

`FailoverClient` is a thin jsonrpsee adapter (`ClientT` / `SubscriptionClientT`)
over the engine. Probe: `header.NetworkHead`.

## gRPC (`celestia-grpc`)

The `#[grpc_method]` macro now delegates to `failover.run(|transport| …)`
instead of the hand-rolled per-method `ArcSwap` swap-to-front loop; the
`arc-swap` dependency is dropped. The background health-check is **enabled by
default** (a no-op for a single endpoint, so a task only starts when a fallback
exists). Probe: lightweight node `Status` (no full block), carrying
per-endpoint auth metadata.

## Head-freshness recovery (configurable)

An endpoint is revived only if it answers the probe **and** its **head
timestamp** is within `max_head_age` (default 30s) — a timestamp check, not a
block-height comparison, so a node that reconnected but is still syncing (old
head) won't be routed back to. Configurable via `ClientBuilder::rpc_max_head_age`
and `GrpcClientBuilder::max_head_age`.

## Configuration example

All failover knobs are optional and default sensibly. Endpoints are tried in the
order given — the first is the most preferred:

```rust
use std::time::Duration;
use celestia_client::{Client, Endpoint, RpcEndpoint};

let client = Client::builder()
    // Preferred RPC endpoint (index 0), then fallbacks tried in order.
    .rpc_url("ws://localhost:26658")
    .rpc_endpoints([
        // A fallback that needs its own auth token (e.g. a provider).
        RpcEndpoint::new("wss://rpc.example.com").auth_token("my-token"),
        "wss://backup.example.com",
    ])
    // gRPC preferred + fallback (per-endpoint metadata / timeout supported).
    .grpc_endpoints([
        Endpoint::from("http://localhost:9090"),
        Endpoint::from("https://grpc.example.com").metadata("x-token", "my-token"),
    ])
    // Failover tuning (optional; defaults shown).
    .rpc_health_check_interval(Duration::from_secs(3)) // recovery probe cadence
    .rpc_max_head_age(Duration::from_secs(30))         // freshness bar to switch back
    .connect_timeout(Duration::from_secs(2))           // detect a down host faster
    .timeout(Duration::from_secs(10))                  // per-request timeout
    .build()
    .await?;
```

A single endpoint works exactly as before — no background task is started. With
two or more, the health-check runs automatically. The gRPC side exposes the
equivalent `GrpcClientBuilder::health_check_interval` / `max_head_age` when
building a `GrpcClient` directly.

## Defaults

- health-check interval: **3s** (lowered from 10s so a recovered endpoint is
  preferred again sooner; overridable per client),
- probe timeout: 5s, max head age: 30s.

## Verification

- `cargo build` (workspace) and `cargo build --target wasm32-unknown-unknown`
  for `celestia-rpc` / `celestia-grpc` / `celestia-client`
- `cargo test -p lumina-utils --features failover` (engine unit tests:
  failover, app-errors-don't-fail-over, recovery + higher-priority promotion,
  stale-head guard, fallback-success-does-not-demote-preferred)
- gRPC / RPC integration tests require a local devnet and were not run in the
  CI-less env.

### DIFF
--- a/utils/src/failover.rs
+++ b/utils/src/failover.rs
@@ -1,0 +1,713 @@
+//! Transport-agnostic multi-endpoint failover engine.
+//!
+//! [`Failover`](crate::failover::Failover) holds a priority-ordered list of endpoints and runs logical
+//! requests against the most-preferred reachable one, transparently falling
+//! back to the next endpoint on a *network* error. An optional background
+//! health-check probes the preferred endpoint and switches back to it once it
+//! has recovered *and its head is fresh* (within a configurable max age).
+//!
+//! The engine knows nothing about the concrete transport (`C`) or its error
+//! type (`E`): callers inject
+//!
+//! - a per-endpoint *build* closure that lazily constructs the transport
+//!   client (or pre-cache it for transports that are built eagerly),
+//! - an `is_network_error` classifier deciding which errors trigger failover,
+//! - a *call* closure performing a single request against one client, and
+//! - (for switch-back) a *probe* closure returning an endpoint's current head
+//!   timestamp.
+//!
+//! Note: the switch-back "healthy" criterion is purely the head's **timestamp
+//! age**, not its block height. An endpoint can be a few blocks behind and still
+//! count as healthy if its head time is within `max_head_age`. This favours a
+//! simple, transport-agnostic signal over strict height parity.
+//!
+//! This lets both the JSON-RPC and the gRPC clients share one implementation.
+
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex, Weak};
+use std::time::Duration;
+
+use tracing::{debug, warn};
+
+use crate::executor::spawn;
+use crate::time::{Interval, timeout};
+
+/// A future that builds a client for a single endpoint.
+///
+/// It is `Send` on all targets (see [`box_build`]) so that the futures awaiting
+/// it stay `Send` even on wasm, matching what transport client traits require.
+/// Produce one with [`box_build`].
+pub type BuildFut<C, E> = std::pin::Pin<Box<dyn std::future::Future<Output = Result<C, E>> + Send>>;
+
+type BuildFn<C, E> = Arc<dyn Fn() -> BuildFut<C, E> + Send + Sync>;
+
+/// Box an endpoint-build future into a `Send` [`BuildFut`].
+///
+/// On wasm the underlying transport futures are `!Send`, so we wrap them in
+/// `send_wrapper::SendWrapper`, which is valid on wasm's single thread.
+#[cfg(not(target_arch = "wasm32"))]
+pub fn box_build<C, E, F>(fut: F) -> BuildFut<C, E>
+where
+    C: 'static,
+    E: 'static,
+    F: std::future::Future<Output = Result<C, E>> + Send + 'static,
+{
+    Box::pin(fut)
+}
+
+/// Box an endpoint-build future into a `Send` [`BuildFut`] (wasm32 variant).
+#[cfg(target_arch = "wasm32")]
+pub fn box_build<C, E, F>(fut: F) -> BuildFut<C, E>
+where
+    C: 'static,
+    E: 'static,
+    F: std::future::Future<Output = Result<C, E>> + 'static,
+{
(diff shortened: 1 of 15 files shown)


utils/src/failover.rs
```diff
@@ -0,0 +1,772 @@
...
+                Ok(value) => {
+                    // Marking only this endpoint healthy is enough; routing is
+                    // derived from the flags, so a fallback success cannot demote
+                    // or override a more-preferred endpoint.
+                    slot.set_healthy(true);
+                    return Ok(value);
+                }
+                Err(err) if (self.is_network_error)(&err) => {
+                    warn!("failover: endpoint {} network error: {err}", slot.label);
+                    slot.set_healthy(false);
```
vgonkivs, line 316: The cached client is never invalidated here — we mark the endpoint unhealthy but keep its cached client, and get_or_build() only rebuilds when the cache is empty. That's fine for gRPC (tonic channels self-heal), but celestia_rpc::Client is a jsonrpsee ws client that won't reconnect itself, so after a real disconnect the recovery probe reuses the dead client and the endpoint never switches back. Should we drop the cached client for lazy  endpoints (build.is_some()) on a network error so recovery reconnects, while keeping prebuilt gRPC transports cached?

mcrakhman: Good catch that the engine never invalidates the cache — that part is accurate. But the dead-client scenario doesn't actually occur, because the RPC client already self-heals.

`celestia_rpc::Client::Ws` isn't a bare jsonrpsee `WsClient` — it's `WsReconnectClient<WsClient>` (`rpc/src/client.rs:40`). On `ClientError::RestartNeeded` (the signal that a ws connection dropped) it marks itself poisoned (`client.rs:187-199`, `162-168`), and the next call rebuilds the inner `WsClient` via the stored factory (`client.rs:140-160`). `Client::Ws(c) => c.request(...)` delegates straight to that (`client.rs:289-290`).

So after a real disconnect: the failing request returns `RestartNeeded` → the engine marks the endpoint unhealthy **and** the wrapper poisons itself → the recovery probe (`header_network_head`) goes through the *same cached* `Client`, `get_inner()` rebuilds the connection, and if the node is back and fresh the endpoint switches back. gRPC self-heals too — transports are built with `connect_lazy` (`grpc/src/builder.rs:481`, `512`), so a tonic `Channel` reconnects on demand.

vgonkivs (approved):

PR #983, opened 2026-08-27
perf(grpc): right-size broadcast encoding buffers

citizen-stig:
## Overview

Use a request-sized protobuf codec for `GrpcClient::broadcast_tx`.

Tonic normally grows its encode buffer while serializing large requests. The
custom codec allocates the required frame size up front while preserving the
existing path, metadata, timeout, failover, and message limits.

The local timeout covers both transport readiness and the RPC call. A focused
regression test exercises a transport that never becomes ready.

## Results vs #982

| Phase | Total allocated before | Total allocated after |
|---|---:|---:|
| broadcast | 28.08 MiB | 14.07 MiB |
| grpc-submit | 86.62 MiB | 72.61 MiB |
| full | 117.10 MiB | 103.09 MiB |

Peak live memory for the 7 MiB request falls from 49.01 MiB to 42.01 MiB in
the gRPC-submit phase.

Client-side latency is unchanged within run-to-run noise; the benefit is lower
allocation volume and peak memory.

### DIFF
--- a/client/benches/submit_blob_allocations/native.rs
+++ b/client/benches/submit_blob_allocations/native.rs
@@ -305,6 +305,9 @@ async fn profile_broadcast(config: &Config, profile_path: &Path) -> Result<Profi
         Ok(response) => {
             black_box(response);
         }
+        Err(error) if error.is_network_error() => {
+            return Err(format!("broadcast benchmark transport failed: {error}").into());
+        }
         Err(error) => println!("server response after receiving transport payload: {error}"),
     }
     Ok(summary)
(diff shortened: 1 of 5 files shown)


rach-id (commented):
LGTM. LLM flagged this:
```
  ### High — Transport readiness is outside the request timeout

  At grpc/src/client.rs:305 (https://github.com/celestiaorg/lumina/blob/f8fba1356323e4d96f204db398388e6c522d07f9/grpc/src/client.rs#L305-L317), client.ready().await executes before the future passed to
  lumina_utils::time::timeout.

  Previously, the macro created the generated broadcast_tx future—which includes tonic’s readiness wait—and then wrapped that entire future in the timeout (macro implementation
  (https://github.com/celestiaorg/lumina/blob/5666453c55d3573ac69ab8667efddae6e7a594c5/grpc/grpc-macros/src/lib.rs#L91-L99)).

  Consequences when poll_ready remains pending because of transport backpressure or a custom transport:

  - Per-call and endpoint timeouts are ignored during readiness.
  - The call can exceed its advertised deadline or hang indefinitely.
  - No DeadlineExceeded reaches the failover engine, so fallback endpoints are never attempted.

  I reproduced this using the public custom-transport API: a 10 ms broadcast timeout remained pending until an independent 100 ms watchdog expired.

  The readiness check and unary call should be placed in one future, with the WASM wrapper and local timeout applied around that entire future.
```

Will defer to @mcrakhman for final approval

PR #985, opened 2026-08-27
chore(fibre): use more existing types instead of bytes and strings

citizen-stig:
## Overview

Use existing `Namespace`, `AccAddress`. Use `NonZeroU64` to have non-zero checks to be done at the conversion edge.


rach-id (approved):
utAck

rach-id (approved):

PR #992, opened 2026-08-28
fix(fibre): fibre-client does proper download

citizen-stig:
## Overview

Update to the recent celestia-app logic and protobuf:

* `BlobShard` previously distinguished upload coefficients from a download-only RLC root. It now always carries the full original-row RLC vector in field 2.
* Rows are verified before entering reconstruction

Extract row validation into helper methods, so download flow is smaller and easier to reason about

Minor things:

* use `.address_hex` helper for consistent hex formatting
* type alias for better readability

### DIFF
--- a/fibre/src/client/download.rs
+++ b/fibre/src/client/download.rs
@@ -3,12 +3,10 @@
 //! Retrieves a blob from validators and reconstructs it using erasure coding
 //! with an adaptive fan-out strategy.
 
-use futures::stream::{FuturesUnordered, StreamExt};
+use futures::StreamExt;
 use tokio_util::sync::CancellationToken;
 
-use lumina_utils::cond_send::BoxFuture;
-
-use crate::client::task::spawn_task;
+use crate::client::task::{TaskSet, spawn_task};
 
 use crate::ValidatorInfo;
 use crate::blob::{Blob, BlobID};
@@ -17,6 +15,7 @@ use crate::client::FibreClient;
 use crate::config::BlobConfig;
 use crate::error::FibreError;
 use crate::validator::ValidatorSet;
+use crate::validator_client::DownloadResponse;
 
 /// Options for configuring blob download behavior.
 #[derive(Default)]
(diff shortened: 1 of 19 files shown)


fibre/src/client/download.rs
```diff
@@ -437,7 +506,11 @@ mod tests {
             for i in 0..total_rows {
                 proofs.push(blob.row(i).unwrap());
             }
-            conn.store_proofs(blob_id.commitment(), proofs);
+            conn.store_proofs(
+                blob_id.commitment(),
+                proofs,
+                blob.rlc_coeffs().unwrap().to_vec(),
```
citizen-stig, line 634: why unwrap

fibre/src/domain/blob.rs
```diff
@@ -252,31 +253,40 @@ impl Blob {
...
+                applied += 1;
+            }
+        }
+        Ok(applied)
+    }
 
-        rsema1d::verify_row_inclusion_proof(&proof, &self.id.commitment(), &params)?;
+    /// Returns which row indices are already stored, for pre-filtering
+    /// downloaded shards before verification.
+    pub(crate) fn stored_rows_bitmap(&self) -> Vec<bool> {
```
citizen-stig, line 272: bitmap is not great name for vec of bools

fibre/src/domain/blob.rs
```diff
@@ -363,6 +373,123 @@ impl Blob {
...
+                    rsema1d::verify_row_with_context(&needed[0], &self.commitment, &context)?;
+                    *cache = Some((rlcs.to_vec(), Arc::clone(&context)));
+                    (context, 1)
+                }
+            }
+        };
+
+        let remaining = &needed[verified..];
+        for (index, proof) in remaining.iter().enumerate() {
+            rsema1d::verify_row_with_context(proof, &self.commitment, &context)?;
```
mcrakhman, line 470: **performance:** Internally this re-derives coeffs for each row, we can try to calculate the coeffs beforehand and then reuse them, because we already have row root even before the download

mcrakhman (commented):
Left one perf comment. The other important correctness comment is:
- it seems we should remove in `blob.reconstruct()?` after `reconstruct` the following lines:
```rust
        // Verify commitment by re-encoding
        let (ext_data, reconstructed_commitment, rlc_coeffs) =
            rsema1d::encode(&reconstructed, &params)?;

        if self.id.commitment() != reconstructed_commitment {
            return Err(FibreError::CommitmentMismatch {
                expected: hex::encode(self.id.commitment()),
                actual: hex::encode(reconstructed_commitment),
            });
        }
```        

We already checked that all the rows are correct and we don't need to check the commitment. Also the reconstructed commitment can be different from the initial one (see the whole tampered blobs problem here - https://linear.app/celestia/issue/PROTOCO-1330/check-encoding-correctness-rlc-when-downloading-rows)

fibre/src/domain/blob.rs
citizen-stig: Could you please elaborate where exactly row roots are available and how coefficients can be recalculated?

mcrakhman (approved):

PR #1006, opened 2026-09-02
perf(fibre): zero copy rows

citizen-stig:
## Overview

Avoid copying encoded Fibre rows while building upload requests.

After encoding, row matrices are backed by shared `Bytes` storage. Row inclusion proofs and generated protobuf `BlobRow` messages retain slices of that storage instead of allocating and copying each row. Mutable `RowMatrix` access remains supported through copy-on-write, and the protobuf wire format is unchanged.

For a maximum-sized blob, the upload-shard build and encode benchmark improves from 501 µs to 211 µs (~58%).

### DIFF
--- a/fibre/src/transport/proto_conv.rs
+++ b/fibre/src/transport/proto_conv.rs
@@ -79,7 +79,8 @@ pub(crate) fn blob_row_to_row_proof(
 ///
 /// Shards carry the RLC vector of the K original rows (16 bytes per row) so
 /// the validator can verify each row without having enough rows to reconstruct.
-pub(crate) fn build_upload_shard(
+#[doc(hidden)]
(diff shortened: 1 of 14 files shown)


fibre/src/transport/proto_conv.rs
```diff
@@ -79,7 +79,8 @@ pub(crate) fn blob_row_to_row_proof(
 ///
 /// Shards carry the RLC vector of the K original rows (16 bytes per row) so
 /// the validator can verify each row without having enough rows to reconstruct.
-pub(crate) fn build_upload_shard(
+#[doc(hidden)]
+pub fn build_upload_shard(
```
citizen-stig: no bueno

PR #1019, opened 2026-09-04
[no-merge]: fibre throughput eval harness

citizen-stig:
## Overview

Tested against mock-validator implementation in ....

Tested on 2 `c6in.32xlarge` with same placement group. NIC traffic goes to ~70 Gbps.

The maximum-throughput configuration preloads jemalloc in both the evaluator and mock-server processes (`LD_PRELOAD=/lib/x86_64-linux-gnu/libjemalloc.so.2`).

Mock server advertised 32 validators to reduce single gRPC channel limitations

## Numbers

* Single signer `~1.5 GiB/s` against 10 mock validators, 8 encode threads, 64 rayon.
* 16 signers `2.52 GiB/s` against 10 mock validators, 4 encode / 4 rayon per signer.
* 16 signers `2.7 GiB/s` against 32 mock validators, 4 encode / 4 rayon per signer.

This is with jemalloc allocator

## Follow-up performance opportunity

Evaluator throughput may improve slightly by generating payload bytes directly into the upload-ready row matrix instead of allocating a separate payload `Vec` and copying it in `EncodedBlob::new`. That would require an initializer-style encoded-blob API and is intentionally left as a follow-up.


fibre/src/transport/tls.rs
```diff
@@ -223,6 +223,12 @@ impl FibreH2TransportInner {
             .await
             .map_err(BoxError::from)?;
         let (sender, connection) = hyper::client::conn::http2::Builder::new(h2_executor())
+            // On a fast, low-latency link the default 64 KiB windows and 1 MiB
+            // send buffer throttle shard uploads while CPU and NIC sit idle.
+            // Adaptive windowing tunes the receive side to the BDP; the larger
+            // send buffer keeps the upload pipe full between WINDOW_UPDATEs.
+            .adaptive_window(true)
```
citizen-stig, line 230: todo: validate that and extract to separate PR

PR #1055, opened 2026-09-15
perf(rsema1d): parallelize reconstruction with bounded work buffers

citizen-stig:
## Overview

Parallelizes reconstruction across stripes, reuses decoder buffers, and skips decoding when all original rows are supplied. Adds an explicit work-buffer budget API and corresponding benchmark parameter support.

### Performance

Same machine and sampling configuration as the baseline, using the default work budget.

```
   Input rows                Baseline       This PR
  ━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━━━
   Original                581.870 ms     11.528 ms
  ─────────────────────  ─────────────  ────────────
   Mixed, 25% original    1697.685 ms    185.017 ms
  ─────────────────────  ─────────────  ────────────
   Parity                 1680.720 ms    185.381 ms
```

### Reproduce

Run on each branch, substituting original or parity for the other cases:

```
RAYON_NUM_THREADS=8 cargo bench -p rsema1d --bench reconstruct_sweep -- mixed
```

Override the work budget:

```
  RSEMA1D_RECONSTRUCT_WORK_BUDGET_MIB=32 RAYON_NUM_THREADS=8 \
    cargo bench -p rsema1d --bench reconstruct_sweep -- mixed
```

###  What reconstruction does

Imagine four original rows, A, B, C, and D, plus parity rows containing recovery information.

You receive A, C, and two parity rows. Reconstruction uses those four rows to recover B and D, returning all four originals. That is the purpose of this API: recover the original data from enough sampled rows.

### Before and after

Previously, one decoder received entire rows and reconstructed the missing originals.

Now, each decoder receives a stripe: the same byte range from every supplied row. Different stripes can be reconstructed independently.

For example, suppose rows B and D are missing and each row contains 1,024 bytes:

```
  Decoder 1: recover bytes   0–255 of B and D
  Decoder 2: recover bytes 256–511 of B and D
             ↓ copy results into the output
  Reuse both decoders for the remaining bytes
```

Each batch finishes decoding, copies its results into the output, then reuses the decoders for the next batch. Decoding and copying are separate phases.

**References:**

original implementation:
https://github.com/celestiaorg/lumina/blob/728f1c6dec1c200edae2404dec79b92621e8aae8/rsema1d/src/codec/reconstruct.rs#L53-L93
  
parallel implementation 

https://github.com/celestiaorg/lumina/blob/f222592a1a300a6a266b292cf3aea6af559eaa68/rsema1d/src/codec/reconstruct.rs#L159-L202

### Why this could help

Consider the benchmark dimensions: 4,096 original rows, 12,288 parity rows, and 32 KiB per row—128 MiB of original data.

With missing originals and eight Rayon workers, the calculated decoder storage is:

```
                                                Before    After, with a 32 MiB budget
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   Decoder instances                                 1                              8
  ───────────────────────────────────────────  ────────  ─────────────────────────────
   Bytes processed from each row per decoder    32 KiB                      128 bytes
  ───────────────────────────────────────────  ────────  ─────────────────────────────
   Combined main shard buffers                   1 GiB                         32 MiB
  ───────────────────────────────────────────  ────────  ─────────────────────────────
   Decode batches                                    1                             32
``

This allows several workers to contribute to one reconstruction and gives each decoder a smaller working buffer. Better cache reuse is a plausible benefit.

The budget covers decoder working buffers, not total memory. The 128 MiB output and input storage are additional. It also applies per call: concurrent reconstructions each have their own allocation.

References: stripe planner (https://github.com/celestiaorg/lumina/blob/f222592a1a300a6a266b292cf3aea6af559eaa68/rsema1d/src/codec/rs.rs#L49-L68), budget contract
  (https://github.com/celestiaorg/lumina/blob/f222592a1a300a6a266b292cf3aea6af559eaa68/rsema1d/src/codec/mod.rs#L129-L139).

### What it costs

Smaller pieces mean more operations. In that example:

  - Input submission changes from 4,096 whole-row additions to 1,048,576 stripe additions. The total supplied payload remains 128 MiB.
  - The backend repeats its missing-row setup 256 times instead of once, even though the missing-row pattern is unchanged.
  - Copying recovered data also happens in many smaller pieces.

### CODE REVIEW

File: rsema1d/src/codec/reconstruct.rs [hunk 1]
```diff
@@ -223,33 +181,31 @@ fn reconstruct_data_with_plan(
...
+                if !missing[index] {
                     return;
                 }
 
-                for (&stripe, restored) in batch.iter().zip(&restored) {
-                    let start = position * stripe.len;
+                for (&stripe, result) in batch.iter().zip(&results) {
                     row[stripe.offset..stripe.offset + stripe.len]
-                        .copy_from_slice(&restored[start..start + stripe.len]);
+                        .copy_from_slice(result.restored_original(index).unwrap());
```
citizen-stig: unwrap no-bueono, why

mcrakhman (approved):

PR #1062, opened 2026-09-16
fix(fibre): offload native blob reconstruction on bloking tokio

citizen-stig:
## Overview

Fix for tokio scheduler, so CPU intensive reconstruction does not block other async requests.


fibre/src/client/download.rs
```diff
@@ -25,6 +25,18 @@ pub struct DownloadOptions {
     pub height: Option<u64>,
 }
 
+#[cfg(not(target_arch = "wasm32"))]
+async fn reconstruct_blob(reconstruction: BlobReconstruction) -> Result<Blob, FibreError> {
+    tokio::task::spawn_blocking(move || reconstruction.reconstruct())
+        .await
+        .expect("blob reconstruction task failed")
```
citizen-stig: why this expect?

mcrakhman (approved):

PR #1070, opened 2026-09-17
fix(fibre): offload native upload proof generation

citizen-stig:
## Overview

To prevent delays in tokio scheduling


citizen-stig:
No signs of improvement

PR #1075, opened 2026-09-18
fix(types): check share proofs cover a continuous range

mcrakhman:
- Closes the bug where `ShareProof::verify` accepted shares which are not adjacent in the square. It only proved that each share is included in some row, so shares from different places could be presented as one range. `RowProof::verify` did not bind row roots to the rows they claim to be either. Both now do.

- Add methods `ShareProof::verify_range` that checks that we got exactly the range that we asked for, also add `covered_range` which returns the range that proofs covers.

- `ShareProof` stays responsible only for a continuous range of shares, not for a whole blob. It is what `share.GetRange` (celestia-node) returns, and that api proves any range of shares within a single namespace: a part of a blob, several blobs, a blob with the namespace padding after it, or shares of a reserved namespace. Blob level checks would reject those valid proofs.

- `BlobProof` is added for proving blobs. It holds the same data as `ShareProof` and on top of its verification checks that the range starts with a first share of a blob and covers exactly the amount of shares the blob occupies, so it can't be a part of a blob or a blob with something else appended.


types/src/blob/proof.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// [`Blob::reconstruct`]: crate::Blob::reconstruct
+    /// [`Commitment`]: crate::Commitment
+    /// [`DataAvailabilityHeader`]: crate::DataAvailabilityHeader
+    /// [`Namespace`]: crate::nmt::Namespace
+    pub fn verify(&self, root: Hash) -> Result<()> {
+        self.0.verify(root)?;
+
+        // reserved namespaces hold compact shares, which have a different layout
+        if self.0.namespace_id.is_reserved() {
+            bail_verification!(
```
citizen-stig: So this error converts to VerificationError::Other(String)`

But there's an error `Error::UnexpectedReservedNamespace`, wouldn't that be more appropriate here?

citizen-stig (approved):

